PR #216, opened 2020-08-31
Started the interactive module

dolphingarlic:
Should I add more problems? I could totally just refer the reader to https://oj.uz/problems?search=&type=INTERACTIVE though...


dolphingarlic:
Are all the tips fine?

bqi343:
Sure.

bqi343:
I don't think "saveit" should be spoiled though :)

bqi343:
I don't rlly agree with these difficulties ("Kanji Shiritori" doesn't seem harder than "Stray Cat") but ok for now

PR #240, opened 2020-09-10
Gold descriptions

andrewwangva:


bqi343:
Oops I don't particularly like these (also why was the description for DP NT changed?)

andrewwangva:
> Oops I don't particularly like these (also why was the description for DP NT changed?)

Yea, I couldn't think of a good way to describe a lot of them, but I thought something was better than nothing. Also I thought the wording for DP NT was a little weird. I'll probably change some of the descriptions tmmrw.

bqi343:
ig we'll see whether anyone can do a better job of this

PR #242, opened 2020-09-10
Sorting sets section

Farrius:
In sorting with comparators I starred some problems , in this problems the main solution requires sorting.
I changed mountain view to normal since I think it requires some important observations that are not just sorting.
Also I added the c++ solutions.
For rental service I used spanish variables so if you have the solution for that problem put your solution instead of mine.

In more operation on ordered sets I changed:
jury marks hard => normal, becuase the O(n^2) solution works and it is just check if exists.
snow boots normal => hard, I think it is harder, also the silver version is labeled as hard in the dfs section.
mahomoud & Ehab & function and Torunament are too hard for me. I did not understand the solution at all and in tournament the solution says something about strongly connected components. (I tried this problems like a month ago and I still not understand the solutions so I give up for now).


Farrius:
Also I put tournament insane because I think it is harder than ehab function.

thecodingwizard:
jan18-rental service variable names should probably be translated, but otherwise looks fine to me? (I didn't check difficulty tags but they're probably fine)

bqi343:
> I did not understand the solution at all and in tournament the solution says something about strongly connected components. (I tried this problems like a month ago and I still not understand the solutions so I give up for now).

Basically, you add an undirected edge between two participants $a$ and $b$ if it is not the case that $a$ is better than $b$ at every sport (or the other way around). Maintain each connected component in this graph. Then for every two connected components $A$ and $B$, if 

$\min_{a\in A}s_{a1}<\min_{b\in B}s_{b1}$ 

then it holds that 

$\max_{a\in A}s_{ai}<\min_{b\in B}s_{bi}$ for each $i\in [1,k]$.

(so we can maintain the components in sorted order by $\min_{a\in A}s_{a1}$)

bqi343:
You should rewrite the solution for split (just rerun the x solution twice with coordinates transposed).

Farrius:
I don't control such power to do that :(

Farrius:
fixed

PR #245, opened 2020-09-11
Fixed up the LIS section quite a bit

dolphingarlic:


bqi343:
seems okay at a glance

PR #259, opened 2020-09-22
Ordered sets module java

andrewwangva:


thecodingwizard (approved):
Seems fine? I just quickly glanced thru oops

content/3_Silver/Intro_Ordered.mdx
```diff
@@ -93,20 +93,62 @@ export const problems = {
...
+
 <Resources>
 	<Resource source="CPH" title="4.4 - Set Iterators" starred></Resource>
 </Resources>
 
 </CPPSection>
 
+<JavaSection>
+
+In java, iterators are helpful in looping through sets.
```
dolphingarlic, line 113:
```suggestion
In Java, iterators are helpful for looping through sets.
```

content/3_Silver/Intro_Ordered.mdx
```diff
@@ -93,20 +93,62 @@ export const problems = {
...
+
+Iterators used on **HashSet** would have the elements in a random order:
```
dolphingarlic, line 115:
```suggestion
Iterators used with `HashSet` would yield the elements in random order:
```

content/3_Silver/Intro_Ordered.mdx
```diff
@@ -93,20 +93,62 @@ export const problems = {
...
+```java
+Set<Integer> set = new HashSet<Integer>();
+set.add(1); set.add(3); set.add(0); set.add(-2);
+Iterator it = set.iterator();
+while(it.hasNext()){
+   Integer i = (Integer)it.next();
+   System.out.print(i + " "); // returns 0 1 -2 3
+}
+```
+But in **TreeSet** the elements are in a sorted order:
```
dolphingarlic, line 125:
```suggestion
But with `TreeSet` the elements are in sorted order:
```

content/3_Silver/Intro_Ordered.mdx
```diff
@@ -93,20 +93,62 @@ export const problems = {
...
+Set<Integer> set = new TreeSet<Integer>();
+set.add(1); set.add(3); set.add(0); set.add(-2);
+Iterator it = set.iterator();
+while(it.hasNext()){
+   Integer i = (Integer)it.next();
+   System.out.print(i + " "); // returns -2 0 1 3
+}
+```
+
+Instead of creating an iterator and looping through it, Java provides a for-each loop which creates a hidden iterator and loops through it automatically:
```
dolphingarlic, line 136:
```suggestion
Instead of creating an iterator and looping with it like in C++, Java provides a for-each loop which creates a hidden iterator and loops with it automatically:
```

content/3_Silver/Intro_Ordered.mdx
```diff
@@ -93,20 +93,62 @@ export const problems = {
...
+```java
+Set<Integer> set = new TreeSet<Integer>();
+set.add(1); set.add(3); set.add(0); set.add(-2);
+for (int i : set){
+   System.out.print(i + " "); // returns -2 0 1 3
+}
+```
+<Warning>
+
+You shouldn't modify sets when traversing it with set iterators like in any other iteterators for Collections (this **INCLUDES** when using a for-each loop). The only modification possible is using the iterator `remove()` method which can only be used once before calling a `next()` method.
```
dolphingarlic, line 146:
```suggestion
You shouldn't modify sets when traversing it with set iterators like in any other iterators for Collections (this **INCLUDES** when using a for-each loop). The only modification possible is using the iterator `remove()` method which can only be used once before calling a `next()` method.
```

dolphingarlic (changes requested):

PR #280, opened 2020-09-30
Update USACO jan17-hoof-paper-scissors.mdx

MrinallU:
Added a java implementation.


thecodingwizard (changes requested):
- update ID
- fix javasection tags
- link to problem

See #282

Thanks! and let me know if anything doesn't make sense :)

PR #281, opened 2020-09-30
Create USACO feb17-maxcross.mdx

MrinallU:
Added a java implementation. Please let me know how to add the problem ID as I don't know that.


thecodingwizard (changes requested):
- update ID
- fix javasection tags
- link to problem

See https://github.com/cpinitiative/usaco-guide/pull/282

Thanks! and let me know if anything doesn't make sense :)

thecodingwizard:
The JavaSection tags still need to be fixed, and the code should be wrapped in ```java ... ```

PR #283, opened 2020-09-30
Add non-HLD cowland solution (as the section implies)

xyzyzl:
Finished cowland solution, includes C++ implementation.


thecodingwizard (approved):

bqi343:
seems okay, although the variables used in the explanation aren't the same as those used in the code.

PR #288, opened 2020-10-01
fix comeback timer on load

bqi343:
Previously it would initially display smth like 7 hrs 36 minutes and then switch to the correct time.

Can't replicate the negative timer issue though. Any ideas?


thecodingwizard:
hmm this is interesting... the initial display flicker is probably due to the server-side rendering -- the Date.now() value is the value when the site is built, and then when the page finishes loading then it's updated to something else. We should probably just hide the daily streak during server-side rendering and only show it when it's run in the browser.

I'm not sure why there's negative values though oops

thecodingwizard:
I added this fix + stopped rendering of dailystreak on server: 10d769747afc8daa47073745d717f519bc356d64

PR #320, opened 2020-10-19
Trapezoids explanation

dolphingarlic:


bqi343:
Move counting LISes (trapezoids, zoltan, sort it out) into a separate module with segtree or fenwick as prereq

bqi343:
Merging for now (even though module / explanation are not complete)

PR #326, opened 2020-10-21
Replace mdx editor with monaco

TanayB11:
I found a Monaco React component, so I just added the dependency and used the component. Also added lazy-loading, and I changed the Tailwind styles a bit so the scrolling issue should be gone. Still yet to connect it to the renderer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "@mdx-js/mdx": "^1.6.16",
     "@mdx-js/react": "^1.6.16",
     "@mdx-js/runtime": "^1.6.16",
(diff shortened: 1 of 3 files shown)


thecodingwizard:
Suspense isn't supported in server-side rendering yet; you might need to render it conditionally:

```
{typeof window !== 'undefined' && (
              <React.Suspense fallback={'Loading'}>
                <RawMarkdownRenderer markdown={markdown} />
              </React.Suspense>
            )}
```

TanayB11:
Re-opening from the new fork. See #331

Issue #333, opened 2020-10-22
Colors for Dropdown


bqi343:
> I think it would be nice to add a small color badge on all the dropdown options (complete, reading, solving, etc.). The quick visual recognition would immensely improve UX in that aspect -- which is a pretty big aspect, as I'm sure a lot of people use it for every single module. :)

MrinallU:
@bqi343 Sorry about the silly question as it is my first time making a website contribution in this particular repository but, where would I find the file to edit the dropdown?

thecodingwizard:
I think here: https://github.com/cpinitiative/usaco-guide/blob/master/src/components/MarkdownLayout/MarkCompleteButton.tsx

MrinallU:
@thecodingwizard Alright, thanks for the fast response, I'll try to get this done by the end of the week.

MrinallU:
Added a PR.

MrinallU:
Would you like the icons to be one color or different colors?

MrinallU:
![Capture](https://user-images.githubusercontent.com/61529929/97055232-6623f200-1554-11eb-9994-bedbf626ded1.JPG)

thecodingwizard:
err probably a different color? (also is it just me or are the skipped/ignored icons sort of weird :P)

Also, can we left align the text/icons?

(We can alternatively just use colored dots rather than icons if that's easier)

MrinallU:
Right I was about to change the skipped and ignored icons and I will change the color to match the nav.

MrinallU:
Ok I think I finished

MrinallU:
![Capture](https://user-images.githubusercontent.com/61529929/97058491-fdd90e80-155b-11eb-853e-903194266b0e.JPG)

MrinallU:
![Capture](https://user-images.githubusercontent.com/61529929/97058608-4a244e80-155c-11eb-8f89-2d3c8d355f60.JPG)

thecodingwizard:
can we make the icons smaller and left-align everything? (basically make the text aligned)

MrinallU:
So the smaller icons makes them look weird and blocky so I added the medium sized ones and they look normal. And I just left aligned the text.

thecodingwizard:
Did you try using the small icons from here: https://heroicons.com/ ?

MrinallU:
Yes I did

Here is a shot with the small icons:

[...]

PR #335, opened 2020-10-23
Added Icons

MrinallU:
Resolved issue #333


thecodingwizard:
Hmm is it possible to do this with tailwind CSS / tailwind UI rather than material since we have to import an entirely new rather large dependency just to use material?

MrinallU:
I'll do that then.

PR #356, opened 2020-10-29
content: fix small typo in bubblesort editorial

thecodingwizard:
At least, I think this is a typo?


dolphingarlic:
Oops yeah you're right

PR #369, opened 2020-11-03
content: add CEOI 09-Harbingers

thecodingwizard:
lol I haven't actually solved this problem yet so I'm not sure if the solution / the difficulty is correct XD

### DIFF
--- a/content/5_Plat/CHT.mdx
+++ b/content/5_Plat/CHT.mdx
@@ -47,7 +47,7 @@ export const problems = {
 			'Normal',
 			false,
 			['DP', 'convex'],
-			'CHT on a tree -- dfs on the tree and keep track of which lines you remove so you can "restore" the deque of lines when leaving a node. Editorial: https://infosb.net/ceoi2009/tasks/solutions.doc'
+			'https://infosb.net/ceoi2009/tasks/solutions.doc'
 		),
 		new Problem(
 			'IOI',
--- a/content/5_Plat/CHT.mdx
+++ b/content/5_Plat/CHT.mdx
@@ -41,6 +41,15 @@ export const problems = {
 			''
 		),
 		new Problem(
+			'CEOI',
+			'2009 - Harbingers',
+			'https://oj.uz/problem/view/CEOI09_harbingers',
+			'Easy',
+			false,
+			['DP', 'convex'],
+			'CHT on a tree -- dfs on the tree and keep track of which lines you remove so you can "restore" the deque of lines when leaving a node'
+		),
+		new Problem(
 			'IOI',
 			'2002 - Batch Scheduling',
 			'https://dmoj.ca/problem/ioi02p4',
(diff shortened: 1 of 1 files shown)


dolphingarlic:
I think it should be at least "Normal". I didn't find the CHT on trees thing immediately obvious, and the trick isn't mentioned anywhere else in the module.

content/5_Plat/CHT.mdx
```diff
@@ -40,6 +40,15 @@ export const problems = {
...
 			''
 		),
+		new Problem(
+			'CEOI',
+			'2009 - Harbingers',
+			'https://oj.uz/problem/view/CEOI09_harbingers',
+			'Normal',
+			false,
+			['DP', 'convex'],
+			'CHT on a tree -- dfs on the tree and keep track of which lines you remove so you can "restore" the deque of lines when leaving a node'
```
dolphingarlic, line 50: Also, the official editorial is https://infosb.net/ceoi2009/tasks/solutions.doc (Mostafa has pdf versions on his GitHub as well)

bqi343:
I always thought you were just supposed to use line container and small to large merging :P

nhtrnm:
> I always thought you were just supposed to use line container and small to large merging :P

That is only true is the answer for the current node depends on the CHT of the nodes in the subtree. In this problem, it's CHT of the ancestor nodes.

Or in other words, if the DP is from leaves up to root then you are correct, it's just line container and small to large merging. If it's the other way round then persistent CHT. And in this case, it can nicely be implemented in a stack.

thecodingwizard:
uhh does it need to be removed from advanced - linecontainer then (?)

nhtrnm:
> uhh does it need to be removed from advanced - linecontainer then (?)

My apologies, I just stumbled across your repo and while solving this problem so I don't know what levels there are. But this problem certainly does not require line container (presumably you are talking about the KACTL one). But it's also not a standard CHT problem as in you can't just come up with the recurrence and voila you are done. I'd say it's hard but not too complex in terms of the data structures that one needs to understand to solve this.

thecodingwizard:
ah ok perfect, thanks!! :)

nhtrnm:
Do you collect solutions for problems? Asking because I could give you mine.

dolphingarlic:
We do collect editorials for [problems in the guide](https://usaco.guide/problems/), but since this particular problem already has an official editorial, you don't need to write one for us. You're more than welcome to give us your solutions to problems that don't currently have editorials though (e.g. many of the POI problems)!

nhtrnm:
Gotcha!

bqi343:
Ok, it's probably because I never implemented this :P

content/5_Plat/CHT.mdx
```diff
@@ -40,6 +40,15 @@ export const problems = {
...
 			''
 		),
+		new Problem(
+			'CEOI',
+			'2009 - Harbingers',
+			'https://oj.uz/problem/view/CEOI09_harbingers',
+			'Normal',
+			false,
+			['DP', 'convex'],
+			'CHT on a tree -- dfs on the tree and keep track of which lines you remove so you can "restore" the deque of lines when leaving a node. Editorial: https://infosb.net/ceoi2009/tasks/solutions.doc'
```
dolphingarlic, line 50:
```suggestion
			'https://infosb.net/ceoi2009/tasks/solutions.doc'
```

Issue #376, opened 2020-11-06
Two Pointers - SAPO Problem


bqi343:
https://usaco.guide/silver/2P

> Could you write a solution for the SAPO one please?

@dolphingarlic

PR #381, opened 2020-11-07
SAPO stadium editorial

dolphingarlic:
Closes https://github.com/cpinitiative/usaco-guide/issues/376


bqi343:
Isn't $y$-coordinate usually referred to as up/down, not left/right? Should also mention that you should know how to solve bit inversions first:

https://usaco.guide/silver/intro-ordered#harder-example---bit-inversions

dolphingarlic:
I said y-coordinate because my code used y instead of x for some reason, but I'll change it

PR #382, opened 2020-11-07
content: bitmasks over primes

thecodingwizard:


bqi343:
Hm, I don't think the "Rough Idea" section adds much to the module.

frodakcin:
yea ig the idea itself is pretty self-explanatory... Should I should walk through the soln of a sample problem instead?

bqi343:
If you want. :)

PR #388, opened 2020-11-08
Dfs stack size

andrewwangva:
not sure how to deal with language-specific issues. Should I even mention stackoverflow for c++?

### DIFF
--- a/content/3_Silver/DFS.mdx
+++ b/content/3_Silver/DFS.mdx
@@ -1,7 +1,7 @@
 ---
 id: dfs
 title: Depth First Search (DFS)
-author: Siyong Huang
+author: Siyong Huang, Andrew Wang
 prerequisites:
   - intro-graphs
   - complete-rec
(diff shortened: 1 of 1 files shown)


bqi343:
You should give an example of a problem + solution that actually exceeds the stack limit.

For C++ link https://usaco.guide/general/cpp-command/#adjusting-stack-size

bqi343:
why was this closed?

frodakcin:
If this helps at all: Museum Tour (https://codeforces.com/contest/1137/problem/C) had a lot of stack overflow issues w/ solutions.

andrewwangva:
> If this helps at all: Museum Tour (https://codeforces.com/contest/1137/problem/C) had a lot of stack overflow issues w/ solutions.

I feel like this problem is too hard to introduce in silver dfs, but I also can't find any dfs problems with stackoverflow, only floodfill :(

frodakcin:
I'd say probably just warn that the problem is difficult (considering this issue only really shows up, rarely, in hard problems)

Oh also just remembered JOI 2018 Virus needed this as well.

andrewwangva:
Can i put  [http://www.usaco.org/current/data/sol_snowboots_gold_feb18.html ](http://www.usaco.org/current/data/sol_snowboots_gold_feb18.html ) for stackoverflow? It's kinda dfs but you could also classify it as recursion....

bqi343:
Do you mean the silver version?

andrewwangva:
> Do you mean the silver version?

Yea whoops [this](http://www.usaco.org/index.php?page=viewproblem2&cpid=811)

bqi343:
Left only Python for now.

PR #395, opened 2020-11-10
Add C++ solution (both naive and constant time) to Fence Painting

dliu426:


bqi343:
Already wrote the first solution here: https://usaco.guide/general/io#method-1. Also can simplify second solution.

PR #398, opened 2020-11-10
Content: Add language tags 

dliu426:
Remove c++ content when another language is selected.
- https://usaco.guide/bronze/intro-sets#inserting--deleting-keys-while-iterating

### DIFF
--- a/content/2_Bronze/Intro_Sets.mdx
+++ b/content/2_Bronze/Intro_Sets.mdx
@@ -486,6 +486,10 @@ for key in dict.keys():
 
 While you are free to change the *values* in a map when iterating over it (as demonstrated above), be careful about inserting and deleting keys while iterating. This code will work:
 
+<LanguageSection>
+
+<CPPSection>
+
 ```cpp
 void iterate_insert() {
 	map<int,int> m;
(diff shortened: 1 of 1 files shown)


thecodingwizard:
@nchn27 @andrewwangva are deleting keys an issue in Java? Or can we safely hide this section in Java?

andrewwangva:
> @nchn27 @andrewwangva are deleting keys an issue in Java? Or can we safely hide this section in Java?

I'm pretty sure there's an error thrown when deleting items that are part of the collection when using an iterator(ever hidden iterators such as for-each loops). The only way to delete from collections while using iterators is using the iterator.remove() method if I remember correctly.

andrewwangva:
> > @nchn27 @andrewwangva are deleting keys an issue in Java? Or can we safely hide this section in Java?
> 
> I'm pretty sure there's an error thrown when deleting items that are part of the collection when using an iterator(ever hidden iterators such as for-each loops). The only way to delete from collections while using iterators is using the iterator.remove() method if I remember correctly.

It talks a little about it in [this](https://usaco.guide/silver/intro-ordered) but maybe we should expand on it if needed?

thecodingwizard:
okay sure. I'll merge this PR for now, but @andrewwangva can you add the corresponding Java section if you feel it's appropriate?

PR #403, opened 2020-11-12
Added Java implementation for Creating Strings 1

UsersBad:


bqi343:
where's the surrounding class? :P

UsersBad:
> where's the surrounding class? :P

Oh sorry, the club powerpoint I was working on only used the snippets, so I just transferred that to here. I'll fix it.

PR #407, opened 2020-11-12
Add python code for Binary_Search_Ans.mdx

topazand:
Add python code

### DIFF
--- a/content/3_Silver/Binary_Search_Ans.mdx
+++ b/content/3_Silver/Binary_Search_Ans.mdx
@@ -292,6 +292,43 @@ public class test{
 
 </JavaSection>
 
+<PySection>
+
+```python
+def lstTrue(f, lo, hi):
+	""" Binary Search
+    :param f: (lambda function) check a state
+	:param lo: (int) lower bound
+	:param hi: (int) upper bound
+    :return res: (int) the maximum x such that f(x) is true
+    """
+	res = lo-1
+	while lo <= hi:
+		mid = (lo+hi)//2 # find the middle of the current range
+		if f(mid): 
+			# if mid works, then all numbers smaller than mid also work
+			# so we only care about the part after mid
+			res = mid # update the answer
+			lo = mid+1 # cut the part before mid
+			# notice that we already handled mid itself so we cut it as well
+		else:
+			# if mid does not work, greater values would not work too
+			# so we don't care about them
+			hi = mid-1 # cut mid and after
+	# now res is the answer
+	# if res is l-1 that means no value in the range satisfies the condition
+	return res
+print(lstTrue(lambda x:True, 2, 10)) # 10
+# all numbers satisfy the condition
+print(lstTrue(lambda x:x*x <= 30, 2, 10)) # 5
+print(lstTrue(lambda x:False, 2, 10)) # 1
+# no numbers satisfy the condition
+```
+See [Lambda Expressions](https://www.w3schools.com/python/python_lambda.asp) if you're not familiar with the syntax used in the program. 
+
+</PySection>
+
+
 </LanguageSection>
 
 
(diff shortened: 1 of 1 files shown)


content/3_Silver/Binary_Search_Ans.mdx
```diff
@@ -328,6 +365,20 @@ public static int lstTrue(int lo, int hi) {
 
 </JavaSection>
 
+<PySection>
+
+```python
+def lstTrue(f, lo, hi): # python scoping...
+    lo-=1
+    locs=locals()
```
allenli873, line 373: I'm a little bit confused about where all the local and global things came from, and why this is necessary. If it is necessary, it needs to be explained more (because if I'm confused, then they'll probably be very confused). If it can be taken out, it would be a good idea to.

content/3_Silver/Binary_Search_Ans.mdx
```diff
@@ -362,6 +413,21 @@ static long search(){
 
 </JavaSection>
 
+<PySection>
+
+```python
+def search():
+    pos = 0
+	a = mx = 2E9
+	while a >= 1:
```
allenli873, line 422: indentation looks a little inconsistent here, and this matters for python.

content/3_Silver/Binary_Search_Ans.mdx
```diff
@@ -536,6 +625,40 @@ static boolean check(long x){
...
+<PySection>
+
+```python
+# checks whether the number of given operations is sufficient
+# to raise the median of the array to x
+def check(x):
+    operationsNeeded = 0
+	for i in range((n-1)//2,n):
+		operationsNeeded += max(0, x-v[i])
+    if operationsNeeded <= k:
```
allenli873, line 637: weird indentation going on here

content/3_Silver/Binary_Search_Ans.mdx
```diff
@@ -536,6 +625,40 @@ static boolean check(long x){
...
+		return True
+    else:
+		return False
+
+
+# binary searches for the correct answer
+def search():
+    pos = 0
+	a = mx = 2E9
+	while a >= 1:
```
allenli873, line 647: and here

allenli873 (commented):
yes

content/3_Silver/Binary_Search_Ans.mdx
topazand: Oh this is just some weird python scoping rules..
If we want to use python ternary expressions like (a = b if cond else c) here then we need exec and those locales.
You can just remove this python code since that python ternary expression only accept expressions but not statements.

content/3_Silver/Binary_Search_Ans.mdx
topazand: Why... the indentation is correct in my vscode
image

content/3_Silver/Binary_Search_Ans.mdx
topazand: This is really weird

topazand: > and here

I'll try to fix this

allenli873: It looks like the problem is that you're using both tabs and spaces.

content/3_Silver/Binary_Search_Ans.mdx
thecodingwizard: be careful with tabs vs spaces.... https://usaco.guide/liveupdate/

content/3_Silver/Binary_Search_Ans.mdx
allenli873: yea i see what you mean...
honestly this implementation is far too cumbersome to be worth it (ternary should shorten the code not lengthen it), so you have 2 options:
1. leave out the python implementation for this (or add a small blurb stating reasons why there isn't a python imp)
2. put a small blurb above the code basically recommending people to not do this in python, why people shouldn't do this, and having the code there for anyone who is curious.
either option is fine but as it is right now it's counterintuitive

content/3_Silver/Binary_Search_Ans.mdx
topazand: I see

content/3_Silver/Binary_Search_Ans.mdx
topazand: sure

allenli873 (approved):
LGTM

PR #409, opened 2020-11-12
content: add f-string warning

dliu426:


bqi343:
I don't think we need to include this. f-strings have been around since python 3.6 but USACO uses python 3.4.

dliu426:
Yeah, thats true. Just thought people would get confused because there were some python code [here](https://usaco.guide/bronze/intro-sorting#dynamic-arrays-of-pairs--tuples) that used f-strings.

bqi343:
can't hurt

dongliu0426:
ah now this has to be reverted

PR #411, opened 2020-11-13
Added python code for open16-diamond

xyzyzl:
Sorry for the misspelled commit message. Python is not an undergarment.


bqi343:
Use liveupdate to check whether it renders ...

PR #425, opened 2020-11-16
Add cowmbat to problem list

dongliu0426:
- https://usaco-guide-git-dliu426-patch-11.cpinitiative.vercel.app/gold/intro-dp#harder-usaco

### DIFF
--- a/content/4_Gold/Intro_DP.mdx
+++ b/content/4_Gold/Intro_DP.mdx
@@ -33,7 +33,7 @@ export const problems = {
 			'Gold',
 			'Teamwork',
 			'863',
-			'Easy',
+			'Normal',
 			false,
 			['DP'],
 			'let dp[i] -> max sum of skill levels for the first i cows, then precompute subarray maximums'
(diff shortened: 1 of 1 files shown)


frodakcin:
Looks good. 
While you're there, I find it kinda ironic the "Harder USACO" problems are labelled "Normal." Perhaps we can label the section "Additional Problems" instead.

thecodingwizard:
Can you do this problem without Floyd Warshall's? That's not introduced at this point yet...

dongliu0426:
dfs+dp instead?

thecodingwizard:
can we include a note mentioning that you can do it w/o Floyd Warshall's? I think (?) the editorial only explains the floyd warshall solution, so people may be confused?

bqi343:
I increased most of the difficulties (maybe not so accurate though). 

I think it's fine since Floyd-Warshall isn't required (and it isn't a big part of the solution). You can add a note if you want :)

PR #426, opened 2020-11-16
Add exercise to knapsack problem list

dongliu0426:


frodakcin:
Actually, it's already in the DP w/ Number Theory section.
@bqi343 opinion?

bqi343:
tbh could move most of DP NT back to knapsack and scrap that module ...

thecodingwizard:
yeah, I'm down to do that

bqi343:
Done in 8c0158b.

PR #427, opened 2020-11-16
Bitwise operators & SOS dp

frodakcin:


thecodingwizard:
Maybe we should link to that module in Bronze Complete Search with Recursion? And add a note that it's not required for bronze. https://usaco.guide/bronze/complete-rec#generating-subsets-with-bitmasks

thecodingwizard:
also is knowing how to do addition / multiplication with bitwise operators really necessary...?

frodakcin:
Sure, I just thought it was a nice way to use/check understanding of bitwise operations. Other examples (like bitmask dp) are more complex and need to be in the context of an example problem.

thecodingwizard:
okay, sounds good. I'll add a note and merge :)

PR #433, opened 2020-11-18
Implement testing Framework

MrinallU:
Finished adding end to end testing framework.


MrinallU:
@thecodingwizard You may want to make another check to run the cypress tests as an autocheck for everyone that makes a PR. I believe you can use Circle CI or some other continuous integration tool to add cypress checks for everyone that makes a PR.

thecodingwizard:
The tests don't really test our code right now... let's just put this on the backlog?

PR #442, opened 2020-11-24
typo fix from Time_Comp.mdx

amaniedu71:
not sure if it was intentional but "node" was used instead of "code."


thecodingwizard:
Thanks!

PR #443, opened 2020-11-25
Interactive & communication editorials

dolphingarlic:


dolphingarlic:
Still Stray Cat and Kanji Shiritori left but
- I don't know the solution of Kanji Shiritori
- My solution for Stray Cat is super handwavey and it shouldn't work but does

bqi343 (commented):
ok

PR #450, opened 2020-11-29
option to hide skipped problems, changed appearance of active problems

bqi343:


thecodingwizard:
Isn't this equivalent to marking problems as "ignored"?

bqi343:
Yeah. Though in that case, it might be useful to show/hide ignored problems.

thecodingwizard:
Oh yes, that seems like a good idea.

src/pages/dashboard.tsx
```diff
@@ -54,36 +54,43 @@ export default function DashboardPage(props: PageProps) {
...
         url: moduleIDToURLMap[x],
-        status:
-          userProgressOnModules[x] === 'Skipped' ? 'Skipped' : 'In Progress',
+        status: userProgressOnModules[x] as
+          | 'Skipped'
+          | 'Reading'
+          | 'Practicing'
+          | 'Ignored',
       }));
   }, [userProgressOnModules]);
```
thecodingwizard, line 81: need to also add `showIgnored` to the dependency array so that the memoized function reruns when showIgnored is changed

thecodingwizard: (I fixed this and merged)

PR #458, opened 2020-12-02
content: remove incomplete section from intro to dp module

thecodingwizard:
seems fine if it's long since we starred one of them (& have a video)?

### DIFF
--- a/content/4_Gold/Intro_DP.mdx
+++ b/content/4_Gold/Intro_DP.mdx
@@ -104,10 +104,6 @@ Although it is not too difficult to grasp the general ideas behind DP, the techn
 
 ## General Resources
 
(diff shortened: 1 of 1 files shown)


bqi343:
Have you gone through all of the resources / videos? :o

Also I don't rlly like that the table looks so long. Maybe put videos in a separate table?

thecodingwizard:
I've (very) briefly skimmed through the resources; though they overlap a lot I think it's fine since DP is somewhat difficult and some people benefit from having things explained in multiple ways?

I'm down to put the videos in a separate table though

bqi343:
alright

bqi343 (approved):

PR #462, opened 2020-12-03
add solutions to focus problems

### LINKED ISSUE #459: Point Update Range Sum
https://usaco.guide/gold/PURS

>  Add solutions to the cses problems given at the top.

dongliu0426:
closes #459


dongliu0426:
interesting, i thought it would deploy automatically

bqi343:
1-indexed segment tree should be initialized with `n+1` instead of `n` to avoid going out of the bounds of `seg` (though it still seems to pass ...)

PR #463, opened 2020-12-03
Implementations + resources for 2DRQ module

dolphingarlic:
Got rid of the `<IncompleteSection/>` tags. I just linked to an implementation of a 2D segtree


bqi343:
move persistent segtree part into the corresponding advanced module if you want

PR #470, opened 2020-12-08
Typo for Python 3 example solution "Weird Algorithm"

KingWitherBrine:
In Python 3 "/" will result in a float, so it must be "//" which will be an int.


thecodingwizard:
thanks!

Issue #453, opened 2020-11-30
Rewrite Simulation Example Code


bqi343:
https://usaco.guide/bronze/simulation

Input & Output should be consistent with IO module (https://usaco.guide/general/io)

 - C++ - IO module doesn't mention use of `FILE*`, either mention it in IO module or change in this module.
 - Java - cleaner with `Kattio`
 - Python - redirect `sys.stdin` and `sys.stdout`

bqi343:
> It would be nice if there are comments in code explaining each line of code

frodakcin:
> * Python - redirect `sys.stdin` and `sys.stdout`

Currently, I'm using Method 1 in the IO section, except without the `with ...  as ...:` operator. Is redirecting stdin & stdout better?

bqi343:
Yes.

frodakcin:
Ok I'll change it

frodakcin:
Out of curiosity, why is it better?

Is it because it's the easiest to convert between standard io and file io?

bqi343:
> Can you try to make the array names for the examples a bit more specific. It will be a little bit more convenient.

frodakcin:
Both problems?

For mixmilk, I think `c[]` for capacity and `m[]` for milk is understandable.

frodakcin:
For shell game, I think `cnt[]` for counter is fine. For the other array, which tracks which shell is at which position, I called it `ord[]`, but I'm not sure what to change it to... Possibly `shell[]` or `pos[]`?

bqi343:
>  For the other array, which tracks which shell is at which position, I called it `ord[]`, but I'm not sure what to change it to... 

Maybe `shell_at_position[]`?

frodakcin:
Ok, wordy but clear. Might as well un-abbreviate the others too.

frodakcin:
https://github.com/cpinitiative/usaco-guide/pull/509

PR #472, opened 2020-12-08
Revise sample code in simulation module

frodakcin:
https://github.com/cpinitiative/usaco-guide/issues/453


bqi343:
Shell Game C++ is missing `#include`s. Also, [`ord`](https://www.programiz.com/python-programming/methods/built-in/ord) is a builtin in Python; preferably use a different variable name.

PR #477, opened 2020-12-10
Various problems

dolphingarlic:
The difficulty might be a bit off for some of these


thecodingwizard (approved):
oops I don't know a lot of these problems, but seems fine to me?

bqi343:
Faktai uses bit operations, I wouldn't call it a bitset problem (also, it would be `Very Easy`).

PR #485, opened 2020-12-16
Editorial for POI 16 Parade

dolphingarlic:


bqi343:
> ($-1$ if $v$ is the root of the tree)

Why?

dolphingarlic:
Because the root doesn't have a parent, so |C| = degree, but for each other node, |C| = degree - 1

bqi343:
Why don't you write it in terms of degree then?

dolphingarlic:
I think I just wanted to reuse C because it's shorter. I'll change it to be in terms of degree though

PR #486, opened 2020-12-16
Update USACO open19-500.mdx

Aryansh-S:
Added a concrete solution for the math approach.


bqi343:
You should mention Kruskal's somewhere.

Aryansh-S:
> You should mention Kruskal's somewhere.

It's already mentioned in your solution.

bqi343:
Yeah, but this solution is essentially just performing Kruskal's algorithm as well.

Aryansh-S:
> Yeah, but this solution is essentially just performing Kruskal's algorithm as well.

Good point, added to the end

PR #488, opened 2020-12-16
content: add diagrams for buy low sell high

thecodingwizard:
oops I'm not sure how to make graphs; are these alright or should I change them?


bqi343:
nice

Issue #386, opened 2020-11-08
Copying Problem URLs to Clipboard


bqi343:
When accessed from the dashboard, you get smth like the following:

https://usaco.guide/bronze/complete-rec/#problem-http://www.usaco.org/index.php?page=viewproblem2&cpid=857 

There should be a way to obtain such a link from the module directly.

MrinallU:
@thecodingwizard How do you guys store the completion status of the problems, as someone could probably add a new row to the table and store the current date using the same storage process as the problem status.

thecodingwizard:
It's stored as a key-value pair where key = problem ID and value = problem status; it's certainly possible to store the current date too but I'm not sure how DB migrations will work, and I'm also not sure how to add a date into the UI. (Also adding a date is for issue #385, not this issue)

MrinallU:
@thecodingwizard Oops, I typed in the wrong issue lol.

Hanaffi:
@thecodingwizard  Can you assign it to me?

thecodingwizard:
@Hanaffi finish your other issue first (FAQ Page)! ;)

bqi343:
What would this look like? (Double click on a row of the table to display or copy a URL to clipboard?)

thecodingwizard:
maybe hover? (like what the heading tags are right now). On mobile, perhaps add another column at the very end? not sure

MrinallU:
@thecodingwizard Where in the dashboard would this be as there is only a place for the modules?

bqi343:
A bit confused, what does this have to do with modifying the dashboard?

MrinallU:
I'm kinda confused here, I thought you guys said the problem links should be displayed from the dashboard.

bqi343:
> There should be a way to obtain such a link from the module directly.

MrinallU:
@thecodingwizard This I can work on this if you want me to.

MrinallU:
So did you want a hover effect or another row on the table which displays the problems?

thecodingwizard:
hover effect on desktop, another row on mobile?

MrinallU:
So like how would the hober work here, like just a simple message that they can  view and how would they be able to copy  paste the hover message as it disappears after they take their mouse away from the problem.

thecodingwizard:
When you hover over a header you can see a link:

[...]

PR #491, opened 2020-12-17
+ extra col for URL

bqi343:
#386 smth like this, but idk how to get a link to appear to one side of the table if hovered over ...


thecodingwizard:
okay sure, this seems fine (adding the hover effect seems like a lot of work LOL)

thecodingwizard (approved):

PR #492, opened 2020-12-18
Clean up example code

jeffarjeffar:
In my opinion, the new example code is much easier to read


bqi343:
what's wrong with `tie`? D:

PR #493, opened 2020-12-18
USACO problem tables

bqi343:
See `/general/usaco-bronze`.

Todo:
 - Allow solution / tags / difficulty (?) for USACO problems to be added without problem having to be part of a module.
   - Ex. sprinklers platinum may be an application of prefix sums, but maybe we don't actually want it to appear in the silver prefix sums module since it's too hard ...
 - Headers for seasons (ex. 2019-20) should appear as normal markdown headers (##)


thecodingwizard:
> Allow solution / tags / difficulty (?) for USACO problems to be added without problem having to be part of a module.

Is this referring to the problem search?

bqi343:
I was referring to the tables I added :P

thecodingwizard:
ah I see. Would problems with a module associated with them inherit difficulty / tags & stuff from the module?

bqi343:
> ah I see. Would problems with a module associated with them inherit difficulty / tags & stuff from the module?

Yeah.

thecodingwizard:
hmm, why not make the problems w/ no associated modules part of the Bronze/Silver/Gold/Plat modules? :O (export the problems without using them anywhere)

thecodingwizard (approved):

bqi343:
Copied over some parts from `MarkCompleteButton` and `ProblemStatusCheckbox`, maybe messed up some of the CSS

thecodingwizard:
image

The title 2016-2017 is no longer needed right?

thecodingwizard (approved):

PR #494, opened 2020-12-18
Create CF Santa's Bot.mdx

jeffarjeffar:
Created solution for problem 1279D (Santa's bot)

### DIFF
--- a/solutions/CF Santa's Bot.mdx
+++ b/solutions/CF Santa's Bot.mdx
@@ -1,0 +1,76 @@
+---
+id: codeforces-1279
+source: CF
+title: Santa's Bot
+author: David Lee
+---
+
+[Official Analysis](https://codeforces.com/blog/entry/72577)
+
+<LanguageSection>
+
+<CPPSection>
+
+**Time Complexity:** $\mathcal{O}(\displaystyle\sum \limits_{i = 1}^{n} k_i)$
+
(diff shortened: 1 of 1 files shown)


bqi343:
You need to link the solution in the module as described [here](https://usaco.guide/general/contributing#adding-a-solution) (though isn't there an editorial already?)

Also, prefix the id with `cf` rather than `codeforces` (like the other solutions).

jeffarjeffar:
> You need to link the solution in the module as described [here](https://usaco.guide/general/contributing#adding-a-solution) (though isn't there an editorial already?)
> 
> Also, prefix the id with `cf` rather than `codeforces` (like the other solutions).

Yeah, there is a codeforces editorial, but no internal sol.
Is an internal solution preferred?

bqi343:
> > You need to link the solution in the module as described [here](https://usaco.guide/general/contributing#adding-a-solution) (though isn't there an editorial already?)
> > Also, prefix the id with `cf` rather than `codeforces` (like the other solutions).
> 
> Yeah, there is a codeforces editorial, but no internal sol.
> Is an internal solution preferred?

No, if the CF editorial is already okay (though it looks like it has some typos).

thecodingwizard (changes requested):
> You need to link the solution in the module as described here (though isn't there an editorial already?)
> 
> Also, prefix the id with cf rather than codeforces (like the other solutions).

PR #496, opened 2020-12-19
Edit solution to Road Reparation

dongliu0426:
DSU should be initialized with size n+1 instead of n because its 1 indexed


dongliu0426:
for some reason it still ACs the test cases

bqi343:
> for some reason it still ACs the test cases

https://stackoverflow.com/questions/16620222/vector-going-out-of-bounds-without-giving-error

PR #498, opened 2020-12-19
Create JOI18-Snake-Escaping.mdx

KWang31:
Editorial to [(https://usaco.guide/adv/dp-sos#problem-https://oj.uz/problem/view/JOI18_snake_escaping)]

### DIFF
--- a/solutions/JOI18-Snake-Escaping.mdx
+++ b/solutions/JOI18-Snake-Escaping.mdx
@@ -6,7 +6,7 @@ title: JOI 2018 Snake Escaping
 author: Kai Wang
 ---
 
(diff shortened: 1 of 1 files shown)


bqi343:
> You need to link the solution in the module as described [here](https://usaco.guide/general/contributing#adding-a-solution)

KWang31:
> > You need to link the solution in the module as described [here](https://usaco.guide/general/contributing#adding-a-solution)

Can you check what I'm doing now is right?

thecodingwizard:
Not quite -- you need to create one new file (the solution file, which you've done) and you need to edit the existing module file (https://github.com/cpinitiative/usaco-guide/blob/master/content/6_Advanced/DP_SOS.mdx) and link the solution file to it. lmk if this doesn't make sense

PR #500, opened 2020-12-19
display percentages in DivisionList, hide difficulty

bqi343:
pls fix table styling thanks

also settings modal checkboxes for the fields I added don't seem to work, is there something else I need to change?


thecodingwizard (approved):
looks good to me; I'm assuming percent = percent who got 100 in that contest?

(Also, why is platinum excluded?)

PR #520, opened 2020-12-23
Solutions point upd without template

Farrius:
Except mega inversions, the solutions are basically benq code translated (all solutions had been tested).  By the way, I think benq code using templates (but not copying the actual 200 hundred lines template) for gold and platinum is fine because you can search benq template and figure out what does every part of the code. But since I did the problems with clean c++ (I think is clean lol) and people complain for everything, here are the solutions without templates for point update range sum in gold.


thecodingwizard:
Lol rip templates :P

@bqi343 @Farrius do you guys think it's worth leaving code with template (so just have two code blocks) for people who prefer it?

bqi343:
> Lol rip templates :P
> 
> @bqi343 @Farrius do you guys think it's worth leaving code with template (so just have two code blocks) for people who prefer it?

Sure (not really because of the macros, but because I prefer the spacing of my original solutions :P).

Farrius:
Sure, but I think it is better to put the one without template before the other. I remember debugging for an hour because I was lazy to check that remDup also removes duplicates, 

> Lol rip templates :P
> 
> @bqi343 @Farrius do you guys think it's worth leaving code with template (so just have two code blocks) for people who prefer it?



> Lol rip templates :P
> 
> @bqi343 @Farrius do you guys think it's worth leaving code with template (so just have two code blocks) for people who prefer it?



> Lol rip templates :P
> 
> @bqi343 @Farrius do you guys think it's worth leaving code with template (so just have two code blocks) for people who prefer it?

Yep, that's better.

Farrius:
> Sure (not because I'm so attached to the macros, but because I prefer the spacing of my original solutions :P). It's not a big deal either way.

Don't mess with my spacing :P

Farrius:
How do you want to put the two blocks of code, just like it was two different solutions?

thecodingwizard:
Just two different blocks of code probably?

Farrius:
Ok, like this is fine ?

thecodingwizard:
Maybe add a header or a note above each code block so they know why there's two? like "implementation" / "Benq's Implementation" or something idk

Farrius:
I was think about adding a header too but since the author is Benq, if we put Benq implementation is a little strange.

thecodingwizard:
In that case, maybe "Implementation (No Templates)" and "Implemenation (Wtih Templates)" or something like that?

Farrius:
I don't know how to not commit just one file, to avoid the conflict. If someone can fix this, I don't want to break something :(

PR #521, opened 2020-12-23
start integrate liveupdate

bqi343:


src/components/MarkdownLayout/TableOfContents/getGithubUrl.tsx
```diff
@@ -4,18 +4,35 @@ import { graphql, useStaticQuery } from 'gatsby';
...
     // /\d_.*/.test(relativePath)
-    githubUrl += 'content/' + relativePath;
+    suffix = 'content/' + relativePath;
   } else {
-    githubUrl += 'solutions/' + relativePath;
+    suffix = 'solutions/' + relativePath;
   }
-  githubUrl = githubUrl.replace(' ', '%20'); // is there some function that can automatically escape chars?
+
+  const escapeSpaces = url => url.replace(' ', '%20');
```
thecodingwizard, line 23: I think this is what you're looking for: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

src/components/MarkdownLayout/TableOfContents/getGithubUrl.tsx
```diff
@@ -4,18 +4,35 @@ import { graphql, useStaticQuery } from 'gatsby';
...
+
+  // https://www.scrapingbee.com/blog/web-scraping-javascript/
+  // https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
+  const getContents = async () => {
+    const data = await fetch(rawGithubUrl);
+    const text = await data.text();
+    console.log('TEXT', text); // ok what do I do with this
+  };
+
+  getContents();
```
thecodingwizard, line 36: maybe move this to LiveUpdate? And in the `LinksToEdit` component, rather than use an OutboundLink, use a regular internal Gatsby Link pointing to `/liveupdate`, passing in a state of the Github URL: https://www.gatsbyjs.com/docs/location-data-from-props/#example-of-providing-state-to-a-link-component

And then in the LiveUpdate component, based on the incoming state, fetch the data from Github and populate the content of the editor?

src/components/MarkdownLayout/TableOfContents/getGithubUrl.tsx
bqi343: Yeah, `encodeURI` works.

bqi343:
@thecodingwizard tbh I'm not totally sure how to work with React Async / I don't completely understand how liveupdate works. I assume you can take it from here? Really laggy with edits rn (I assume the async function is being called multiple times?) Also doesn't persist on a reload.

pretty sure this line (in liveupdate) is bad ...

```
  const suffix = state && 'suffix' in state ? state.suffix : null; // oops is there a better way to do this
```

Also, is there any way to remove the error here?

```
                        if (editor.current !== undefined)
                          editor.current.layout();
```

bqi343:
Include a link to `https://github.com/cpinitiative/usaco-guide/blob/master/[corresponding suffix]` within liveupdate.

thecodingwizard:
@bqi343 do you think the navbar is needed (other than the settings menu)? Or should I get rid of the navbar?

image

bqi343:
I don't think so, other than dashboard / settings / link back to https://usaco.guide/general/contributing/#adding-a-solution

thecodingwizard:
@bqi343 Also we can link to an edit link (https://github.com/cpinitiative/usaco-guide/edit/master/content/4_Gold/Modular.mdx) by replacing blob with edit -- is this preferable? (you wouldn't have to fork the repo as described here https://usaco.guide/general/contributing#adding-a-solution ; github would do it for you automatically I think)

bqi343:
I think you need to be logged in for that to work.

bqi343:
@thecodingwizard you can merge if you think it's complete

PR #525, opened 2020-12-24
Copying URL to Clipboard

MrinallU:
Draft PR


MrinallU:
@thecodingwizard So the copy and paste stuff works, however, I have a problem with the tooltip when clicking to copy to the clipboard the tooltip disappears (have a look in the preview) any ideas?

MrinallU:
@thecodingwizard Another problem is getting the current page url. right now I am using window.href.location and that works in the development server, and when using the console in the actual guide it works there as well, however when getting put through the vercel tests it fails during the server side tests the window object does not exist thus giving it a error if you know any other alternatives to getting the page url, let me know.

thecodingwizard:
@MrinallU this PR seems like it's a combination of both the problem feedback model and the copy URL; can you change it so that it just has the copy URL changes?

You can't use window.href.location when doing server side rendering; either find a way to not use it or wrap it in React.useEffect()

(See https://www.gatsbyjs.com/docs/debugging-html-builds/)

MrinallU:
@thecodingwizard Yeah I see that now, I will probably make a new branch now.

thecodingwizard:
closing for now (just make a new PR when you fix the history :)

MrinallU:
> @MrinallU this PR seems like it's a combination of both the problem feedback model and the copy URL; can you change it so that it just has the copy URL changes?
> 
> You can't use window.href.location when doing server side rendering; either find a way to not use it or wrap it in React.useEffect()
> 
> (See https://www.gatsbyjs.com/docs/debugging-html-builds/)

Thanks you!

PR #535, opened 2020-12-29
Add a Java testing script

NewoIsTaken:
Hi everyone!

I have added an interactive Java testing script. To use it, you need to put the script, the `.java` file and the input and output files (`1.in`, `1.out`, etc.) in one directory.

Feel free to reach out if you have any changes you would like me to make.

Thanks!

PR #537, opened 2020-12-29
Add installation instructions for Java

NewoIsTaken:
Added installation instructions for Linux and macOS. Added more instructions for Windows.

### DIFF
--- a/content/1_General/Running_Code_Locally.mdx
+++ b/content/1_General/Running_Code_Locally.mdx
@@ -345,7 +345,7 @@ java Main
 
 If everything goes accordingly, the console should output `Hello World!`.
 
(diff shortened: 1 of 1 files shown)


NewoIsTaken:
This pull request depends on #535.

bqi343:
Why link to that Github specifically rather than https://www.oracle.com/java/technologies/javase-jdk15-downloads.html ?

NewoIsTaken:
@bqi343 I linked it to the AdoptOpenJDK OpenJDK binaries. The main difference between Oracle Java and OpenJDK is the OpenJDK is licensed under the `GNU General Public License (GNU GPL) version 2 with a linking exception` while Oracle Java is licensed under the `Oracle Binary Code License Agreement`. This really only affects business, commercial, or production use but OpenJDK can be used freely. So, I linked OpenJDK.

PR #540, opened 2020-12-30
content: add p vs np video

thecodingwizard:
I thought this video was kinda cool, but I'm not sure where to put it? hopefully this is fine (I left a comment stating that it's optional)


bqi343:
should probably surround it in an optional block

thecodingwizard:
Like this?

bqi343:
That looks fine (except I would do `[Youtube video](youtube link...)`)

PR #544, opened 2020-12-30
Explanation for hack case

HarryWangATX:
If it's not good just thumb this message down ;).


content/4_Gold/SP.mdx
```diff
@@ -358,7 +358,7 @@ public class test {
 
 <Warning>
 
-It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem!
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! Since the edge count on the last testcase is greater than the node count, multiple nodes will point to the same node. Without `continue`, "duplicates" of the same node with different values might get added to the priority queue. This will cause a TLE on the last testcase because now the program has to traverse the same nodes multiple times which is suboptimal. Furthermore, because we only want the shortest path, there is no point in keeping these "duplicates", we only have to keep the minimum amount. As a result, we can `continue` if we get to a value that is more that the shortest at that node and significantly reduce the runtime.
```
bqi343, line 361: The number of edges is almost always greater than the number of nodes ...

bqi343: Processing the same node multiple times is fine if that node has small degree.

bqi343 (changes requested):

content/4_Gold/SP.mdx
bqi343: Should also mention what the test case is and that the running time is worse than N^2.

HarryWangATX: @bqi343 okay i changed the whole explanation

dolphingarlic:
I think you should split the explanation into separate paragraphs to make it a bit easier to read

content/4_Gold/SP.mdx
```diff
@@ -358,7 +358,11 @@ public class test {
 
 <Warning>
 
-It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem!
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
+
+The last testcase contains 100000 destinations and 149997 routes. City 1 has around 50000 routes to cities 2-50000. Cities 2-50000 have routes that go to City 50001. City 50001 has routes that go to Cities 50002-100000. Without the `continue`, after the program processes the routes from City 1, the priority queue will contain 49998 routes that point to City 50001. After getting the shortest route to City 50001 from these routes, it will move on to the cities 50002-100000 that can be reached from City 50001. However, the priority queue now has 49998 instances of these routes. As you can see, this runtime is worse than $\mathcal{O}(N^2)$, which results in a TLE.
```
dolphingarlic, line 363: Why are random words capitalized in this paragraph?

content/4_Gold/SP.mdx
```diff
@@ -358,7 +358,11 @@ public class test {
...
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
+
+The last testcase contains 100000 destinations and 149997 routes. City 1 has around 50000 routes to cities 2-50000. Cities 2-50000 have routes that go to City 50001. City 50001 has routes that go to Cities 50002-100000. Without the `continue`, after the program processes the routes from City 1, the priority queue will contain 49998 routes that point to City 50001. After getting the shortest route to City 50001 from these routes, it will move on to the cities 50002-100000 that can be reached from City 50001. However, the priority queue now has 49998 instances of these routes. As you can see, this runtime is worse than $\mathcal{O}(N^2)$, which results in a TLE. 
+
+On the other hand, if we did include the `continue`, after setting the cost of City 50001, the program will just pop off the other 49997 routes that does not have the shortest path. Thus, it now only visits every city once.
```
dolphingarlic, line 365: I think it's better to say that we *process* each city once

content/4_Gold/SP.mdx
dolphingarlic, line 363: Also, rather say stuff like "cities 2 to 50000"

dolphingarlic, line 363: > As you can see

It's not immediately obvious why, so many people might not see. I think you should include something like "this means that city 50001 will be processed 49998 more times"

dolphingarlic (changes requested):

content/4_Gold/SP.mdx
```diff
@@ -358,13 +358,11 @@ public class test {
...
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
 
-<IncompleteSection>
+The last testcase contains 100000 destinations and 149997 routes. City 1 has around 50000 routes to cities 2 through 50000. Cities 2 through 50000 have routes that go to city 50001. City 50001 has routes that go to cities 50002 through 100000. Without the `continue`, after the program processes the routes from city 1, the priority queue will contain 49998 routes that point to city 50001. After getting the shortest route to city 50001 from these routes, it will move on to the cities 50002 through 100000 that can be reached from City 50001. However, the priority queue now has 49998 instances of these routes. This means that city 50001 will be processed 49998 more times and so on for the 49998 cities that branches out from city 50001. This results in a worse runtime than $\mathcal{O}(N^2)$, which will TLE.
```
bqi343, line 363: Shouldn't 49998 be 49999?

bqi343: > after the program processes the routes from city 1

shouldn't this be - after the program pops cities 1 - 50000 off the queue?

bqi343: > However, the priority queue now has 49998 instances of these routes.

which routes?

bqi343: > cities that branches out

huh

bqi343: > This results in a worse runtime than $\mathcal{O}(N^2)$, which will TLE. 

I guess you should say $\Theta(N^2\log N)$ instead.

bqi343: > City 1 has around 50000 routes to cities 2 through 50000

Why not "City 1 has flights to cities 2 through 50000"? Shouldn't use "route" and "flight" interchangeably.

bqi343 (changes requested):

content/4_Gold/SP.mdx
```diff
@@ -358,13 +358,11 @@ public class test {
...
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
 
-<IncompleteSection>
+The last testcase contains 100000 destinations and 149997 routes. City 1 has around 50000 flights to cities 2 through 50000. Cities 2 through 50000 have flights that go to city 50001. City 50001 has flights that go to cities 50002 through 100000. Without the `continue`, after the program pops off cities 1 through 50000, the priority queue will contain 49999 routes that point to city 50001. After getting the shortest route to city 50001 from these routes, it will move on to the cities 50002 through 100000 that have flights from city 50001. This means that city 50001 will be processed 49998 more times than needed; furthermore, it carries over to the 49999 other flights from city 50001 to cities 50002-100000. This results in a worse runtime than $\Theta(N^2\log N)$, which will TLE.
```
thecodingwizard, line 363: "149997 routes" ==> "149997 flights" (?)

HarryWangATX: yes, ill fix that

content/4_Gold/SP.mdx
```diff
@@ -358,13 +358,11 @@ public class test {
...
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
 
-<IncompleteSection>
+The last testcase contains 100000 destinations and 149997 flights. City 1 has around 50000 flights to cities 2 through 50000. Cities 2 through 50000 have flights that go to city 50001. City 50001 has flights that go to cities 50002 through 100000. Without the `continue`, after the program pops off cities 1 through 50000, the priority queue will contain 49999 flights that point to city 50001. After getting the shortest cost to city 50001 from these flights, it will move on to the cities 50002 through 100000 that have flights from city 50001. This means that city 50001 will be processed 49998 more times than needed; furthermore, it carries over to the 49999 other flights from city 50001 to cities 50002-100000. This results in a worse runtime than $\Theta(N^2\log N)$, which will TLE.
```
bqi343, line 363: Still need to address some of my comments. Also:

> the priority queue will contain 49999 flights that point to city 50001

should this be "routes that end at city 50001"?

> it carries over to the 49999 other flights from city 50001 to cities 50002-100000.

What exactly does this mean? Cities 50002-100000 aren't pushed to the queue more than once.

> This results in a worse runtime than $\Theta(N^2\log N)$

Well, a runtime of $\Theta(N^2\log N)$, not worse than.

bqi343 (changes requested):

content/4_Gold/SP.mdx
HarryWangATX: @bqi343 thx, ill fix sry i misread some of the previous comments

HarryWangATX: > Cities 50002-100000 aren't pushed to the queue more than once.
I was saying because the instances of city 50001 don't get popped off, those instances get carried on and all of the city 50001s will then add the adjacent cities 50002-100000.

content/4_Gold/SP.mdx
```diff
@@ -358,13 +358,11 @@ public class test {
...
+It's important to include `continue`. This ensures that when all edge weights are non-negative, we will never go through the adjacency list of any vertex more than once. Removing it will cause TLE on the last test of the above problem! 
 
-<IncompleteSection>
+The last testcase contains 100000 destinations and 149997 flights. City 1 has around 50000 flights to cities 2 through 50000. Cities 2 through 50000 have flights that go to city 50001. City 50001 has flights that go to cities 50002 through 100000. Without the `continue`, after the program pops cities 1 through 50000 off the queue, the priority queue will contain 49999 routes that end at city 50001. After getting the shortest cost to city 50001 from these flights, it will move on to the cities 50002 through 100000 that have flights from city 50001. This means that city 50001 will be processed 49998 more times than needed (deleted the part but maybe i should rephrase it?). This results in a runtime of $\Theta(N^2\log N)$, which will TLE.
```
bqi343, line 363: remove "around 50000"

bqi343: use "flights to" rather than "flights that go to"

bqi343: > After getting the shortest cost to city 50001 from these flights

routes, not flights

bqi343: > This means that city 50001 will be processed 49998 more times than needed (deleted the part but maybe i should rephrase it?).

What about - Every one of the 49999 times city 50001 is popped off the queue and processed, we must iterate over all of its outgoing flights (cities 50002 - 10^5).

bqi343 (changes requested):

PR #552, opened 2020-12-31
Got rid of unnecessary lines

HarryWangATX:


dolphingarlic:
Lol nice catch

dolphingarlic:
Oops I didn't read the code carefully - the lines aren't actually unnecessary

dolphingarlic:
Basically, those lines compare the second dimension when the first dimensions are equal (so that the cows are sorted by both dimensions instead of only by the first and then randomly by the second)

HarryWangATX:
@dolphingarlic I submitted my code again and it passed. but idk because i think we only need to care about x or y values

HarryWangATX:
cus idt the second dimensuon matters as it will be checked with the if statements

PR #566, opened 2021-01-02
CSES Subarray Divisibility, CPP sol

HarryWangATX:


bqi343:
Looks like there's a mistake when I render with liveupdate. Also,

> I think you should split the explanation into separate paragraphs to make it a bit easier to read

This includes putting some of the equations on separate lines:

```
$$
equation
$$
```

Also use $\cdot$ instead of $*$, and use $\texttt{}$ around variable names.

HarryWangATX:
@bqi343 do you know why the links are failing? I don't think I changed any links.

bqi343:
Try pasting it into liveupdate.

HarryWangATX:
K i saw an rendering error

HarryWangATX:
@bqi343 i think i fixed everything, lmk if I need to edit more.

HarryWangATX:
looks nice!

PR #567, opened 2021-01-02
Add line #define endl "\n" to make it optimized

tachithanhdanh:
Sorry I forgot to redefine the endl after clearing the macros

### DIFF
--- a/solutions/CSES Concert Tickets.mdx
+++ b/solutions/CSES Concert Tickets.mdx
@@ -18,8 +18,6 @@ Another way of solving the problem is to use the `multiset::upper_bound` functio
  
 using namespace std;
 
-#define endl "\n"
-
 //variables used for the current problem
 int n,m,h,t; multiset<int> tickets;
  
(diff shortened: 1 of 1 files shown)


thecodingwizard:
This seems extremely sketchy :eyes: I'm worried it might confuse newcomers (or inspire questionable practices like `#define int long long`)? idk

Wouldn't it be better to just replace `endl` with `\n`? (Or if `endl` passes in time just leave it as is?)

tachithanhdanh:
I found this may be useful for them
https://usaco.guide/general/fast-io/#fast-output
https://usaco.guide/general/generic/#macros
Should I add some more explanations too?

tachithanhdanh:
It seems that endl makes the program run significantly slower than "\n", espcially on this current problem
Using endl only: https://imgur.com/a/QBTlNMj
After redirecting the endl into "\n" by using #define: https://imgur.com/a/p4kBxsk

tachithanhdanh:
> IDK what do other people think? Is this a good macro to promote?
Personally, I prefer endl to "\n" because it helps me to type faster since the quotation marks are harder to reach (you have to press 2 buttons simultaneously which is annoying to me).

HarryWangATX:
> IDK what do other people think? Is this a good macro to promote?

does seem a little sketchy to me. I do #define nl "\n" in my code for that.

PR #569, opened 2021-01-02
Share url + lang

### LINKED ISSUE #495: Sharing URL + Language
Ex. say I want to share a URL that refers to a section that is only present in the java version of some module.

bqi343:
closes #495

### DIFF
--- a/gatsby-browser.tsx
+++ b/gatsby-browser.tsx
@@ -59,8 +59,3 @@ export const onClientEntry = () => {
     event.preventDefault();
   });
 };
-
-export const onRouteUpdate = ({ location, prevLocation }) => {
-  console.log('new pathname', location.pathname);
-  console.log('old pathname', prevLocation ? prevLocation.pathname : null);
-};
--- a/gatsby-browser.tsx
+++ b/gatsby-browser.tsx
@@ -59,3 +59,8 @@ export const onClientEntry = () => {
     event.preventDefault();
   });
 };
+
+export const onRouteUpdate = ({ location, prevLocation }) => {
+  console.log('new pathname', location.pathname);
+  console.log('old pathname', prevLocation ? prevLocation.pathname : null);
+};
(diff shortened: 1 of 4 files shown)


thecodingwizard:
looks like the query param isn't being persisted when the URL changes -- a quick hack could be to use https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#onRouteUpdate

This would cause even non-module/solution routes to have the query param though :/

bqi343:
When I'm logged in, everything has a query param.

> Or maybe the query param should only affect that immediate page, and clicking a different module link would revert the language back to the user setting (this seems reasonable/easier?)

seems reasonable

gatsby-browser.tsx
```diff
@@ -59,3 +59,8 @@ export const onClientEntry = () => {
     event.preventDefault();
   });
 };
+
+export const onRouteUpdate = ({ location, prevLocation }) => {
```
bqi343, line 63: @thecodingwizard is this supposed to print anything to console?

src/components/MarkdownLayout/MarkdownLayout.tsx
```diff
@@ -64,9 +64,15 @@ export default function MarkdownLayout({
...
-  );
+  const {
+    userProgressOnModules,
+    setModuleProgress,
+    setLang,
+    lang,
+  } = useContext(UserDataContext);
+  React.useEffect(() => {
+    setLang(lang);
+  }, []);
```
bqi343, line 75: oops when the same module is clicked twice the lang disappears D:

gatsby-browser.tsx
thecodingwizard: yes, though you might need to restart dev server (?)

thecodingwizard:
remaining issues:

- clicking a module link twice causes the lang parameter to disappear (let's just pretend this is a feature...)
- multiple tabs open at the same time can cause issues with firebase synchronization (let's just pretend this bug doesn't exist; if it does happen it's just a minor inconvenience to switch the language anyway)

thecodingwizard (approved):
this is probably ready to merge, though I would probably test this a bit more ://

bqi343:
> clicking a module link twice causes the lang parameter to disappear (let's just pretend this is a feature...)

@thecodingwizard do you know how to fix this?

thecodingwizard:
Possibly by setting the language on every rerender (but idk how that will affect performance). But I think it can be considered a feature -- it provides a somewhat simple way to get rid of the lang query param in case the user wants to share the module link only without sharing the language.

PR #573, opened 2021-01-03
Added solution for Packmen and Magic Ship

SansPapyrus683:
~~So far there's only a Java sol for Magic Ship, but I can probably code up a C++ one if it's needed.~~
EDIT: Added solutions in all 3 languages for both problems


frodakcin:
A few comments:

* Both problems: Capitalize your name. 
* Both problems: link to solution from problem list
* Magic Ship Explanation: "since we can just move in the opposite direction of the wind" is vague. Use something similar to the editorial phrasing "since we can stay in the destination point by moving to the opposite of the wind direction"
  * I think analysis on impossible case is useful as well
* Packmen Explanation: I'd prefer a better proof/explanation of greedily assigning leftmost packman to leftmost pellet. Maybe reference [this problem](https://www.facebook.com/codingcompetitions/hacker-cup/2020/round-1/problems/B/solution)?
* Packmen CPP version: I'd prefer if you used references (`const vector<>&` or `vector<>&`). It's good practice. (Also means that `food = std::deque<int>(food);  // make a copy of the food deque` isn't redundant)
* Packmen Python version: delete "idk" from comments (you do know!) and also please add similar comments to cpp/java as well.

SansPapyrus683:
Thanks for the feedback! As for the Packmen CPP comment, I don't really get what you mean. Are you saying that instead of ```cpp
bool all_eatable(vector<int> packmen, std::deque<int> food, int time)
```, I should instead have the arguments passed by reference?

**Comment by frodakcin** [comment]
Yes

**Comment by frodakcin** [comment]
I should have been more specific; my bad

**Comment by SansPapyrus683** [comment]
> Both problems: link to solution from problem list
This is already in the Magic Ship solution, and for Packmen I couldn't seem to find it in the Contest Materials. There was only a pdf with the problem statements.

**Comment by frodakcin** [comment]
I meant in the problems page. (here)[https://usaco-guide-omj7ua89e.vercel.app/problems/]

Some of the problems say "View Solution" under "View Problem Statement"

**Comment by frodakcin** [comment]
[Link](https://usaco-guide-omj7ua89e.vercel.app/problems/). Apparently I cannot properly use md

**Comment by SansPapyrus683** [comment]
How would I go about doing that?

**Comment by frodakcin** [comment]
I don't know; @thecodingwizard, help?

**Comment by thecodingwizard** [comment]
uhh I think the website will auto-link the solution after it's merged into master. As long as the "Internal Sol" link shows up properly on the module page, the problems page will auto-update when it's merged into master.

**Comment by frodakcin** [comment]
oh ok, should be all good them, lemme check it one last time

**Comment by frodakcin** [comment]
then*

File: solutions/CF Packmen.mdx [hunk 1]
```diff
@@ -0,0 +1,295 @@
+---
+id: cf-packmen
+source: CF
+title: Packmen
+author: Kevin Sheng
+---
+
+(No official editorial)
+
+Binary search on the minimum time it will take for all the Packmen to eat all the pellets.
```
dolphingarlic, line 10: Do you mean to add additional newlines between these lines to separate them into paragraphs?

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,295 @@
...
+For each Packman, we can greedily calculate the maximum amount of pellets that it can eat in the given amount of time and then remove those pellets.
+However, each Packman must eat all of the uneaten food to its left, or else all the food can't be eaten in the given time. Assigning other Packmen to the uneaten pellets would be pointless, because they would be farther away and thus have no way of reaching the uneatable pellet.
+
+Note that there are two main cases: One where the Packman first goes to the right then to the left, and another where the Packman first goes to the *left* then the right. We calculate the food eaten in both of these cases and take the maximum. Note that there is also a special case where a Packman doesn't have any pellets to the left, so it can go to the right, eating as many pellets as time allows.
+
+**Time Complexity:** $\mathcal{O}(N\log N)$
```
dolphingarlic, line 16: Move this to the top of the editorial

dolphingarlic (changes requested):

solutions/CF Packmen.mdx
SansPapyrus683: Just the "two main cases" part of the time complexity?

solutions/CF Magic Ship.mdx
```diff
@@ -0,0 +1,115 @@
+---
+id: cf-magship
+source: CF
+title: Magic Ship
+author: Kevin Sheng
+---
+
+[Official Editorial](https://codeforces.com/blog/entry/65365)
```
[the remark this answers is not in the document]
SansPapyrus683: Ok- are you implying I should remove it?

solutions/CF Packmen.mdx
dolphingarlic: Just the time complexity line

solutions/CF Packmen.mdx
SansPapyrus683: I guess?

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,295 @@
...
+source: CF
+title: Packmen
+author: Kevin Sheng
+---
+
+(No official editorial)
+
+Binary search on the minimum time it will take for all the Packmen to eat all the pellets.
+For each Packman, we can greedily calculate the maximum number of pellets that it can eat in the given amount of time and then remove those pellets.
+However, each Packman must eat all of the uneaten food to its left, or else all the food can't be eaten in the given time. Assigning other Packmen to the uneaten pellets would be pointless, because they would be farther away and thus have no way of reaching the uneatable pellet.
```
frodakcin, line 12: Still quite unclear in my opinion.

First sentence can be "For each Packman, we can greedily assign the pellets it will eat"

The second part needs to be proof-like (to a certain extent) and articulate. I know this is difficult to do, but I think it's really beneficial. The way that FHC did it isn't optimal either, but it much more rigorous.

Key points for one possible explanation (similar to FHC):
 * Time only depends on leftmost and rightmost pellets. Therefore, every packman will eat a contiguous, disjoint subsegment
 * Prove that it is never less beneficial to assign packmen to subsegments in sorted order (with care, there are examples of subsegments that require packmen to be assigned in un-sorted order, but these can always be modified to work with sorted order)

frodakcin (changes requested):

solutions/CF Packmen.mdx
SansPapyrus683: What's FHC?

frodakcin: facebook hacker cup. Here's [link](https://www.facebook.com/codingcompetitions/hacker-cup/2020/round-1/problems/B/solution)

SansPapyrus683: >  there are examples of subsegments that require packmen to be assigned in un-sorted order

Could I have a small example?

solutions/CF Magic Ship.mdx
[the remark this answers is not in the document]
dolphingarlic: Yeah pretty much (unless the editorial isn't good? I haven't read it)

SansPapyrus683:
oops i meant python and *cpp* sols, not java sols

solutions/CF Magic Ship.mdx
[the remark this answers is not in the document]
SansPapyrus683: To be honest I feel the explanation of the impossible case is a bit better in my editorial (suggested by frodakcin)

solutions/CF Packmen.mdx
frodakcin: Packmen at positions 2 and 4.
Food at positions 0, 2, 6.

Assign packman 2 -> foods at 2 and 6
Assign packman 4 -> food at 0
The minimum # of steps is 4

You cannot simply swap and say packman 4 eats 2 and 6 because that's impossible in 4 steps. However, you can allocate food 2 to packman 2 to make it doable (and in even fewer steps).

frodakcin: Maybe something along these lines? (probably in a spoiler because its wall of text) I don't know why the proof is so messy and long. If you find a cleaner version please use it. 

1) Time only depends on leftmost and rightmost pellets. Therefore, every packman will eat a contiguous, disjoint subsegment (same as before)
2) From this point on, "solution" refers to a valid assignment of packmen -> pellets that takes less than $d$ time. If there exists a solution where packmen are assigned pellets in overlapping subsegments, there exists a solution where packmen are assigned pellets in non-overlapping subsegments (assign all overlapping pellets to either packman)
3) If there exists any solution, there exists a solution where the packmen are assigned subsegments in sorted order (i.e, the $i$-th packman in sorted order is assigned the $i$-th subsegment in sorted order)

Proof 1: Two consecutive packmen are at positions $x$ and $y$, where subsegment of $x$ is to the right of subsegment $y$. The union of these subsegments is from $l$ to $r$ (and all pellets between). Assume $l \leq (x+y)/2 \leq r$. Otherwise, the claim is true by reassigning all pellets in the range to the closer packman. Since subsegment of $x$ is to the right of that of $y$, Packman at $x$ eats pellet $r$ and packman at $y$ eats pellet $l$. Thus, $max(r-x, y-l) \leq d$. We find a different pellet assignment that requires no greater than the same time $d$. Assign packman $x$ to all pellets in the range $[l, (x+y)/2]$, and packman $y$ to all pellets in the range $[(x+y)/2, r]$. Packman $x$ can travel $(x+y)/2 - x + (x+y)/2 - l = y-l$ time, and packman $y$ can travel $y-(x+y)/2+r-(x+y)/2 = r-x$ time. Thus, a solution with total time $max(y-l, r-x)$, no more than the total time $d$ of the solution, is achieved.

Proof 2: More of a thought experiment?? Number the packmen $1 \dots N$ from left to right. Simulate the packmen's movement as described in problem. However, when two packmen cross, instead switch both of their directions and not allow them to cross (similar to ant on stick problem). Notice how the overall movement and all covered regions (and thus eaten pellets) stay exactly the same. Also, this requires exactly the same amount of time. Now, packman $i$ is always to the left of packman $i+1$, so it's interval must be also to the left of packman $i+1$. 

4) Since leftmost packman must cover a leftmost subsegment, simply assign it the largest such subsegment.

SansPapyrus683: > Food at positions 0, 2, 6.

I thought all the food and packmen had to be at distinct positions (not that it really matters)

frodakcin: oh, ok. well idea is still there

solutions/CF Magic Ship.mdx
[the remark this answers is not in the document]
frodakcin: The official editorial for this problem is superb. Perhaps dolphin is right -- we should either just link or quote it.

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,293 @@
...
+source: CF
+title: Packmen
+author: Kevin Sheng
+---
+
+**Time Complexity:** $\mathcal{O}(N\log N)$
+
+Binary search on the minimum time it will take for all the Packmen to eat all the pellets.
+For each Packman, we can greedily calculate the maximum number of pellets that it can eat in the given amount of time and then remove those pellets.
+Note that it is optimal for all of the Packmen to eat continuous segments of the food, and all of the segments must be non-intersecting. They must be continuous because if they weren't, a Packman could always eat it while going from one segment to another, thus making it continuous. Also, they must be non-intersecting because if they *were*, two Packmen would target the same pellet, thus resulting in wasted time.
```
frodakcin, line 12: Maybe phrase it as
"A packman will visit all pellets between the leftmost and rightmost pellets a it eats, so it can eat all of them without any time penalty" ?

frodakcin: "Also, they must be non-intersecting because if they *were*, either packman can ignore the pellets of the other packman." might be better?

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,293 @@
...
+Each Packman must also eat all of the uneaten food to its left, or else all the food can't be eaten in the given time. Assigning other Packmen to the uneaten pellets would be pointless, because they would be farther away and thus have no way of reaching the uneatable pellet.
```
frodakcin, line 13: "or else all the food can't be eaten in the given time" isn't necessarily accurate, as is "because they would be farther away and thus have no way of reaching the uneatable pellet"

frodakcin (changes requested):

content/3_Silver/Binary_Search_Ans.mdx
```diff
@@ -105,7 +105,7 @@ export const problems = {
 			'Normal',
 			true,
 			[],
-			'binary search on time with epsilon and then find if there exist non-empty intersection of the ranges (maximum of left points and minimum of right points)'
+			''
```
dolphingarlic, line 108: I think you should leave the solution sketch for now (bad solution sketch > no solution sketch)

solutions/CF Packmen.mdx
dolphingarlic: So are you going to add those newlines? Right now, it's rendering as one big paragraph

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+There are two main cases we consider for each Packman: One where the Packman first goes to the right then to the left, and another where the Packman first goes to the *left* then the right. We calculate the food eaten in both of these cases and take the maximum. Note that there is also a special case where a Packman doesn't have any pellets to the left, so it can go to the right, eating as many pellets as time allows.
```
dolphingarlic, line 15:
```suggestion
There are two main cases we consider for each Packman: one where the Packman first goes to the right and then to the left, and another where the Packman first goes to the *left* and then to the right. We calculate the food eaten in both of these cases and take the maximum. Note that there is also a special case where a Packman doesn't have any pellets to the left, so it can go to the right, eating as many pellets as time allows.
```

dolphingarlic: Also, why is the case where there are no remaining pellets to the left a "special case"?

dolphingarlic (changes requested):
When I first read your solution, I thought that you were just calculating the maximum number of pellets that each Packman can eat without considering what the others do. Now I realize that you mean you process the Packmen from left to right and for each Packman, you 1) eat as many pellets as possible so that there are no more pellets to its left and then 2) remove the eaten pellets from a list.

I think you should make it more clear that this is what your solution is doing, as it would explain 1) why we can "greedily calculate the maximum number of pellets" and 2) why binsearching works

solutions/CF Packmen.mdx
SansPapyrus683: Oh yeah, I wasn't really thinking when I wrote that. I probably put it as a special case because I had to implement an if-statement in my own solution to account for that case.

solutions/CF Packmen.mdx
SansPapyrus683: Actually, sorry, I didn't know what you meant by that. I meant to render it as a big paragraph, yet I added newlines for ease of editing.

solutions/CF Packmen.mdx
SansPapyrus683: Yeah, I cleared that up in the new version.

bqi343:
@SansPapyrus683 you should re-request reviews

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,294 @@
...
+source: CF
+title: Packmen
+author: Kevin Sheng
+---
+
+**Time Complexity:** $\mathcal{O}(N\log N)$
+
+Binary search on the minimum time it will take for all the Packmen to eat all the pellets.
+For each Packman, we can greedily calculate the maximum number of pellets that it can eat in the given amount of time and then remove those pellets.
+Note that a Packman will visit all pellets between the leftmost and rightmost pellets a it eats, so it can eat all of them without any time penalty. Also, they must be non-intersecting because if they were, either Packman can ignore the pellets of the other Packman.
```
frodakcin, line 12: Add clarification sentence "This means that all packmen will eat contiguous segments," or "All packmen will eat a continuous range of pellets" or something along these lines.

frodakcin: Maybe phrase it like this?
```
Suppose we want to know whether the Packmen can eat all the pellets in $d$ units of time.

First, notice a few observations:
 * Each packman will eat a continuous subarray of the pellets, because a packman will visit all pellets between the leftmost and rightmost ones, and thus can eat all pellets between them with no additional time penalty.
 * No two packmen need to eat overlapping subarrays, as the region of overlap can be assigned to one of the packman and be ignored by the other
 * The leftmost packman will eat the leftmost subarray of pellets, because if this were not the case, the leftmost packman will necessarily cross over another packman. If this occurs, simply simulate (prevent) the crossing by reversing the directions of both packmen instead. This ensures the leftmost packman is always to the left of all other packmen, and thus will eat a subarray of pellets to the left of all others

We can repeatedly greedily assign the leftmost packman to the largest possible subarray of pellets that includes the leftmost pellet. Using this strategy, we can check whether all pellets can be eaten when the packmen are limited to $d$ units of time.
```

frodakcin (changes requested):
Sorry for being so strict; maybe I or someone else should have taken over this PR instead of putting so much pressure on you.

Anyway, I think we are really close.

SansPapyrus683:
> @SansPapyrus683 you should re-request reviews if you've made the requested changes

Oh ok then, sorry about that.

SansPapyrus683:
> Sorry for being so strict; maybe I or someone else should have taken over this PR instead of putting so much pressure on you.

@frodakcin No worries! I'll implement the changes as quickly as possible.

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,298 @@
...
+
+We can repeatedly greedily assign the leftmost Packman to the largest possible subarray of pellets that includes the leftmost pellet. Using this strategy, we can check whether all pellets can be eaten when the Packmen are limited to $d$ units of time.
+In the below solution, we keep a sorted deque of all the pellet positions. For each Packman, we calculate the maximum amount of pellets it can eat so that no more pellets are to the left, and then we remove all those pellets from the deque.
+However, if a Packman can't eat the leftmost uneaten pellet, then this time would be invalid.
+There are two main cases we consider for each Packman: one where the Packman first goes to the right and then to the left, and another where the Packman first goes to the *left* and then to the right. We calculate the pellet eaten in both of these cases and take the maximum.
```
frodakcin, line 20: Paragraph breaks.

I definitely suggest one before "There are two main cases," but feel free to add more to make it less wall of text.

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,298 @@
...
+ * No two Packmen need to eat overlapping subarrays, as the region of overlap can be assigned to one of the Packman and be ignored by the other.
+ * The leftmost Packman will eat the leftmost subarray of pellets, because if this were not the case, the leftmost Packman will definitely cross over another Packman. If this occurs, simply simulate (prevent) the crossing by reversing the directions of both Packmen instead. This ensures the leftmost Packman is always to the left of all other Packmen, and thus will eat a subarray of pellets to the left of all others.
+
+We can repeatedly greedily assign the leftmost Packman to the largest possible subarray of pellets that includes the leftmost pellet. Using this strategy, we can check whether all pellets can be eaten when the Packmen are limited to $d$ units of time.
+In the below solution, we keep a sorted deque of all the pellet positions. For each Packman, we calculate the maximum amount of pellets it can eat so that no more pellets are to the left, and then we remove all those pellets from the deque.
+However, if a Packman can't eat the leftmost uneaten pellet, then this time would be invalid.
```
frodakcin, line 19: "then the time $d$ would be invalid, as we have shown that given a valid time $d$ there must exist an assignment where the leftmost Packman does eat the leftmost uneaten pellet"

frodakcin (changes requested):
Minor fixes. (BTW feel free to rephrase/rewrite what I put in quotes. They're mainly guidelines)

solutions/CF Magic Ship.mdx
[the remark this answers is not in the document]
dolphingarlic: So what's the decision for this?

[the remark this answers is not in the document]
frodakcin: Link original editorial

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+We can repeatedly greedily assign the leftmost Packman to the largest possible subarray of pellets that includes the leftmost pellet. Using this strategy, we can check whether all pellets can be eaten when the Packmen are limited to $d$ units of time.
+
+In the below solution, we keep a sorted deque of all the pellet positions. For each Packman, we calculate the maximum amount of pellets it can eat so that no more pellets are to the left, and then we remove all those pellets from the deque.
+
+However, if a Packman can't eat the leftmost uneaten pellet, then the time $d$ would be invalid, as we have shown previously that given a valid time $d$ there must exist an assignment where the leftmost Packman does eat the leftmost uneaten pellet.
+
+There are two main cases we consider for each Packman: one where the Packman first goes to the right and then to the left, and another where the Packman first goes to the *left* and then to the right. We calculate the pellet eaten in both of these cases and take the maximum.
```
frodakcin, line 24: Last sentence needs grammar fix.
Maybe "We calculate the subarray of pellets eaten in both of these cases and assign the larger one to the packman."

frodakcin (changes requested):
Last fix

frodakcin (approved):
All set. Let me find someone else to double check.

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,302 @@
...
+In that case, we would have to check whether the Packmen can eat all the pellets in $d$ units of time.
+
+First, notice a few observations:
+ * Each Packman will eat a continuous subarray of the pellets, because a Packman will visit all pellets between the leftmost and rightmost ones, and thus can eat all pellets between them with no additional time penalty.
+ * No two Packmen need to eat overlapping subarrays, as the region of overlap can be assigned to one of the Packman and be ignored by the other.
+ * The leftmost Packman will eat the leftmost subarray of pellets, because if this were not the case, the leftmost Packman will definitely cross over another Packman. If this occurs, simply simulate (prevent) the crossing by reversing the directions of both Packmen instead. This ensures the leftmost Packman is always to the left of all other Packmen, and thus will eat a subarray of pellets to the left of all others.
+
+We can repeatedly greedily assign the leftmost Packman to the largest possible subarray of pellets that includes the leftmost pellet. Using this strategy, we can check whether all pellets can be eaten when the Packmen are limited to $d$ units of time.
```
dolphingarlic, line 18:
```suggestion
We can repeatedly greedily assign the leftmost unprocessed Packman to the largest possible subarray of pellets that includes the leftmost uneaten pellet. Using this strategy, we can check whether all pellets can be eaten when the Packmen are limited to $d$ units of time.
```

solutions/CF Packmen.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+In the below solution, we keep a sorted deque of all the pellet positions. For each Packman, we calculate the maximum amount of pellets it can eat so that no more pellets are to the left, and then we remove all those pellets from the deque.
```
dolphingarlic, line 20:
```suggestion
In the below solution, we keep a sorted deque of all the pellet positions. For each Packman, we calculate the maximum number of pellets it can eat so that no pellets to the left are uneaten, and then we remove the eaten pellets from the deque.
```

dolphingarlic (changes requested):
Just two small grammar/style fixes and it should be good to merge

PR #574, opened 2021-01-03
start import user data

bqi343:


thecodingwizard:
In theory this should work? But would probably test this a lot since it overrides your data permanently ://

bqi343:
> In theory this should work? But would probably test this a lot since it overrides your data permanently ://

if someone really wants to erase his data that's fine with me :)

thecodingwizard:
I don't think class imports work. @jeffreymeng?

jeffreymeng:
wdym by class imports

thecodingwizard:
Export user data then import user data -- you lose the list of classes you're enrolled in I think

jeffreymeng:
Where is user data exported?

Thanks,
Jeffrey Meng

--
Jeffrey Meng | jeffkmeng@gmail.com | jeffkmeng.com

dongliu0426:
won't this let the user override the data (ex. daily streak)?

thecodingwizard:
Yes, but they technically can already anyway. (Maybe we can add some
restrictions on streaks...)

thecodingwizard:
hmm interesting I just tried again and user class import worked (and looking at the code it should work). This code seems fine but should probably be tested a bit more (?).

thecodingwizard:
oops idk why so many files are changed, I'll move this to #706

PR #577, opened 2021-01-04
Update Greedy.mdx

tachithanhdanh:

PR #578, opened 2021-01-04
Change the solution link for Range Xor Queries

tachithanhdanh:


bqi343:
same as https://github.com/cpinitiative/usaco-guide/pull/577#issuecomment-753717581

PR #580, opened 2021-01-04
Fixed file name, added java sol for milk measurements

HarryWangATX:


HarryWangATX:
There's another file for bovine shuffle but with a different implementation so... 🤷‍♂️

HarryWangATX:
@bqi343 should I create a new file? or is it fine just doing what I did.

PR #586, opened 2021-01-05
Copy problem URL to clipboard 

MrinallU:
Done and should be ready to merge.


bqi343:
Is it possible to make "problem copied to clipboard" appear immediately when the URL is clicked (not when it's hovered over later?).

MrinallU:
> Is it possible to make "problem copied to clipboard" appear immediately when the URL is clicked (not when it's hovered over later?).

The tooltip automatically disappeared when clicked and there is no option to make it appear onClick, only on hover so unfortunately I couldn't make it like that.

MrinallU:
@bqi343 I also thought of adding toasts similar to how codeforces does it but I don't think thecodingwizard wants react toasts in the site.

thecodingwizard:
Just implement toasts manually (or use react toasts or equivalent library and style with tailwind ui?)

MrinallU:
@thecodingwizard https://www.creative-tim.com/learning-lab/tailwind-starter-kit/documentation/css/alerts found this

MrinallU:
@bqi343 What if I changed the icon of the url onClick to a clipboard and then after a few seconds it will switch back to the normal icon?

thecodingwizard:
I'm pretty sure there's a way to make the tooltip appear/disappear programmatically idk :/

bqi343:
> @bqi343 What if I changed the icon of the url when clicked to a clipboard icon and then (to indicate success) and then after a few seconds it will switch back to the normal icon?

Sure.

Also ig I don't completely understand how React works. Is there a separate instance of `currProblemCopied` for each instance of `ProblemComponent`?

bqi343:
Hm if I want to display "Problem already copied to clipboard!" only on the last copied problem, is this really the easiest way to do it?

https://dev.to/yezyilomo/global-state-management-in-react-with-global-variables-and-hooks-state-management-doesn-t-have-to-be-so-hard-2n2c

thecodingwizard:
Typically you can just use a timeout of like 5 seconds instead? Otherwise you'll have to use some sketchy context or something :c

MrinallU:
@thecodingwizard Got it.

MrinallU:
@thecodingwizard Sorry I haven't been able to work on this all month, as I have many competitions I have to get ready for.

bqi343:
reopen when ready ig

PR #589, opened 2021-01-05
Update Merging.mdx for Small-To-Large Merging section

wengh:


content/5_Plat/Merging.mdx
```diff
@@ -156,13 +156,7 @@ int main() {
...
 ## Generalizing
 
-A set doesn't have to be an `std::set`. Many data structures can be merged, such as `std::map` or `std:unordered_map`. However, `std::swap` doesn't necessarily work in $\mathcal{O}(1)$ time; for example, swapping two [arrays](http://www.cplusplus.com/reference/array/array/swap/) takes time linear in the sum of the sizes of the arrays, and the same goes for indexed sets. For two indexed sets `a` and `b` we can use `a.swap(b)` in place of `swap(a,b)`.
-
-<IncompleteSection>
-
-when exactly can we use `a.swap(b)` vs `swap(a,b)`?
-
-</IncompleteSection>
+A set doesn't have to be an `std::set`. Many data structures can be merged, such as `std::map` or `std:unordered_map`. However, `std::swap` work in $\mathcal{O}(1)$ time for STL data structures; for example, swapping two [arrays](http://www.cplusplus.com/reference/array/array/swap/) takes time linear in the sum of the sizes of the arrays, and the same goes for [GCC policy-based data structures](https://codeforces.com/blog/entry/11080). For two PBDS objects `a` and `b` we can use `a.swap(b)` in place of `swap(a,b)` to reach $\mathcal{O}(1)$ time.
```
frodakcin, line 159: However, `std::swap` work in $\mathcal{O}(1)$ time for STL data structures;

Typo?

frodakcin: Also, I'd add a warning saying that using `std::swap` instead of `a.swap(b)` is a very common mistake. In general, `a.swap(b)` is the safer choice. Check cpp documentation on [`std::swap` specializations](https://en.cppreference.com/w/cpp/algorithm/swap#Specializations) for when `std::swap` can be used instead of `a.swap(b)`.

frodakcin: Actually, this is beyond my knowledge. @bqi343 should probably be doing the review :P

wengh: > However, `std::swap` work in $\mathcal{O}(1)$ time for STL data structures;
> 
> Typo?

Yeah I meant that std::swap works in constant time _only_ for STL stuff.

bqi343: >  In general, a.swap(b) is the safer choice. 

Uhh why? `swap(a,b)` for arrays just calls `a.swap(b)` (https://en.cppreference.com/w/cpp/container/array/swap2).

bqi343: > > However, `std::swap` work in $\mathcal{O}(1)$ time for STL data structures;
> > Typo?
> 
> Yeah I meant that std::swap works in constant time _only_ for STL stuff.

`std::array` is STL though ...

wengh: Apparently data structures that use pointers internally support O(1) `swap`, while those that are accessed directly can only `swap` in linear time. It's interesting however that `string` also support O(1) `swap`.

wengh: And for some non-STL data structures O(1) swap is supported but they don't have a `std::swap` overload, so we need to use a method of the object instead.

bqi343: > Apparently data structures that use pointers internally support O(1) swap, while those that are accessed directly can only swap in linear time. It's interesting however that string also support O(1) swap.

Source?

> And for some non-STL data structures O(1) swap is supported but they don't have a std::swap overload, so we need to use a method of the object instead.

You can overload `std::swap`:

```cpp
#include <bits/stdc++.h>

struct test { int el; };

namespace std {
	void swap(test& x, test& y) {
		cout << "OOPS\n";
	}
}

int main() {
	test a,b;
	a.el = 5, b.el = 7;
	std::swap(a,b);
	std::cout << a.el;
}
```

wengh: > Source?
[`std::array` implementation](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/std/array#L50)
[`std::vector` implementation](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/bits/stl_vector.h#L93)

content/5_Plat/Merging.mdx
```diff
@@ -156,13 +156,14 @@ int main() {
 
 ## Generalizing
 
-A set doesn't have to be an `std::set`. Many data structures can be merged, such as `std::map` or `std:unordered_map`. However, `std::swap` doesn't necessarily work in $\mathcal{O}(1)$ time; for example, swapping two [arrays](http://www.cplusplus.com/reference/array/array/swap/) takes time linear in the sum of the sizes of the arrays, and the same goes for indexed sets. For two indexed sets `a` and `b` we can use `a.swap(b)` in place of `swap(a,b)`.
+A set doesn't have to be an `std::set`. Many data structures can be merged, such as `std::map` or `std:unordered_map`. However, $\mathcal{O}(1)$ time of `std::swap` may not apply for some low-level data structures: swapping two [arrays](http://www.cplusplus.com/reference/array/array/swap/) takes time linear in the sum of the sizes of the arrays, and the same goes for [GCC policy-based data structures](https://codeforces.com/blog/entry/11080). For two `__gnu_pbds::tree` or `__gnu_pbds::gp_hash_table` `a` and `b` we can use `a.swap(b)` in place of `swap(a,b)` to reach $\mathcal{O}(1)$ time.
```
bqi343, line 159: why "low-level"?

PR #591, opened 2021-01-05
Add warning

NewoIsTaken:
I added a warning to tell people to make sure that they have Java installed onto their computers before running the script.


dolphingarlic:
I don't think anyone will try to debug a Java program without Java installed though

NewoIsTaken:
@dolphingarlic Yeah, I guess, but I added it just in case.

thecodingwizard:
Oops, I really don't see how someone would run a Java testing script without having Java installed :P

Thanks for the PR though! :)

(If someone feels otherwise feel free to reopen ig)

PR #593, opened 2021-01-06
Update CSES Increasing Subsequence II.mdx

timothygao8710:
Added better explanation of solution, Java solution

### DIFF
--- a/solutions/CSES Increasing Subsequence II.mdx
+++ b/solutions/CSES Increasing Subsequence II.mdx
@@ -8,7 +8,7 @@ author: Benjamin Qi, Óscar Garries, Timothy Gao
 Let $dp(i)$ be the number of increasing subsequences ending on the right at $i$.
 Notice that $dp(i)$ can be calculated as the summation of all $dp(j)$ such that $j$ is in the range $[0,i]$ and nums[$j$] < nums[$i$].
 
-Naively, this gives an $O(N^2)$ solution. In order to speed this up, we can
+Naively, this gives an $\mathcal O(N^2)$ solution. In order to speed this up, we can
 coordinate compress, then range sum.
 
 Particularly, we sort all numbers in nums and map each distinct number to its index in the sorted array, then create a BIT based on these numbers. Now, when we query, we simply find the summation of all $j<k$ in the BIT, where $k$ is the index nums[$i$] maps to. 
--- a/solutions/CSES Increasing Subsequence II.mdx
+++ b/solutions/CSES Increasing Subsequence II.mdx
@@ -2,16 +2,24 @@
 id: cses-1748
 source: CSES
 title: Increasing Subsequence II
-author: Benjamin Qi, Óscar Garries, Timothy Gao
+author: Benjamin Qi, Óscar Garries and Timothy Gao
 ---
 
-Let $dp(i)$ be the number of increasing subsequences ending on the right at $i$.
-Notice that $dp(i)$ can be calculated as the summation of all $dp(j)$ such that $j$ is in the range $[0,i]$ and nums[$j$] < nums[$i$].
+Let $dp[i]$ be the number of increasing subsequences ending on the right at $i$.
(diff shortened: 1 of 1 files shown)


solutions/CSES Increasing Subsequence II.mdx
```diff
@@ -2,10 +2,16 @@
 id: cses-1748
 source: CSES
 title: Increasing Subsequence II
-author: Benjamin Qi, Óscar Garries
+author: Benjamin Qi, Óscar Garries, Timothy Gao
 ---
 
-Coordinate compress, then range sum.
+Let $dp(i)$ be the number of increasing subsequences ending on the right at $i$.
+Notice that $dp(i)$ can be calculated as the summation of all $dp(j)$ such that $j$ is in the range $[0,i]$ and nums[$j$] < nums[$i$].
```
dolphingarlic, line 9: Why is `nums[` not in latex? Also, the original array is called $a$ I think

solutions/CSES Increasing Subsequence II.mdx
```diff
@@ -2,10 +2,16 @@
...
+
+Naively, this gives an $O(N^2)$ solution. In order to speed this up, we can
```
dolphingarlic, line 11:
```suggestion
Naively, this gives an $\mathcal O(N^2)$ solution. In order to speed this up, we can
```

solutions/CSES Increasing Subsequence II.mdx
```diff
@@ -2,10 +2,16 @@
...
+coordinate compress, then range sum.
+
+Particularly, we sort all numbers in nums and map each distinct number to its index in the sorted array, then create a BIT based on these numbers. Now, when we query, we simply find the summation of all $j<k$ in the BIT, where $k$ is the index nums[$i$] maps to.
```
dolphingarlic, line 14: I don't think "particularly" is the right word to use here. Something like "More specifically" maybe?

Also, that's not exactly how the BIT works. I think you should mention how you update it as you sweep through the array. (Right now, it sounds like you initialize the BIT with all values already instead of all 0s initially)

solutions/CSES Increasing Subsequence II.mdx
```diff
@@ -58,6 +64,77 @@ int main () {
 	cout << sol << '\n';
 }
 ```
+Java implementation without template:
```
dolphingarlic, line 67: Use the `<LanguageSection>` and `<JavaSection>` tags (see [step 2](https://usaco.guide/general/contributing?lang=java#adding-a-solution))

dolphingarlic (changes requested):

solutions/CSES Increasing Subsequence II.mdx
timothygao8710: Right. "More specifically" sounds like a better word choice. I'll change that.

I wasn't being very clear with my explanation on the BIT, I'll update that. I kind of assumed that the reader already had knowledge of using a range query structure for the LIS problem which on hindsight I should not do.

solutions/CSES Increasing Subsequence II.mdx
timothygao8710: Ok. I'm pretty new the latex so I'm not sure what to put in latex. Thanks for pointing that out.

solutions/CSES Increasing Subsequence II.mdx
```diff
@@ -2,10 +2,16 @@
 id: cses-1748
 source: CSES
 title: Increasing Subsequence II
-author: Benjamin Qi, Óscar Garries
+author: Benjamin Qi, Óscar Garries, Timothy Gao
 ---
 
-Coordinate compress, then range sum.
+Let $dp(i)$ be the number of increasing subsequences ending on the right at $i$.
```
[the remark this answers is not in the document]
timothygao8710: Will update that.

solutions/CSES Increasing Subsequence II.mdx
timothygao8710: I think I will keep the array name as "nums" though because that makes it more clear what I'm referring to.

timothygao8710 (commented):
made requested changes

PR #595, opened 2021-01-06
Alternate solution to CF Orac-LCM

timothygao8710:
Java solution + explanation of alternate solution

### DIFF
--- a/solutions/CF Orac-LCM.mdx
+++ b/solutions/CF Orac-LCM.mdx
@@ -52,7 +52,7 @@ int main() {
 import java.util.*;
 import java.io.*;
 
-public class ocralcm {
+public class oraclcm {
 	public static void main(String[] args) throws IOException {
 		BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
 		int N = Integer.parseInt(br.readLine());
--- a/solutions/CF Orac-LCM.mdx
+++ b/solutions/CF Orac-LCM.mdx
@@ -2,12 +2,12 @@
 id: orac-lcm
 source: CF
 title: Orac & LCM
-author: Benjamin Qi
+author: Benjamin Qi and Timothy Gao
 ---
 
 [Official Editorial](https://codeforces.com/blog/entry/77284)
 
-## Solution
+## Solution: Prime Factorization
 
 For each prime, the second-to-lowest exponent of the prime that occurs in any of the numbers in the input is the exponent of this prime that will appear in the final answer. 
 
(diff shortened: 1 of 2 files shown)


solutions/CF Orac-LCM.mdx
```diff
@@ -25,22 +25,61 @@ ll a,b;
...
+    cout << b << endl;
 }
 ```
 
 If it's hard to understand what exactly this code is doing at first glance, a good first step is to simulating the code in the case where each $a_i$ is a power of the same prime (say, $2^k$). If the algorithm works for this case, then it will also work in the general case since the contributions of different primes are computed independently and multiplied together.
 
 <!-- the GCD function is [multiplicative](https://en.wikipedia.org/wiki/Multiplicative_function). -->
+
+## Solution: GCD/LCM
+Consider fixing $j$. How can the gcd of all lcms of nums[$j$] and all nums[$i$] (such that $i$ is in range $[0,j]$) be quickly computed? The result would be nums[$j$]$*C$, where $C$ is the gcd of all nums[$i$], and we can keep track of this gcd on the run.
```
frodakcin, line 49: I'd prefer if we used array `a` instead of `nums` as described in the problem statement.

Some wording changes might help too:

* "Consider fixing an index $j$" ...
* ... and all numbers $\text{a}[i]$ ($1 \leq i < j$) ...
* ... where $C$ is $\text{GCD}(\left{ a_1, a_2, \dots, a_{j-1} \right})$ ...

frodakcin (changes requested):
Optionally, you can call them Solution 1: Factorization and Solution 2: GCD/LCM. Up to you though.

bqi343:
> Optionally, you can call them Solution 1: Factorization and Solution 2: GCD/LCM. Up to you though.

Aren't the two solutions the same?

solutions/CF Orac-LCM.mdx
timothygao8710: Thank you. I will make those edits.

timothygao8710:
Unless I'm understanding it wrong, doesn't the first solution find the 2nd lowest power(0th pow included) for every prime divisor?

bqi343:
> Unless I'm understanding it wrong, doesn't the first solution find the 2nd lowest power(0th pow included) for every prime divisor?

It stores exactly the same information as your solution does.

timothygao8710:
True. I'm not sure if they are actually the same though(I can't code in C++). The explanations of the solutions are different regardless of the actual code.

frodakcin:
Alternatively, you can explain: "one way to implement this is to ...", or, "another explanation is ..." maybe?

solutions/CF Orac-LCM.mdx
```diff
@@ -2,17 +2,20 @@
...
 
-## Solution
+## Solution: Prime Factorization
 
 For each prime, the second-to-lowest exponent of the prime that occurs in any of the numbers in the input is the exponent of this prime that will appear in the final answer. 
 
 Here's a short solution that accomplishes this without explicitly computing any prime factorizations!
 
+<LanguageSection>
+ <CPPSection>
```
dolphingarlic, line 17:
```suggestion
<CPPSection>
```
This should fix the build error (I think?)

frodakcin:
Is someone able to fix deployment failed? Or, can someone teach me how to access this push request locally?

thecodingwizard:
I think I fixed it -- the newline after the CPPSection/JavaSection tag can't have any spaces or tabs.

timothygao8710:
Ohh I see. Thanks for the fix.

solutions/CF Orac-LCM.mdx
```diff
@@ -2,17 +2,20 @@
 id: orac-lcm
 source: CF
 title: Orac & LCM
-author: Benjamin Qi
+author: Benjamin Qi and Timothy Gao
```
bqi343, line 5: comma instead of and

solutions/CF Orac-LCM.mdx
```diff
@@ -25,22 +28,67 @@ ll a,b;
...
+
+</JavaSection>
+</LanguageSection>
+
 If it's hard to understand what exactly this code is doing at first glance, a good first step is to simulating the code in the case where each $a_i$ is a power of the same prime (say, $2^k$). If the algorithm works for this case, then it will also work in the general case since the contributions of different primes are computed independently and multiplied together.
 
 <!-- the GCD function is [multiplicative](https://en.wikipedia.org/wiki/Multiplicative_function). -->
+Here's another way of thinking about this:
+
+Consider fixing $j$. How can the gcd of all lcms of $a_j$, and all $a_i$ (such that $i$ is in range $[0,j]$) be quickly computed? The result would be $a_j*C$, where $C$ is the gcd of all $a_i$, and we can keep track of this gcd on the run.
```
bqi343, line 92: [0,j)?

bqi343: Wouldn't it be lcm(a_j,C)? Also, use `\cdot` instead of `*`

bqi343 (changes requested):

solutions/CF Orac-LCM.mdx
timothygao8710: Yes and yes. Not sure why I put a_j*C. Changes have been made.

PR #596, opened 2021-01-06
display modification times on syllabus for recently modified files

bqi343:
also Jan promotion counting, change default division value

### DIFF
--- a/content/1_General/Choosing_Lang.mdx
+++ b/content/1_General/Choosing_Lang.mdx
@@ -7,6 +7,8 @@ prerequisites:
  - intro
 ---
 
+test update
+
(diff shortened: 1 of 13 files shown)


bqi343:
jk modification times failed ...

thecodingwizard:
weird ://///// this feels like it should work? is the git command wrong or something? or is it Vercel's problem ://

thecodingwizard:
Okay I think (hope?) this works...

My best guess is that Vercel only fetches the very last commit so it doesn't have history info.

Unfortunately this hack only works if the repository is open source. If we ever close source the repository then the build will break...

Alternatives:

- Pre-commit hook to update a big JSON file (this won't work with Github web editor though)
- Github Action to commit once a day a big JSON file with the latest timestamps (but then we only have the history updated once a day)

PR #609, opened 2021-01-10
CSES Exponentation II Solution

tachithanhdanh:


dolphingarlic:
The unofficial editorial already has a C++ implementation, so maybe just link to that in the module instead of creating a new editorial?

tachithanhdanh:
Ok I will remove my implementation

PR #622, opened 2021-01-15
Faster, more space efficient and simpler solution to "Subsequences Summing to Sevens"

thetazero:
The solution passes all test cases and is more straight forward than the current solution. 
It uses simpler data structures and eliminates redundant writes to the prefix sum array.
Hope this helps!

### DIFF
--- a/solutions/USACO jan16-subseq-summing-to-sevens.mdx
+++ b/solutions/USACO jan16-subseq-summing-to-sevens.mdx
@@ -5,7 +5,7 @@ title: Subsequences Summing to Sevens
 author: Óscar Garries
 ---
 
-[Official Analysis](http://www.usaco.org/current/data/sol_div7_silver_jan16.html)
+[Official Analysis](http://www.usaco.org/prefrent/data/sol_div7_silver_jan16.html)
 
 <LanguageSection>
 <CPPSection>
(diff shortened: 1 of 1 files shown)


thecodingwizard:
I think this should be provided as an alternate solution (along with an explanation for how it works) rather than replacing the current solution? The problem is part of the prefix sums module so it's nice for at least one implementation to explicitly use prefix sums.

thetazero:
Yeah, I considered that however, the current solution seems not to make use of prefix sums, I believe it just generates the array. It exclusively uses pref[i + 1] which seems to be a roundabout equivalent to just having an int. However, I might be wrong. 

I can either do what you suggested or perhaps this problem should be moved from the prefix sums module. Regardless, I'll add an explanation.
Thoughts?

frodakcin:
Yea honestly the original solution isn't very good 😐

Farrius:
Yes, you are right pref[i + 1] is equivalent to have an int, and your solution is way better. However I would  replace  cur for a "prefix sum" array in order to use prefix sums in some way.

frodakcin:
Sorry, I sounded very rude. To elaborate, I thought using the pref array was fine. The only thing I dislike about the soln was the use of sets and maps, which could be so easily avoided.

thetazero:
Changed the solution to use a prefix array as suggested.

thetazero:
Sorry for a second commit, I realized I broke a link

solutions/USACO jan16-subseq-summing-to-sevens.mdx
```diff
@@ -5,7 +5,7 @@ title: Subsequences Summing to Sevens
 author: Óscar Garries
 ---
 
-[Official Analysis](http://www.usaco.org/current/data/sol_div7_silver_jan16.html)
+[Official Analysis](http://www.usaco.org/currrent/data/sol_div7_silver_jan16.html)
```
dongliu0426, line 8: typo: "currrent" should be "current"

dongliu0426 (changes requested):

solutions/USACO jan16-subseq-summing-to-sevens.mdx
thetazero: thanks! can't believe I missed that

bqi343: > thanks! can't believe I missed that

please fix this (and also update the author)

PR #623, opened 2021-01-16
Add Editorial for maxcross

MrinallU:


andrewwangva (commented):
Why is the original editorial not ok? The one USACO gives seems pretty straightforward to me.

MrinallU:
@andrewwangva They used binary search whereas the code I wrote uses prefix sums and two pointers. I'm not really writing them because the original editorials are bad, I'm explaining the solution I wrote which is quite different than the official editorial.

MrinallU:
@andrewwangva With their prefix sum approach they mainly left it as an afterthought so I went into more detail.

andrewwangva:
> @andrewwangva With their prefix sum approach they mainly left it as an afterthought so I went into more detail.

I still think their explanation of prefix sums is good enough

MrinallU:
@andrewwangva They never explicitly go out of their way to specify that you also need two pointers.

andrewwangva:
> @andrewwangva They never explicitly go out of their way to specify that you also need two pointers.

you don't really need two pointers? The code provided in your solution and the way they explained it is fine with just querying with a K-length range.

MrinallU:
@andrewwangva I used it though, and the editorial I wrote is to explain the thought process of my code.

andrewwangva:
> @andrewwangva I used it though, and the editorial I wrote is to explain the thought process of my code.

your code doesn't use two pointers? It loops through and queries K-length windows.

PR #624, opened 2021-01-16
Make editorial for lemonade line

MrinallU:


solutions/USACO feb17-lemonade.mdx
```diff
@@ -7,6 +7,8 @@ author: Mrinall Umasudhan
 
 [Official Analysis](http://www.usaco.org/current/data/sol_lemonade_silver_open18.html)
 
+Editorial: In this problem, we simply have to create a greedy algorithm to make the cows line up in decreasing order. By doing this we get to have the cows with the smaller line tolerance go later which in turn causes less cows to join the line which gives us the most optimal answer. Now after sorting the cow list then iterate over the new list and then proceed to check if the cow is able to tolerate the line length, if it is then we add the cow to the line by incrementing the line variable and then proceed to process the next cow. If the cow is not able to tolerate the line length then we stop sorting our cows and output our answer because we know that no other cows further down the list can tolerate the line because the rest of the cows' line tolerance will be less than or equal to the tolerance of our current cow so if we did process them the line would not increase anymore, again showing the advantage of sorting the list.
```
SansPapyrus683, line 10: Just my 2 cents: maybe break up this wall of text? Also, maybe break up the last sentence into 2 or 3 as well.

MrinallU:
@SansPapyrus683 Ok I got rid of the long sentence at the end.

andrewwangva (commented):
Again, I think the explanation USACO gives is good.

MrinallU:
@andrewwangva I wrote the code for the internal solution  so rather than add comments I put this explanation as I convey  information better in paragraph format, also my code is quite different than the c++ solution they give. Also the official editorial link is always there in the guide.

andrewwangva:
> @andrewwangva I wrote the code for the internal solution so rather than add comments I put this explanation as I convey information better in paragraph format, also my code is quite different than the c++ solution they give. Also the official editorial link is always there in the guide.

In your code, there's no need to continually sort, since it would already be sorted when removing the first element. Also, a cleaner implementation can be done using a [priority queue](https://usaco.guide/silver/intro-ordered?lang=java). Your logic is almost identical to the one USACO presented with the only difference being you remove elements while they simply loop through the array.

PR #625, opened 2021-01-16
Add editorial for hps

MrinallU:


solutions/USACO jan17-hoof-paper-scissors.mdx
```diff
@@ -2,11 +2,13 @@
...
...blem is narrowed down to two questions. How do we efficiently calculate the best place we need to change the gesture or if we need to change it at all? And how do we calculate the number of wins of a certain gesture given the range of turns the gesture is played? Well to calculate the number of wins a gesture makes given the range of turns it is used we have to maintain a prefix sum for each of the three gestures so any given range we can find the number of wins for a specific gesture. Now to find the area where the gesture should be changed to maximize wins we then go through each index of the array and use the given ranges of $\texttt{moves[1]}$  to $\texttt{moves[i]}$, apply it to the prefix sums and add the number of wins to the prefix sum that outputs the greatest amount of wins; Meaning that gesture gives the largest amount of wins in that range. We then repeat the same process for the ranges of $\texttt{moves[i]}$ to $\texttt{moves[N]}$ and add the value to the greatest amount of wins in the range $\texttt{moves[1]}$ to $\texttt{moves[i]}$ as specified in a previous step. Finally, we store the greatest sum of wins in the simulation of changing gestures and output the answer.
```
andrewwangva, line 10: > Editorial: As opposed to the official editorial there also exists a simpler solution in which one does not need to get involved with multiple 2D arrays, instead there is also a fairly straightforward solution using prefix sums with 1D arrays.
unnecessary

andrewwangva, line 10: > Now the problem is narrowed down to two questions. How do we efficiently calculate the best place we need to change the gesture or if we need to change it at all? And how do we calculate the number of wins of a certain gesture given the range of turns the gesture is played? Well 

Also unnecessary

andrewwangva (commented):
Clean up the explanation a bit. Try to shorten it and make it more straightforward.

MrinallU:
@andrewwangva Sorry, I went on a bit of a ramble when explaining the problem. It should be more straightforward now.

HarryWangATX:
I think someone replaced your code with theirs some time ago

MrinallU:
Both of our code is the same as he just changed the variable names, we
follow the exact same logic.

frodakcin (changes requested):
Still quite unclear. Maybe try paragraph breaks?

The reader shouldn't need to think very hard to understand the solution.

PR #626, opened 2021-01-16
Add python code for Greedy.mdx, Prefix_Sums.mdx, and solution to Subordinates

topazand:
I wonder if I need to write python solution according to `sorting-custom.mdx` using classes instead of lists.


topazand:
open pull request from a new branch

PR #630, opened 2021-01-17
add_cpp_solution

Farrius:
C++ solution for CEOI 2010 - A Huge Tower

### DIFF
--- a/solutions/CEOI 10-tower.mdx
+++ b/solutions/CEOI 10-tower.mdx
@@ -1,0 +1,41 @@
+---
+id: ceoi-10-tower
+source: CEOI 2010
+title: A Huge Tower
+author: Óscar Garries
+---
+
+[Official Analysis](http://ceoi2010.ics.upjs.sk/pub/documents/ceoi_summary.pdf#page=14)
+
+<LanguageSection>
+<CPPSection>
+
+## C++ Implementation
+
+```cpp
+#include <bits/stdc++.h>
+
+using namespace std;
+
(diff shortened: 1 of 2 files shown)


solutions/CEOI 10-tower.mdx
```diff
@@ -0,0 +1,41 @@
+---
+id: ceoi-10-tower
+source: CEOI 2010
+title: A Huge Tower
+author: Óscar Garries
+---
+
+[Official Analysis](http://ceoi2010.ics.upjs.sk/pub/documents/ceoi_summary.pdf#page=14)
```
andrewwangva, line 8: Looks good but is there any chance you can add an explanation? The one they give seems kinda long and wordy so it would be nice.

dolphingarlic: I don't think the official one is that bad. The full-points solution is only 4 paragraphs (the rest explains sub-optimal solutions)

andrewwangva: 4 paragraphs is long for me 😆

Farrius: I think the explanation for 100 points it's fine but the solution with all the subtask is a little messy. Maybe add something like just look the final paragraph for the 100 points explanation ?

andrewwangva: you know what, it doesn't matter that much. It's fine without an explanation.

andrewwangva (approved):

PR #637, opened 2021-01-19
fix prefix sum module errors

frodakcin:


content/3_Silver/Prefix_Sums.mdx
```diff
@@ -700,13 +700,13 @@ A common application of prefix sums is to determine, given a set of intervals, t
...
 
 This results in a $\mathcal{O}(nV)$ time complexity (where $0 \leq a \leq b \leq V$), which easily times out (think what happens when the interval $[0, V]$ is queried $n$ times).
 
 We can do better. It's easy to see that an increment of $x$ (before computation) in $\texttt{arr}[i]$ causes all subsequent $\texttt{prefix}[i...V]$ (after computation) to increase by $x$. We can also "undo" this operation by adding $-x$ to $\texttt{arr}[i]$. This concept can be conceptualized through increment and decrement points. An increment point increases (and decrement decreases) all subsequent cells. Note that our decrement point is located at $B+1$ because the interval is inclusive - decrementing at point $B$ turns the interval to $[A, B)$.
 
-Example 1: Add two to each point in the interval $[2, 4]$
+**Example 1**: Add two to each point in the interval $[2, 4]$
 
 Our matrix after adding 2 at our increment (start) point (before computation)
```
bqi343, line 711: huh

bqi343 (changes requested):
also, restaurant customers is used in sorting w/ custom comparators. Should prefix sums be a prerequisite for that module?

content/3_Silver/Prefix_Sums.mdx
frodakcin: What's wrong here?

bqi343: why is it matrix

frodakcin:
> also, restaurant customers is used in sorting w/ custom comparators. Should prefix sums be a prerequisite for that module?

It would be a bit weird to make that module have a prerequisite that appears after it sequentially. Maybe we can rearrange the modules or just remove restaurant customers from the custom comparator problem list. What do you think?

content/3_Silver/Prefix_Sums.mdx
frodakcin: Oh, I see. For some reason i thought the comment was referring to line 709

bqi343:
Another way to split the module is problems which involve sorting or sets or maps vs those that don't. :o

bqi343:
Why is "2-Dimensional" under "Creation from Endpoints"?

frodakcin:
> Why is "2-Dimensional" under "Creation from Endpoints"?

It's the same idea, but for rectangles instead of segments.

> Maybe. I'm fine with just listing prefix sums as a prerequisite for sorting custom? and maybe adding an additional note at the start of that module.

Sure. We can add a note saying that since sorting is a very general topic, it may involve ideas from other silver modules.

frodakcin:
> Another way to split the module is problems that involve sorting or sets or maps vs those that don't.

How would this be done?

bqi343:
> > Another way to split the module is problems that involve sorting or sets or maps vs those that don't.
> 
> How would this be done?

Nvm, I don't think any of the problems involve sorting or sorted sets aside from Restaurant Customers ...

bqi343 (approved):

frodakcin (commented):
Prefix sum module looks fine as is, but maybe we can put 2D before "Additional Applications" section?

I think Restaurant Customers' explanation needs to more clearly address the problem statement, maybe even adding an extra sentence to the beginning so it doesn't start so abruptly with "A common application of prefix sums is to determine..."

content/3_Silver/Sorting_Custom.mdx
```diff
@@ -591,4 +592,10 @@ public class Sol {
 
 ## Problems
 
+<Info>
+
+Sorting is a very general idea, so many of the problems below may use other Silver concepts, such as [prefix sums](/silver/prefix-sums), or [binary search](/silver/binary-search-ans), etc.
```
bqi343, line 597: binsearch-ans isn't a prerequisite though :o

bqi343 (changes requested):

content/3_Silver/Sorting_Custom.mdx
frodakcin: I don't think it really makes sense to have a prereq that isn't very involved with sorting. Wouldn't it be sufficient just to say that the problems below may use other silver concepts?

Alternative, we can move this module to the end so it's more in sequential order.

bqi343: > I don't think it really makes sense to have a prereq that isn't very involved with sorting. Wouldn't it be sufficient just to say that the problems below may use other silver concepts?

Sure (and just include prefix sums)?

bqi343:
looks alright

PR #648, opened 2021-01-23
Modifying solution for Ferris Wheel

Nekotizimo:
Added alternate C++ implementation, wrote brief explanation for both solutions, added problem tags

### DIFF
--- a/solutions/CSES Ferris Wheel.mdx
+++ b/solutions/CSES Ferris Wheel.mdx
@@ -9,16 +9,19 @@ author: Danh Ta Chi Thanh, Kenny Cho
 
 [Unofficial Editorial](https://codeforces.com/blog/entry/83295)
 
-Since each gondola can contain either 1 or 2 children, starting with the heaviest child, we can either
+Since each gondola can contain either 1 or 2 children, for each gondola, we can either
+1. pair the lightest child with the heaviest child possible without exceeding the weight limit; if pairing isn't possible, we
+2. only include the lightest child.
+
+Those left unpaired each get their own gondola. The implementation below uses this method.
+
+Alternatively, for each gondola, we can either
 1. pair the heaviest child with the lightest child if possible; or
 2. only include the heaviest child.
 
-Alternatively, we can start with the lightest child. For each gondola, we pair the lightest child with
-the heaviest child possible without exceeding the weight limit. Those that are left each get their own gondola. 
+Both pairing methods can be achieved in $\mathcal{O}(n)$ time using two pointers, though sorting is more expensive.
 
-**Time Complexity:** $\mathcal{O}(n\log n)$ (from sorting)
-
-This solution starts with the lightest child:
+**Overall Time Complexity:** $\mathcal{O}(n\log n)$ (from sorting)
 
 <LanguageSection>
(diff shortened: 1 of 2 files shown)


frodakcin:
The two solutions are pretty much identical?

Nekotizimo:
Right, I guess it's more of a slightly simpler implementation rather than a different solution. Do you think there's value added?

tachithanhdanh:
> The two solutions are pretty much identical?

I think the original two pointers solution is simple enough to implement and as you said, the two new solutions are quite similar to each other

dongliu0426:
it's basically identical to the solution in the unofficial editorial.

frodakcin:
Is it ok to close this pr?

Nekotizimo:
Sure :P

tachithanhdanh:
We can merge the commits on the module .mdx file because the tags which @Nekotizimo added are very informative

Nekotizimo:
Perhaps I could remove the duplicate solution but leave the explanations and the tags I added. I think those will be helpful to some.

frodakcin:
Yea sure, that sounds good

solutions/CSES Ferris Wheel.mdx
```diff
@@ -2,14 +2,27 @@
...
-author: Danh Ta Chi Thanh
+author: Danh Ta Chi Thanh, Kenny Cho
 ---
 
-
 ## Solution
 
 [Unofficial Editorial](https://codeforces.com/blog/entry/83295)
 
+Since each gondola can contain either 1 or 2 children, starting with the heaviest child, we can either
```
frodakcin, line 12:
```suggestion
Since each gondola can contain either 1 or 2 children, we can either

```

solutions/CSES Ferris Wheel.mdx
```diff
@@ -2,14 +2,27 @@
...
+1. pair the heaviest child with the lightest child if possible; or
+2. only include the heaviest child.
+
+Alternatively, we can start with the lightest child. For each gondola, we pair the lightest child with
+the heaviest child possible without exceeding the weight limit. Those that are left each get their own gondola. 
+
+**Time Complexity:** $\mathcal{O}(n\log n)$ (from sorting)
+
+This solution starts with the lightest child:
```
frodakcin, line 21: Depending on how you look at this, it *is* starting with the heaviest child (focus on variable $j$).

frodakcin (changes requested):

solutions/CSES Ferris Wheel.mdx
tachithanhdanh: Yeah, that's how the two pointers technique works

solutions/CSES Ferris Wheel.mdx
Nekotizimo: Removed this phrase in new commit

solutions/CSES Ferris Wheel.mdx
Nekotizimo: True. I made edits to try and clarify the explanation.

frodakcin (approved):
looks good.

PR #651, opened 2021-01-23
Added another way of File I/O in Python

rickastley1:
I have added another way, dare I say better, of using file input-output in Python

### DIFF
--- a/content/1_General/IO.mdx
+++ b/content/1_General/IO.mdx
@@ -555,6 +555,18 @@ sys.stdin = open("problemname.in", "r")
 sys.stdout = open("problemname.out", "w")
 ```
 
+Another way to do File I/O is as follows:
+
+```py
+with open("problemname.in", "r") as fin:
+    lines = fin.readlines()
+
+# Your Code
+
+with open("problemname.out", "w") as fout:
+    fout.write(result)
+```
+
 <Warning title="USACO Note">
 
 Since USACO currently uses python 3.4.0, [f-strings](https://realpython.com/python-f-strings/) are not supported (they will result in a missing output file).


bqi343:
Why is this better?

rickastley1:
I just thought that instead of changing sys.stdout/stdin this would be better as it sort of cleans up after you. For the existing example, using a print statement anywhere would output onto the file which is not as optimal for debugging, for example. My method only uses file I/O where it is necessary.

rickastley1:
> Why is reading in all lines at the start better?

It takes in each line and inserts it into a list

dongliu0426:
i think it's easier to do `fin.readline()` every single time a line needs to be read.

bqi343:
> i think it's easier to do `fin.readline()` every single time a line needs to be read.

`input()` is even easier :P

> For the existing example, using a print statement anywhere would output onto the file which is not as optimal for debugging, for example. 

I assume that when you're running the program locally, you'd comment out the lines with `sys.stdin` and `sys.stdout`, so you wouldn't be printing to a file. Also, do you not view your program output + debug output together?

rickastley1:
Hmm... true, I guess at this point, it's just an alternative rather than a strictly speaking "better" option
> I assume that when you're running the program locally, you'd comment out the lines with `sys.stdin` and `sys.stdout`, so you wouldn't be printing to a file. Also, do you not view your program output + debug output together?
Too much effort XD.

thecodingwizard:
I think it's a good alternative but I think it's already somewhat provided here? https://usaco.guide/general/io?lang=py#method-1-1

I'll close this PR for now; feel free to keep discussing though...

rickastley1:
Ok thank you

PR #653, opened 2021-01-25
content: add binary search impl for cellular network

thecodingwizard:
I'm thinking of making Cellular Network the challenge problem for Binary Search on Sorted Array for CPI clubs.

### DIFF
--- a/solutions/CF Cellular Network.mdx
+++ b/solutions/CF Cellular Network.mdx
@@ -9,14 +9,14 @@ author: Benjamin Qi
 
 [Official Editorial](https://codeforces.com/blog/entry/46324?locale=en)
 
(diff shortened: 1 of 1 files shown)


solutions/CF Cellular Network.mdx
```diff
@@ -9,6 +9,58 @@ author: Benjamin Qi
 
 [Official Editorial](https://codeforces.com/blog/entry/46324?locale=en)
 
+This solution can also be implemented with Binary Search, but it's better to implement it with a set. The code below is provided only if you want additonal practice for Binary Searching a Sorted Array.
```
frodakcin, line 12: Technically, I think "binary search" and "sorted array" should be lowercase. It's definitely fine as is though.

frodakcin (approved):
Code looks very clean to me.

If you'd like, you can write `const vector<int> &arr` instead.

PR #657, opened 2021-01-26
Added #2 solution for CSES - Coin Combinations II

tachithanhdanh:
I added the Codeforces Unofficial Editorial for this problem as the editorial uses another approach which could be useful. I also add some explanation for Michael Cao's solution.


caoash:
I like the unofficial editorial, but the comment seems to just be restating what I had.

caoash (changes requested):
See comment.

PR #659, opened 2021-01-26
Binary lifting

nwatx:
Solution + explanation for company queries I


content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,100 @@ export const problems = {
...
+</Info>
+
 ### Solution
 
-<IncompleteSection />
+To solve this problem, we can use a standard binary lifting implementation where ```query(int v, int j)``` corresponds to the $j$th ancestor of v.
+
+In our ```query(int v, int j)``` if we ever encounter a value of $-1$ or $0$, then such a node does not exist and we can simply return its value. This is because the lowest number a node can be is $1$ (the root of the tree).
+
+To calculate the columns required for the ```up``` array, use the formula $\textrm{ceil}(\log_2{N})+1$ which in our case simplifies to $\textrm{ceil}(\log_2{(2\cdot 10^5)}) + 1 = 19$.
```
bqi343, line 229: you only need 18 columns

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,100 @@ export const problems = {
...
+<Info>
+
+Binary jumping is more commonly referred to as "binary lifting."
+
+</Info>
+
 ### Solution
 
-<IncompleteSection />
+To solve this problem, we can use a standard binary lifting implementation where ```query(int v, int j)``` corresponds to the $j$th ancestor of v.
```
bqi343, line 225: v should be texified

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,100 @@ export const problems = {
...
+
+    FOR(i, 1, 20) {
+        if(up[v][i-1] != -1) up[v][i] = up[up[v][i-1]][i-1]; // construction of our ancestor tree
+        else up[v][i] = -1;
+    }
+
+    for(int e : adj[v]) if(e != p) dfs(e, v); // standard tree dfs
+}
+
+int query(int v, int j) { // a is our vertex, and j is the jump we make
```
bqi343, line 270: this is log^2N :(

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,100 @@ export const problems = {
...
+const int MX = 2e5+5;
+
+int up[MX][20]; // 19 is calculated with our expression but we can add padding to be safe.
+
+vector<int> adj[MX]; // adjacency list
+
+void dfs(int v, int p) { // we keep track of the current node v and its parent p
+    up[v][0] = p;
+
+    FOR(i, 1, 20) {
```
bqi343, line 262: probably more convenient to let the root to be 0 and let it point to itself (`up[0][0]=0`) like here:

https://github.com/bqi343/USACO/blob/master/Implementations/content/graphs%20(12)/Trees%20(10)/LCAjump.h

No need to introduce `-1`.

bqi343 (changes requested):

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,100 @@ export const problems = {
...
+int main() {
+    ios_base::sync_with_stdio(0);
+    cin.tie(0);
+
+    cin >> N >> Q;
+    FOR(i, 2, N+1) { // one based indexing
+        int a; cin >> a;
+        adj[a].push_back(i);
+        adj[i].push_back(a);
+    }
```
dongliu0426, line 287: why not just set `up[i][0]` to `a` and evaluate the rest of the `up[i]` here?

dongliu0426 (changes requested):

content/5_Plat/Bin_Jump.mdx
nwatx: I just changed the implementation to be more consistent with sols but I suppose that would work too.

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+    FOR(i, 1, MS) up[i][x] = up[i-1][up[i-1][x]];
+    trav(y,adj[x]) if (y != up[0][x]) {
+        up[0][y]=x;
+        dfs(y);
+    }
+}
+
+int jmp(int x, int d) {
+    F0R(i, MS) if((d >> i) & 1) x = up[i][x];
+    return x ? x : -1; // modfication to return -1 if not found
```
bqi343, line 279: you can do `x ?: -1`

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+<Info>
+
+Binary jumping is more commonly referred to as "binary lifting."
+
+</Info>
+
 ### Solution
 
-<IncompleteSection />
+To solve this problem, we can use a standard binary lifting implementation where ```jmp(int x, int d)``` corresponds to the $d$th ancestor of $x$.
```
bqi343, line 225: $d$-th

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+
+In our ```jmp(int x, int d)``` if our final value of $x$ is $0$, then such a node does not exist and we can simply return $-1$. This is because the lowest number a node can be is $1$ (the root of the tree). 
+
+In our ```jmp(int x, int d)``` implementation, we can test if we jump in powers of two by using the $\&$ operator. If the $i$th bit on the right is toggled, then we jump. For example, a jump of $13$ would correspond to the binary number $1101$. We would jump 3 times on bits $$i$$.
```
bqi343, line 229: list the bits on which you jump

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+
+<Spoiler title="Illustration of the jump method">
+
+![Binary Lifting Diagram](./assets/binaryliftingjump.png)
+
+</Spoiler>
+
+To calculate the columns required for the ```up``` array, use the formula $\textrm{ceil}(\log_2{N})$ which in our case simplifies to $\textrm{ceil}(\log_2{(2\cdot 10^5)}) = 18$.
```
bqi343, line 237: use $\lceil \rceil$

bqi343: It's not clear why it's "columns" instead of "rows" here. Personally, I would say that `int up[MS][MX]; ` is an array of `MS` rows and `MX` columns.

bqi343 (changes requested):

content/5_Plat/Bin_Jump.mdx
nwatx: woah that's nice

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+int main() {
+    ios_base::sync_with_stdio(0);
+    cin.tie(0);
+
+    cin >> N >> Q;
+
+    FOR(i, 2, N+1) {
+        int a; cin >> a;
+        adj[a].push_back(i);
+        adj[i].push_back(a);
```
dongliu0426, line 290: it's sufficient to just do `adj[a].push_back(i);`.

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,105 @@ export const problems = {
...
+int up[MS][MX]; 
+
+vector<int> adj[MX]; // adjacency list
+
+void dfs(int x) { // build jump table
+    FOR(i, 1, MS) up[i][x] = up[i-1][up[i-1][x]];
+    trav(y,adj[x]) if (y != up[0][x]) {
+        up[0][y]=x;
+        dfs(y);
+    }
```
dongliu0426, line 273: remove the if statement (if you followed my other suggestion)

dongliu0426 (changes requested):

dongliu0426 (approved):

content/5_Plat/Bin_Jump.mdx
bqi343: well, then you might as well not DFS at all

bqi343 (changes requested):
see https://github.com/cpinitiative/usaco-guide/pull/659#discussion_r565810976

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,95 @@ export const problems = {
...
+#define trav(a, x) for (auto& a : x)
+
+int N, Q;
+
+const int MX = 2e5+5; // maximum N
+const int MS = 18; // 18 is calculated by $\log_2 2e5$
+
+int up[MS][MX]; 
+
+vector<int> adj[MX]; // adjacency list
```
dongliu0426, line 266:
```suggestion
```

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,95 @@ export const problems = {
...
+        int a; cin >> a;
+        up[0][i] = a;
+    }
+
+    FOR(i, 1, 18) FOR(j, 1, N+1) up[i][j] = up[i-1][up[i-1][j]]; // construct our parent array
+
+    // uncomment below if you want to see the parent array
+
+    // F0R(i, MS) {
+    //     FOR(j, 1, N+1) { // change this to the depth log_2(depth)
```
dongliu0426, line 289: >  // change this to the depth log_2(depth)

?

dongliu0426 (changes requested):

dongliu0426 (approved):

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,93 @@ export const problems = {
...
+    cin.tie(0);
+
+    cin >> N >> Q;
+
+    FOR(i, 2, N+1) {
+        int a; cin >> a;
+        up[0][i] = a;
+    }
+
+    FOR(i, 1, 18) FOR(j, 1, N+1) up[i][j] = up[i-1][up[i-1][j]]; // construct our parent array
```
bqi343, line 282: use MS instead of 18?

content/5_Plat/Bin_Jump.mdx
```diff
@@ -215,9 +214,93 @@ export const problems = {
...
+
+In our implementation, we test if we jump in powers of two by using the $\&$ operator. If the $i$th bit on the right is toggled, then we jump. For example, a jump of $13$ would correspond to the binary number $1101$. We would jump 3 times on bits $0, 2, 3$ (in that order) counting from the right.
+
+<Spoiler title="Illustration of the jump method">
+
+![Binary Lifting Diagram](./assets/binaryliftingjump.png)
+
+</Spoiler>
+
+To calculate the rows required for the ```int up[MS][MX]``` array, use the formula $\lceil (\log_2{N}) \rceil$ which in our case simplifies to $\lceil (\log_2{(2\cdot 10^5)})\rceil = 18$.
```
bqi343, line 237: \lceil (\log_2{(2\cdot 10^5)})\rceil

don't need the outer parentheses

bqi343 (changes requested):

PR #660, opened 2021-01-27
Fixes and Updates for C++ Command Line Module

mizlan:
More aligned with the spec and some pointers and tips. Also removed some potentially dangerous advice and other unorthodox tips in favor of best practices.

### DIFF
--- a/content/1_General/Cpp_Command.mdx
+++ b/content/1_General/Cpp_Command.mdx
@@ -159,23 +159,32 @@ If you want to code in (neo)vim, you can install WSL and code through WSL bash.
 
     probably suffices.
 
-4.  You should be able to compile with `g++-#`, where # is the version number (ex. 9). Running the following command
+4.  You should be able to compile with `g++-#`, where # is the version number (e.g., 10). Running the following command
 
     ```
-    g++-9 --version
+    g++-10 --version
     ```
 
     should display something like this:
 
     ```
-    g++-9 (Homebrew GCC 9.2.0_2) 9.2.0
-    Copyright (C) 2019 Free Software Foundation, Inc.
+    g++-10 (Homebrew GCC 10.2.0_2) 10.2.0
+    Copyright (C) 2020 Free Software Foundation, Inc.
     This is free software; see the source for copying conditions.  There is NO
     warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
     ```
 
-5.  If you want to be able to compile with just `g++`, write a shell alias! Aliases are discussed [here](TODO: link to alias section). Avoid overriding the system `g++` as that will almost surely cause problems.
+5.  If you want to be able to compile with just `g++`, write a shell alias! Put the following lines into your shell's rc file (`~/.bashrc` if you use `bash`, and `~/.zshrc` if you use `zsh`).
+
+    ```bash
+    alias g++=g++-10
+    ```
+
     Once you do so, `g++ --version` should now output the same thing as `g++-9 --version`.
+
+    Note: avoid overriding the system `g++` with symlinking or hard-linking as
+    that will almost surely cause problems. Don't worry if you don't know what
+    those terms mean.
 
 <!-- <Resources>
 	<Resource
--- a/content/1_General/Cpp_Command.mdx
+++ b/content/1_General/Cpp_Command.mdx
@@ -326,7 +326,7 @@ co() { g++ -std=c++17 -O2 -o "${1%.*}" $1 -Wall; }
 run() { co $1 && ./${1%.*} & fg; }
 ```
 
-Now you can easily compile and run `name.cpp` from the command line with `co name.cpp && ./name` or `run name.cpp`. Note that all occurrences of `$1` in the function are replaced with `name.cpp`, while `${1%.*}` removes the file extension from `$1`.
+Now you can easily compile and run `name.cpp` from the command line with `co name.cpp && ./name` or `run name.cpp`. Note that all occurrences of `$1` in the function are replaced with `name.cpp`, while `${1%.*}` removes the file extension from `$1` to produce `name`.
 
 ### What is `& fg` for?
 
@@ -353,7 +353,7 @@ int main() {
 }
 ```
 
-According to the resource above, the `& fg` is necessary for getting `zsh` on Mac to display crash messages (such as segmentation fault). For example, consider the running the first `prog.cpp` above with `run prog`.
+According to the resource above, the `& fg` is necessary for getting `zsh` on Mac to display crash messages (such as segmentation fault). For example, consider the running the first `prog.cpp` above with `run prog.cpp`.
 
 If `& fg` is removed from the run command above then the terminal displays no message at all. Leaving it in produces the following (ignore the first two lines):
 
--- a/content/1_General/Cpp_Command.mdx
+++ b/content/1_General/Cpp_Command.mdx
@@ -136,7 +136,7 @@ If you want to code in (neo)vim, you can install WSL and code through WSL bash.
 
     If you previously installed these you may need to update them:
 
-    ```
+    ```bash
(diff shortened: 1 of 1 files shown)


content/1_General/Cpp_Command.mdx
```diff
@@ -159,29 +159,32 @@ If you want to code in (neo)vim, you can install WSL and code through WSL bash.
...
+    ```bash
+    alias g++=g++-10
     ```
-    cd /usr/local/bin
-    ln -fs g++-9 g++
-    ```
 
-    `g++ --version` should now output the same thing as `g++-9 --version`.
+    Once you do so, `g++ --version` should now output the same thing as `g++-9 --version`.
+
```
thecodingwizard, line 184: should this be `g++-10 --version`?

content/1_General/Cpp_Command.mdx
```diff
@@ -286,32 +302,32 @@ Of course, retyping the flags above can get tedious. You should define shortcuts
...
+subl ~/.zshrc
 ```
 
 You can add **aliases** and **functions** here, such as the following to compile and run C++ on Mac.
 
 ```
-co() { g++ -std=c++17 -O2 -o $1 $1.cpp -Wall; }
-run() { co $1 && ./$1 & fg; }
+co() { g++ -std=c++17 -O2 -o "${1%.*}" $1 -Wall; }
+run() { co $1 && ./${1%.*} & fg; }
```
thecodingwizard, line 327: Isn't the original shell script preferable since it safes you a few keystrokes (you don't have to type .cpp)?

Or is there something I'm missing? (oops I don't use shell scripts)

thecodingwizard:
Thanks for the PR! 😃

content/1_General/Cpp_Command.mdx
```diff
@@ -136,7 +136,7 @@ If you want to code in (neo)vim, you can install WSL and code through WSL bash.
 
     If you previously installed these you may need to update them:
 
-    ```
+    ```bash
```
thecodingwizard, line 139: I don't believe we have syntax highlighting for bash (I think). (Keep the change though; we might add it in the future.)

If there is sufficient need for bash syntax highlighting we can add it, though this will increase the bundle size.

content/1_General/Cpp_Command.mdx
mizlan: As hinted at in the commit message, this makes completion easier (tab completion normally works by files). The original method would imply completing and then deleting the `.cpp` from the end if that makes sense.

content/1_General/Cpp_Command.mdx
mizlan: Makes sense; I would leave it out. The benefit is very minimal.

content/1_General/Cpp_Command.mdx
mizlan: Yup, thanks for catching that! Will run through another time to make sure everything got updated.

mizlan:
This can be merged at your discretion; I don't have any other changes planned.

content/1_General/Cpp_Command.mdx
```diff
@@ -159,29 +159,32 @@ If you want to code in (neo)vim, you can install WSL and code through WSL bash.
...
-    cd /usr/local/bin
-    ln -fs g++-9 g++
+    ```bash
+    alias g++=g++-10
     ```
 
-    `g++ --version` should now output the same thing as `g++-9 --version`.
+    Once you do so, `g++ --version` should now output the same thing as `g++-10 --version`.
+
+    Note: avoid overriding the system `g++` with symlinking or hard-linking as
```
bqi343, line 185: >  Note: avoid overriding the system `g++` with symlinking or hard-linking as
    that will almost surely cause problems. Don't worry if you don't know what
    those terms mean.

what problems?

content/1_General/Cpp_Command.mdx
```diff
@@ -234,9 +237,23 @@ If you type some integer and then press enter, then the program should produce o
 g++ name.cpp -o name && ./name
 ```
 
+Don't be afraid, the `./name` will only run if `g++ name.cpp -o name` finished successfully. Such is the magic of `&&`!
```
bqi343, line 240: > Don't be afraid,

??

content/1_General/Cpp_Command.mdx
mizlan: System build processes which expect clang.

content/1_General/Cpp_Command.mdx
mizlan: That is to say that you need not worry about `./name` running if the compilation wasn't successful.

content/1_General/Cpp_Command.mdx
bqi343: Could you give an example?

(oops why don't they just use `clang++` then)

bqi343:
looks like the prerequisite module should be updated to `running-code-locally` instead of `running-code`

content/1_General/Cpp_Command.mdx
mizlan: It's a best practice to not override system packages in general, not limited to just `g++`.

bqi343: still waiting on a concrete example

why not override `g++` with the *real* `g++`? :/

content/1_General/Cpp_Command.mdx
thecodingwizard: Maybe rephrase it into something slightly more formal to flow better with the rest of the module? (Maybe just say `&&` only executes the second command if the first command is successful, so `./name` will only run if the compilation finished successfully. or something?)

content/1_General/Cpp_Command.mdx
mizlan: For the same reason Apple chose to link `g++` to `clang++`—probably not a very good reason, but we will never know. :/ Also I just realized it was `/usr/local/bin` versus `/usr/bin`, but the issue of `PATH` resolution still exists. As such, it will always be safer to only activate this in _interactive_ shells, which is what aliasing in the rc file does. In addition, the creation of such a link in `/usr/local/bin` is now not managed by Homebrew, which may lead to a dangling symlink if the target executable is (re)moved.

As for a concrete example, `g++` is not really seen as much in the wild anymore. So I guess you could say this is overblown, but the fact is it's not best practice for multiple reasons.

bqi343: tbh I don't completely understand all of this but seems alright :)

content/1_General/Cpp_Command.mdx
```diff
@@ -254,11 +271,11 @@ g++ -std=c++17 -O2 name.cpp -o name -Wall
 
 Explanation:
 
-- `-O2` tells `g++` to compile your code to run more quickly (see [here](https://www.rapidtables.com/code/linux/gcc/gcc-o.html))
+- `-O2` tells `g++` to compile your code to run more quickly (see [here](https://www.rapidtables.com/code/linux/gcc/gcc-o.html)). Note that `-O2` will slow down compilation time. Think of it as a trade-off between runtime and compile time. 
 - `-std=c++17` allows you to use features that were added to C++ in 2017. USACO currently uses `-std=c++11`.
 - `-Wall` checks your program for common errors. See [Debugging](/general/debugging-lang/#warning-options) for more information.
 
-You should always compile with these flags.
+You should always compile with these flags (with the possible exception of `-O2`, as described above).
```
bqi343, line 278: Is the increase in compile time so significant?

content/1_General/Cpp_Command.mdx
mizlan: Absolutely, will change it.

content/1_General/Cpp_Command.mdx
mizlan: Probably not horribly significant, revert that if you want.

PR #664, opened 2021-01-28
New Problems

HarryWangATX:


bqi343:
uhh let's not put silver problems in bronze ad hoc

bqi343:
It seems odd to keep placing cycle problems in func graphs but idk where else

HarryWangATX:
> uhh let's not put silver problems in bronze ad hoc (also, marking it as easier than the other problems is misleading)

What about intro to greedy?

bqi343:
sure ig we can just dump all of silver greedy there :eyes:

HarryWangATX:
> sure ig we can just dump all of silver non-sorting greedy there 👀

cool

PR #666, opened 2021-01-28
updated info icon

nwatx:
I changed it to the material design icon but feel free to change it if you don't like :D


thecodingwizard:
oops I personally like the old one better :P

HarryWangATX:
hmm, I did like the italics that were there

nwatx:
lol np :p i was thinking it would match the "flat" feel more

PR #670, opened 2021-01-29
remove outdated DMOJ acronym

xiaowuc1:
DMOJ no longer stands for Don Mills Online Judge.


xiaowuc1:
FYI, the DMOJ admins approved of the original commit, but not of this one.

PR #671, opened 2021-01-30
Add solution to usaco-693

topazand:

PR #692, opened 2021-02-04
quick fixes - balanced photos

dongliu0426:
suggestions from #671

### DIFF
--- a/solutions/USACO jan17-balanced-photo.mdx
+++ b/solutions/USACO jan17-balanced-photo.mdx
@@ -23,7 +23,7 @@ We can use [order statistic trees](/gold/PURS/#order-statistic-tree) to achieve 
 
 <Warning>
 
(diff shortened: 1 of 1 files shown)


frodakcin:
I've made some more changed. What you changed was fine.

frodakcin:
I dropped the connection between random access making distance O(N), should probably change wording to add that back

darren-yao (approved):
where's the c++ code/?

dongliu0426:
> where's the c++ code/?

its under the explanation

frodakcin:
I changed phrasing to

"We may **not** use [`std::set`](http://www.cplusplus.com/reference/set/set/) because it does not support indexing. Additionally, since BBSTs do *not* support random access, using operations such as [`std::distance`](http://www.cplusplus.com/reference/iterator/distance/) to index takes linear time."

Is this accurate @bqi343 ?

bqi343:
sure

frodakcin:
ok thanks

frodakcin:
@dongliu0426 are you ok with the changes here?

dongliu0426:
> @dongliu0426 are you ok with the changes here?

yeah

frodakcin:
I'll just merge it

frodakcin (approved):

PR #703, opened 2021-02-06
move even outdegree edges to tree

bqi343:
Since the solution involves finding a spanning tree.

### DIFF
--- a/content/3_Silver/Intro_Tree.mdx
+++ b/content/3_Silver/Intro_Tree.mdx
@@ -27,13 +27,13 @@ export const problems = {
 			'contest/755/problem/C',
 			'Very Easy',
 			false,
-			['Tree', 'DFS']
+			['Tree', 'Connected Components']
 		),
 		new Problem('CF', 'Journey', 'contest/839/problem/C', 'Easy', false, [
 			'Tree',
 			'DFS',
 		]),
-		new Problem('Silver', 'Milk Visits', '968', 'Easy', false, ['DFS'], 'usaco-968'),
+		new Problem('Silver', 'Milk Visits', '968', 'Easy', false, ['Tree', 'Connected Components'], 'usaco-968'),
 		new Problem(
 			'IOI',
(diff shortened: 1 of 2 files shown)


bqi343:
related note - most problems in the DFS / Tree modules should be tagged with "Connected Components" rather than "DFS" since they don't require DFS specifically

content/3_Silver/Intro_Tree.mdx
```diff
@@ -352,10 +361,11 @@ int main()
...
 
-<Problems problems={problems.tree} />
-
-<Info title="Wizard's Tour">
+<Info>
 
-Try solving this problem on a tree first!
+For the last two problems, try solving them on a tree first.
 
 </Info>
```
frodakcin, line 368: I think this is unnecessary info. It's located at the beginning of the problemset but is irrelevant for the majority of the problems.

I say we either move it to the end of the pset, use footnotes, or delete it.

frodakcin: And it might confuse people who just read the module and don't know about spanning trees. On that note, should we define spanning tree in the module?

frodakcin (changes requested):

frodakcin:
> related note - most problems in the DFS / Tree modules should be tagged with "Connected Components" rather than "DFS" since they don't require DFS specifically

Is this issue-worthy?

HarryWangATX:
> > related note - most problems in the DFS / Tree modules should be tagged with "Connected Components" rather than "DFS" since they don't require DFS specifically
> 
> 
> 
> Is this issue-worthy?

i mean it doesnt hurt.

thecodingwizard:
Maybe add that to the pinned tags issue?

content/3_Silver/Intro_Tree.mdx
bqi343: maybe, that can be a separate PR

PR #704, opened 2021-02-06
Convex hull

nwatx:
draft for explanations and cleaned resources for convex hull graham scan section.


content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
 
 <!-- [convex hull](https://en.wikipedia.org/wiki/Convex_hull_algorithms). -->
 
+
+## Introduction
+
+The **Convex Hull** is the subset of points that forms the smallest convex polygon which encloses all points in the set.
+To visualize this, imagine that each point is a pole. Then, imagine what happens if you were to wrap a rope around the outside of all the poles, and then pull infinitely hard, such that the connections between any two points that lie on the edge of the rope are lines. The set of points that touch the rope (and are not colinear) is the convex hull.
```
bqi343, line 91: > colinear

no

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+	<Resource source="Benq" title="Graham Scan Implementation" url="https://github.com/bqi343/USACO/blob/master/Implementations/content/geometry%20(13)/Polygons/ConvexHull2.h" />
+	<Resource source="VisuAlgo" title="Graham Scan Visualization" url="https://visualgo.net/en/convexhull" />
+	<Resource source="UCSD" title="Original Graham Scan Paper" url="http://www.math.ucsd.edu/~ronspubs/72_10_convex_hull.pdf" />
+</Resources>
+
+<Youtube id="9rQMLpQn5xQ"/>
+
+## Tutorial
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around an abitrary pivot $P_0$. This algorithm uses the bottommost (and rightmost if there is a tie), point as $P_0$.
```
bqi343, line 114: > abitrary

no.

You go on to say what $P_0$ is. So it's not arbitrary?

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+	<Resource source="UCSD" title="Original Graham Scan Paper" url="http://www.math.ucsd.edu/~ronspubs/72_10_convex_hull.pdf" />
+</Resources>
+
+<Youtube id="9rQMLpQn5xQ"/>
+
+## Tutorial
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around an abitrary pivot $P_0$. This algorithm uses the bottommost (and rightmost if there is a tie), point as $P_0$. 
+
+To start the creation of the convex hull, we choose 3 points. The pivot, the first, and the last point based on our initial sorting.
```
bqi343, line 116: huh, isn't the pivot the first point

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+
+Then, we maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. In doing so, we only have to keep track of the topmost $2$ points, and the current point we are considering to add to the hull.
+
+There are two only 2 cases for our 3 points: either the turn from $v[i-2] \rightarrow v[i-1] \rightarrow v[i]$ is counterclockwise or not. If the turn is not counterclockwise, then we know that our current set of polygons cannot be convex, and thus we must pop the stack and include the candidate point. Otherwise, we can simply add the candidate point to our stack.
```
bqi343, line 120: > pop the stack and include the candidate point

you don't add the candidate point to the stack at this step

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+</Spoiler>
+
+
+<Spoiler title="Worked Example">
+
+**Example**
+
+Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0)$
+
+Using the steps from our Graham Scan.
```
bqi343, line 137: > Using the steps from our Graham Scan.

colon instead of period

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+    return a+conj((p-a)/d)*d; }
+P foot(const P& p, const Line& l) { 
+    return (p+reflect(p,l))/(T)2; }
+bool p_on_seg(const P& p, const Line& l) {
+    return sgn(cross(l.f,l.s,p)) == 0 && sgn(dot(p-l.f,p-l.s)) <= 0; }
+
+using vi = vector<int>;
+using vP = vector<P>;
+
+vi hullInd(const vP& v) {
```
bqi343, line 236: Whoops I didn't realize this didn't deal with duplicate points. The following works:

```cpp
vi hullInd(const vP& v) {
	int ind = int(min_element(all(v))-begin(v));
	vi cand, C{ind}; F0R(i,sz(v)) if (v[i] != v[ind]) cand.pb(i);
	sort(all(cand),[&](int a, int b) { // sort by angle, tiebreak by distance
		P x = v[a]-v[ind], y = v[b]-v[ind]; T t = cross(x,y);
		return t != 0 ? t > 0 : norm(x) < norm(y); }); 
	each(c,cand) {
		while (sz(C) > 1 && cross(v[end(C)[-2]],v[C.bk],v[c]) <= 0) C.pop_back();
		C.pb(c); }
	return C;
}
```

though this could definitely be more readable (and spaced out)

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+
 <FocusProblem problem={problems.sample[0]} />
 
 ## With Graham Scan
 
-- [cp-algo](https://cp-algorithms.com/geometry/grahams-scan-convex-hull.html)
-- [My Implementation](<https://github.com/bqi343/USACO/blob/master/Implementations/content/geometry%20(13)/Polygons/ConvexHull2.h>)
+<Resources>
+	<Resource source="cp-algo" title="Graham Scan" url="https://cp-algorithms.com/geometry/grahams-scan-convex-hull.html" starred />
+	<Resource source="Benq" title="Graham Scan Implementation" url="https://github.com/bqi343/USACO/blob/master/Implementations/content/geometry%20(13)/Polygons/ConvexHull2.h" />
```
bqi343, line 105: > https://cp-algorithms.com/geometry/grahams-scan-convex-hull.html

I think this is actually monotone chain, not graham scan

bqi343, line 105: Link to this implementation in the solution section instead.

(In general, I think implementations should be kept separate from other resources.)

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,205 @@ export const problems = {
...
+
+Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0)$
+
+Using the steps from our Graham Scan.
+
+1. Sorting the points by counterclockwise angle: Our bottommost point is $(2,0)$, so we then sort all the points by their counterclockwise angle. Our resulting sorted points are $(2,0),(5,3),(3,2),(2,3),(1,2)$.
+
+2. Maintaining the stack:
+
+Our initial stack elements are $v[N-1], v[0], v[1]$, which corresponds to $(1,2), (2,0), (5,3)$. Then, starting from $v[2]$, we process each point.
```
bqi343, line 143: `v` hasn't defined. Maybe put this after the solution?

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
nwatx: ok i moved the whole section to the soluton lmk wyt

content/5_Plat/Convex_Hull.mdx
bqi343: still hasn't been updated

content/5_Plat/Convex_Hull.mdx
bqi343: think it looks better to have the single link in a resources table by itself

content/5_Plat/Convex_Hull.mdx
bqi343: if you could add some space that would be great :P

bqi343 (changes requested):
see comments

content/5_Plat/Convex_Hull.mdx
nwatx: > The pivot (first point), the second point, and the last point based on our initial sorting.

 is what I see.

bqi343: Er it changed, but the last point isn't included in the stack at the start. Also, it's leftmost, not rightmost.

nwatx: huh, I was reading through cp4 and it mentioned it but since we are using your implementation i changed it to match (also wouldn't it not change the hull regarding left or right as long as its an extreme)

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,216 @@ export const problems = {
...
+
+<LanguageSection>
+
+<CPPSection>
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around a pivot $P_0$, breaking ties by distance. This algorithm uses the leftmost (and bottommost if there is a tie), point as $P_0$. 
+
+To start the creation of the convex hull, we choose 2 points. The pivot (first point), and the second point based on our initial sorting.
+
+Then, we maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. In doing so, we only have to keep track of the topmost $2$ points, and the current point we are considering to add to the hull.
```
bqi343, line 125: > we only have to keep track of the topmost $2$ points

it's not clear what this means. technically, we're "keeping track" of the entire stack, which could have size > 2.

content/5_Plat/Convex_Hull.mdx
bqi343: ^

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,216 @@ export const problems = {
 
 <!-- [convex hull](https://en.wikipedia.org/wiki/Convex_hull_algorithms). -->
 
+
+## Introduction
+
+The **Convex Hull** is the subset of points that forms the smallest convex polygon which encloses all points in the set.
+To visualize this, imagine that each point is a pole. Then, imagine what happens if you were to wrap a rope around the outside of all the poles, and then pull infinitely hard, such that the connections between any two points that lie on the edge of the rope are lines. The set of points that touch the rope is the convex hull.
```
bqi343, line 91: Why is line 91 not on the same line as line 90?

content/5_Plat/Convex_Hull.mdx
bqi343: don't need all these geo funcs, only keep the ones that are used (and explain what they do).

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+<LanguageSection>
+
+<CPPSection>
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around a pivot $P_0$, breaking ties by distance. This algorithm uses the leftmost (and bottommost if there is a tie), point as $P_0$. 
+
+To start the creation of the convex hull, we choose 2 points. The pivot (first point), and the second point based on our initial sorting.
+
+Then, we maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. As a result, we only have to check the candidate point with the topmost 2 points because this counterclockwise property is invariant.
```
bqi343, line 124: > As a result, we only have to check the candidate point with the topmost 2 points

with -> against

> because this counterclockwise property is invariant.

not sure what this means

content/5_Plat/Convex_Hull.mdx
bqi343: ^

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
nwatx: meant that it always is counterclockwise but i suppose that was redundant w/ b4 so i removed it

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,204 @@ export const problems = {
...
+
+<LanguageSection>
+
+<CPPSection>
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around a pivot $P_0$, breaking ties by distance. This algorithm uses the leftmost (and bottommost if there is a tie), point as $P_0$. 
+
+To start the creation of the convex hull, we choose 2 points. The pivot (first point), and the second point based on our initial sorting.
+
+Then, we maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. As a result, we only have to check the candidate point against the topmost 2 points.
```
bqi343, line 124: > Then, we maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex.

Should explicitly say here that for the set of points to be convex, every three consecutive points of the stack must form a counterclockwise turn.

content/5_Plat/Convex_Hull.mdx
bqi343: haven't addressed all of these. Spacing is messed up - use tabs.

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
bqi343: actually it's probably best to mention this *before*

> As a result, we only have to check the candidate point against the topmost 2 points.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+
+In order for the hull to be convex, every three consecutive points of the stack must contain a counterclockwise turn.
+
+Denote our stack as $\texttt{C}$, the top element of $\texttt{C}$ as $\texttt{C}[i]$ and $\texttt{cand}[i]$ as the $i$-th sorted point.
```
bqi343, line 128: > the top element of $\texttt{C}$ as $\texttt{C}[i]$

why is $i$ is used twice with different meanings?

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+
+There are 2 cases for our 3 points: either the turn from $\texttt{C}[i-1] \rightarrow \texttt{C}[i] \rightarrow \texttt{cand}[i]$ is counterclockwise or not. If the turn is not counterclockwise, then we know that our current set of points cannot be convex, and thus we pop the top element off $\texttt{C}$ and proceed to check convexity of the next candidate point. Otherwise the turn is counterclockwise, and thus we include it in our hull.
```
bqi343, line 130: > proceed to check convexity of the next candidate point

actually, we continue checking the same point

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+![Illustration of Convex Hull](./assets/convexhullpts.jpg)
+
+</Spoiler>
+
+
+<Spoiler title="Worked Example">
+
+**Example**
+
+Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0)$
```
bqi343, line 145: Add > before this and add a period at the end.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+<Spoiler title="Illustration">
+
+![Illustration of Convex Hull](./assets/convexhullpts.jpg)
+
+</Spoiler>
+
+
+<Spoiler title="Worked Example">
+
+**Example**
```
bqi343, line 143: Remove.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+
+Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0)$
+
+Using the steps from our Graham Scan:
+
+1. Sorting the points by counterclockwise angle: Our leftmost point is $(1,2)$, so we then sort all the points by their counterclockwise angle. If the angles are the same, we can tiebreak by distance. Our resulting sorted points are $(2,0),(3,2),(5,3),(2,3)$ with initial pivot $(1,2)$.
```
bqi343, line 149: >  we can tiebreak by distance

example doesn't include collinear points. Diagram would be helpful.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+
+2. Maintaining the stack:
+
+Let's denote the $i$-th sorted point as $\texttt{pts}[i]$.
+
+Our initial stack elements are $(1,2),(2,0)$. Then, starting from $\texttt{pts}[2]$, we process each point.
```
bqi343, line 155: In the code, `pts` isn't sorted.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,217 @@ export const problems = {
...
+
+Then, we check the turn $(2,0),(3,2),(5,3)$ which is clockwise. Hence, we pop the stack until the counterclockwise property can be satisfied. In this case, only $(5,3)$ is popped. Afterwards, we push $(2, 3)$ to our stack because it satisfies the counterclockwise property.
+
+Our stack now is then $(1,2),(2,0),(5,3)$.
+
+Then, we check whether the turn between $(2,0), (5,3), (2,3)$ is counterclockwise, which it is, so we push $(2,3)$ to our stack.
+
+Since we have processed all points, we have completed the cycle, and our convex hull is complete.
+
+Our stack now is then $(1,2),(2,0),(5,3),(2,3)$, which is the convex hull of all our points.
```
bqi343, line 169: Say "vertices of the convex hull" instead of "the convex hull" because the convex hull consists of all points within the border of the polygon bounded by these points.

content/5_Plat/Convex_Hull.mdx
bqi343: maybe https://codeforces.com/blog/entry/70330 is helpful

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+> Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0),(6,2)$.
+
+[Animated](https://convexhull.tiiny.site/)
+
+Using the steps from our Graham Scan:
+
+1. Sorting the points by counterclockwise angle: Our leftmost point is $(1,2)$, so we then sort all the points by their counterclockwise angle. If the angles are the same, we can tiebreak by distance. Our resulting sorted points are $(2,0),(6,2),(3,2),(5,3),(2,3)$ with initial pivot $(1,2)$. Note that $(6,2), (3,2)$, are colinear but since $(6,2)$ is further away it breaks the tie, thus placing it before $(3,2)$.
+
+2. Maintaining the stack:
```
bqi343, line 150: everything after this needs to be properly indented.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+![Illustration of Convex Hull](./assets/convexhullpts.jpg)
+
+</Spoiler>
+
+<Spoiler title="Worked Example">
+
+> Find the convex hull of the points $(1,2),(2,3),(5,3),(3,2),(2,0),(6,2)$.
+
+[Animated](https://convexhull.tiiny.site/)
```
bqi343, line 144: I prefer the method of animation used in the video:

https://youtu.be/hAZ3GaEzA04?t=692

related code can be found here: https://lukakalinovcic.github.io/geodeb/gallery/hull.html

(though the rainbow is distracting)

(also, you should link to geodeb)

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+<CPPSection>
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around a pivot $P_0$, breaking ties by distance. This algorithm uses the leftmost (and bottommost if there is a tie), point as $P_0$. 
+
+We maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. As a result, we only have to check the candidate point against the topmost 2 points.
+
+To start the creation of the convex hull, we choose 2 points. The pivot (first point), and the second point based on our initial sorting.
+
+In order for the hull to be convex, every three consecutive points of the stack must contain a counterclockwise turn.
```
bqi343, line 126: Specify that CCW turn with (a,b,c) means that you need to rotate (b-a) CCW to make it parallel to (c-b). (at least, that's the definition I was using)

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+Denote our stack as $\texttt{C}$, the top element of $\texttt{C}$ as $\texttt{C}[i]$ and $\texttt{cand}[j]$ as the $j$-th sorted point.
+
+There are 2 cases for our 3 points: either the turn from $\texttt{C}[i-1] \rightarrow \texttt{C}[i] \rightarrow \texttt{cand}[j]$ is counterclockwise or not. If the turn is not counterclockwise, then we know that our current set of points cannot be convex, and thus we pop the top element off $\texttt{C}$ and proceed to check convexity of the point until the turn is counterclockwise. Otherwise the turn is counterclockwise, and thus we include it in our hull.
+
+To understand why, consider those three points: the top two points in the stack, and our candidate point. Since our turn is counterclockwise, we are looking for such points that make a counterclockwise turn. It is easy to see that when a clockwise turn it is made, the polygon can no longer be convex.
```
bqi343, line 132: I think line 132 can be removed, it doesn't seem to add anything.

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+
+The Graham Scan algorithm works in 3 steps. First, it sorts all of the $n$ points by their counterclockwise angle around a pivot $P_0$, breaking ties by distance. This algorithm uses the leftmost (and bottommost if there is a tie), point as $P_0$. 
+
+We maintain a stack containing the points of the convex hull, such that the set of points in the stack is convex. As a result, we only have to check the candidate point against the topmost 2 points.
+
+To start the creation of the convex hull, we choose 2 points. The pivot (first point), and the second point based on our initial sorting.
+
+In order for the hull to be convex, every three consecutive points of the stack must contain a counterclockwise turn.
+
+Denote our stack as $\texttt{C}$, the top element of $\texttt{C}$ as $\texttt{C}[i]$ and $\texttt{cand}[j]$ as the $j$-th sorted point.
```
bqi343, line 128: should use more informative variable names :)

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
```diff
@@ -85,22 +84,197 @@ export const problems = {
...
+	vi cand, C{ind};
+    F0R(i,sz(v)) if (v[i] != v[ind]) cand.pb(i);
+
+	sort(all(cand),[&](int a, int b) { // sort by angle, tiebreak by distance
+        P x = v[a]-v[ind], y = v[b]-v[ind];
+        T t = cross(x,y);
+		return t != 0 ? t > 0 : norm(x) < norm(y); }
+    );
+
+	trav(c,cand) { // for every point
```
bqi343, line 230: indentation is still messed up

bqi343 (changes requested):

content/5_Plat/Convex_Hull.mdx
bqi343: @nwatx indentation is still a mix of spaces / tabs D:

bqi343 (changes requested):
see comment above

bqi343 (approved):
tbh I don't think this isn't any better than the Wikipedia explanation. But probably ok to merge?

dolphingarlic (approved):

PR #705, opened 2021-02-06
haybale stacking, grass planting

### LINKED ISSUE #606: Dead USACO problems
USACO Problems earlier than 2012 US Open seem to no longer accept Analysis Mode submissions.
[Some of these problems are used in USACO Guide](https://usaco.guide/plat/hld#problem-http://www.usaco.org/index.php?page=viewproblem2&cpid=102).

We should probably link to a mirror site such as [Luogu](https://www.luogu.com.cn/problem/P3038) or contact USACO staff.

bqi343:
closes #606

### DIFF
--- a/content/3_Silver/Prefix_Sums.mdx
+++ b/content/3_Silver/Prefix_Sums.mdx
@@ -415,11 +415,13 @@ print(psum([1,2,3,4,5]))
 
 <Problems problems={problems.cum} />
 
-<Warning>
+[Haybale Stacking Submission Link](https://www.acmicpc.net/problem/5912)
+
+<!-- <Warning>
 
 The last problem isn't submittable on USACO. Download the test data and use the script provided [here](/general/debugging-lang#stress-testing) to test your solution.
 
-</Warning>
+</Warning> -->
 
 ## 2D Prefix Sums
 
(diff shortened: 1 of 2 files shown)


frodakcin (commented):
It feels kinda out of place to have an extra link w/o context. Would it be better to wrap it in `Info` or `Warning`? Maybe we can add a footnote instead?

HarryWangATX:
yea i think we should still put the warning there, but add the link inside of the warning

frodakcin:
Actually I think a footnote would be very fitting here, although IDK if superscript is supported

bqi343:
Idt you can add a footnote in the middle of a problem table though. @thecodingwizard what do you think?

thecodingwizard:
Uhh this might be possible (I'm thinking we can add an "info" icon right after the problem name, and then when the user hovers over the icon (or click on mobile) they get a popup).

This might be a bit nontrivial to implement though (might need to modify problem constructor) -- would recommend just leaving it as a comment below the table for now

frodakcin (approved):

frodakcin:
I'll wait for a secondary review, in case someone has a better idea

bqi343:
looks fine, I'll just merge

PR #706, opened 2021-02-06
Import user data v2

thecodingwizard:


bqi343:
@thecodingwizard I think it's fine to merge

PR #708, opened 2021-02-06
feat: copy url to clipboard

bqi343:
not sure what to change so that page doesn't scroll when URL is clicked


thecodingwizard:
now the page doesn't scroll and the page URL doesn't change (the URL is copied to the clipboard directly)

Also, is the setTimeout necessary?

thecodingwizard (approved):

PR #709, opened 2021-02-07
Added New Resources

HarryWangATX:


content/4_Gold/DSU.mdx
```diff
@@ -91,7 +91,8 @@ export const problems = {
 ## Resources
 
 <Resources>
-	<Resource source="CF" title="EDU: DSU Steps 1 & 2" url="https://codeforces.com/edu/course/2/lesson/7/1/practice" starred> </Resource>
+	<Resource source="CF" title="EDU: DSU" url="https://codeforces.com/edu/course/2/lesson/7" starred> video explanation of DSU </Resource>
+	<Resource source="CF" title="EDU: DSU Steps 1 & 2" url="https://codeforces.com/edu/course/2/lesson/7/1/practice" starred> practice problems for DSU </Resource>
```
dolphingarlic, line 95: I think you can actually just get rid of this link since it's already in the above one. Otherwise, it looks good to merge

HarryWangATX: i dont think i can remove urls with these cf sources because i think ill default to parsing the title which returns error? or do you mean deleting the second resource all together?

thecodingwizard: I think he means deleting it altogether (and in description of first resource maybe say "video + problems" or something like that)

PR #714, opened 2021-02-08
Elaborate on fast i/o

bqi343:
Resolves https://forum.usaco.guide/t/different-types-of-fast-io/729 (not complete)

If someone could find an example for `cin.tie(0)` that would be great.

### DIFF
--- a/content/1_General/Fast_IO.mdx
+++ b/content/1_General/Fast_IO.mdx
@@ -5,8 +5,6 @@ author: Benjamin Qi, Nathan Chen
 description: 'Speeding up I/O speeds can make a substantial difference in problems with large inputs.'
 ---
 
-
-
 export const problems = {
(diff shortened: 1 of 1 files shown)


dongliu0426:
should I add another fastio method using `getchar()` and `putchar()`?

bqi343:
Is it faster?

dongliu0426:
it's faster than `scanf/printf` and `cin/cout (unsynced)`. not sure if it's faster than `fread/fwrite` though.

dongliu0426:
oh, its slower than `fread` (takes `99ms`).

thecodingwizard:
ignore the failing check btw -- if the content is okay, this PR is good to merge.

thecodingwizard (approved):

PR #723, opened 2021-02-08
Studying Algorithms

### LINKED ISSUE #721: Add https://usaco.guide/silver/greedy?lang=cpp#example-studying-algorithms to codeforces gym?

dongliu0426:
closes #721

### DIFF
--- a/content/3_Silver/Greedy.mdx
+++ b/content/3_Silver/Greedy.mdx
@@ -11,6 +11,7 @@ frequency: 3
 ---
 
 export const problems = {
+	studyalgo: new Problem('CF', 'Studying Algorithms', 'https://codeforces.com/gym/102951/problem/B', 'Very Easy', false, [], ''),
 	movie: [new Problem('CSES', 'Movie Festival', '1629', 'Easy', false, [], '')],
 	cses: [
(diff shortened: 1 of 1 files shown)


frodakcin:
I'm reading the module here https://usaco-guide-nmydjsszn.vercel.app/silver/greedy?lang=cpp and think it might be better to copy the statement in the guide as well as providing a submission link.

frodakcin:
What do you think? Also the code isn't copy-pastable complete -- let's see if people struggle to solve in on cf to see if we need to change that.

frodakcin:
> I'm reading the module here https://usaco-guide-nmydjsszn.vercel.app/silver/greedy?lang=cpp and think it might be better to copy the statement in the guide as well as providing a submission link.

I guess this would be impractical for certain problems, especially those with long statements, but I think in this case it may be preferable

frodakcin:
@HarryWangATX @andrewwangva Ready for review now (I probably should have marked as draft earlier)

HarryWangATX:
Alr reviewing

HarryWangATX (approved):
Looks good lol

Issue #729, opened 2021-02-09
Ordered Set Implementation


bqi343:
- Add set solution for https://usaco.guide/silver/intro-ordered#problem-https://codeforces.com/contest/702/problem/C. Internal sol doesn't use sets ...

bqi343:
closed by #743

PR #731, opened 2021-02-09
Cellular Network C++ Solution Using Sets

maggie-j-liu:
I added a C++ solution using sets to CF Cellular Network. #729

### DIFF
--- a/solutions/CF Cellular Network.mdx
+++ b/solutions/CF Cellular Network.mdx
@@ -11,62 +11,7 @@ author: Benjamin Qi, Maggie Liu, Nathan Wang
 
 <CPPSection>
 
-## Solution 1 - Using a Set
-
-We use the set $\texttt{towers}$ to store the coordinates of the cellular towers. For each city, we want to find the closest tower and calculate the distance. The minimal $r$ will be the maximum of all these distances. To find the closest tower to a certain city, we check the towers to the left and right and determine which one is closer.
-
-Looping through the cities, we use `lower_bound` to find the closest tower to the right of the city. If this tower is found, calculate the distance between this tower and the city. If the tower is not at the beginning of the set, the previous tower in the set will be the tower to the left of the city. So, $\texttt{dist}$ will be the minimum of the distance to the tower on the right and the distance to the tower on the left. Then, set $r$ to be the maximum of itself and $\texttt{dist}$.
-
-```cpp
-#include <iostream>
-#include <set>
-#include <algorithm>
-using namespace std;
-
-int main()
-{
-	int n, m;
-	cin >> n >> m;
-	int cities[n];
-	set<int> towers;
-	int tower;
-	for (int i = 0; i < n; i++)
-	{
-		cin >> cities[i];
-	}
-	for (int i = 0; i < m; i++)
-	{
-		cin >> tower;
-		towers.insert(tower);
-	}
-	int r = 0;
-	for (int i = 0; i < n; i++)
-	{
-		int dist = 2e9 + 1;
-		//find closest tower to the right of the city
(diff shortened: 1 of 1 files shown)


solutions/CF Cellular Network.mdx
```diff
@@ -2,15 +2,72 @@
...
+
+```cpp
+#include <iostream>
+#include <set>
+#include <algorithm>
+using namespace std;
+
+int main()
+{
+    int n, m;
```
bqi343, line 28: convert to tabs as mentioned here: https://usaco.guide/general/contributing

bqi343 (changes requested):

frodakcin (approved):
Looks good.

solutions/CF Cellular Network.mdx
```diff
@@ -2,62 +2,64 @@
...
+
+<CPPSection>
+
+## Solution 1 - Binary Search on a Sorted Array
 
 ```cpp
 #include <bits/stdc++.h>
 using namespace std;
 
 // returns the first index in the array that is >= value, or arr.size() if no such index exists
```
bqi343, line 20: > // returns the first index in the array that is >= value, or arr.size() if no such index exists

this seems wrong ... @thecodingwizard I think you added this?

bqi343 (changes requested):

solutions/CF Cellular Network.mdx
bqi343: actually it seems right but confusing ... seems odd to end a loop when `lo` *exceeds* `hi`?

thecodingwizard: D: it highkey took me 30 minutes to write that binary search function; our implementations are highly inconsistent across solutions / guide / slides (ex. https://usaco.guide/silver/binary-search-ans?lang=cpp#introduction)

We should standardize binary search impl; should we just do this?

```cpp
int lo = 0, hi = n-1, ans = [defaultAns];

while (lo < hi) {
    int mid = (lo+hi)/2;
    if (f(mid)) lo = mid + 1;
    else hi = mid;
}
```

That was how I learned it at least :/

frodakcin: My go-to:

```
int l=-1, r=n; // [0, l]: valid answer --- [r, n]: invalid answers
while(l<r)
{
  int m=(l+r)/2;
  if(ok(m)) l=m;
  else r=m;
}
return r;
```

frodakcin: what i wrote ^^^ is equivalent to upper_bound in cpp I believe

bqi343: >  ans = [defaultAns];

why do you need an `ans` variable?

```
int l=-1, r=n; // [0, l]: valid answer --- [r, n]: invalid answers
while(l<r)
{
  int m=(l+r)/2;
  if(ok(m)) l=m;
  else r=m;
}
return r;
```

If the return value is supposed to be `-1` then won't this loop infinitely?

bqi343: > https://usaco.guide/silver/binary-search-ans?lang=cpp#introduction

I think we should just remove the first implementation from here

thecodingwizard: > why do you need an ans variable?

Err I think I typically do `if (ok(mid)) l = mid, ans = mid;` or something like that. It's probably not needed though.

> I think we should just remove the first implementation from here

I think the while loop impl is easier to approach compared to the for loop though?

frodakcin: > If the return value is supposed to be -1 then won't this loop infinitely?

yes oops.
I meant `while(l+1<r)`

PR #737, opened 2021-02-09
Flood Fill - reorder, change intro

### LINKED ISSUE #728: Flood Fill - Unnecessary Code
It's not necessary to have both the "Implementation" and the solution to Counting Roads. The two should be merged.

bqi343:
closes #728

### DIFF
--- a/content/3_Silver/Flood_Fill.mdx
+++ b/content/3_Silver/Flood_Fill.mdx
@@ -118,8 +118,6 @@ export const problems = {
 
 <!-- Mooyo Mooyo: Not a sorting problem, but you can use sorting to simulate gravity. - Write a custom comparator which puts zeroes at the front and use `stable_sort` to keep the relative order of other elements the same. -->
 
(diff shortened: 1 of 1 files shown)


content/3_Silver/Flood_Fill.mdx
```diff
@@ -326,7 +350,7 @@ public static void main(String[] args){
```
frodakcin, line 350: I think it may be clearer if we have the red square backtrack instead of jumping from the bottom square to the top. It may not be intuitive why the red square knows to go there.

frodakcin: The only problem would be that this would add like 4 more squares, making it really long. Maybe we can have 2 or 3 square on each row?

frodakcin (commented):
Everything else looks good though.

content/3_Silver/Flood_Fill.mdx
bqi343: this can be done in separate PR

Issue #711, opened 2021-02-07
Functional Graphs module java code is trash: transcribe cpp version


frodakcin:
Context: https://github.com/cpinitiative/usaco-guide/pull/710

Python code might be trash too

frodakcin:
I think this couldve been closed 9 days ago

PR #739, opened 2021-02-10
Func Graphs: fix java impl & add py impl

frodakcin:
https://github.com/cpinitiative/usaco-guide/issues/711


frodakcin:
Todo: remove macros from cpp impl

content/3_Silver/Func_Graphs.mdx
```diff
@@ -390,66 +396,84 @@ import java.util.*;
 
 public class badge
 {
-	static InputReader in = new InputReader(System.in);
-	static PrintWriter out = new PrintWriter(System.out);
+	static BufferedReader reader;
+	static PrintWriter writer;
 	public static final int MN = 1010;
 	public static int N;
```
andrewwangva, line 402: why is 'N'a global variable? Should be fine to leave it as local since you don't use it anywhere else.

content/3_Silver/Func_Graphs.mdx
```diff
@@ -390,66 +396,84 @@ import java.util.*;
 
 public class badge
```
andrewwangva, line 397: Why is the focus problem "badge" at the top of the page and not right above the solution?

content/3_Silver/Func_Graphs.mdx
```diff
@@ -390,66 +396,84 @@ import java.util.*;
...
+			p[i]=Integer.parseInt(st.nextToken());
+		reader.close();
+
 		for (int i=1;i<=N;++i)
-			if(!v[i])
-				dfs(i);
+			dfs(i); // in_cycle is always reset to false
+
 		for (int i=1;i<=N;++i)
 		{
```
andrewwangva, line 432: No need for parentheses here if you don't use it in the previous for loop

andrewwangva (approved):

PR #743, opened 2021-02-10
(quick fix) Delete sentence fragment in Cellular Network solution

maggie-j-liu:


bqi343:
closes #729

solutions/CF Cellular Network.mdx
```diff
@@ -113,7 +113,7 @@ int main() {
...
 
-Looping through the cities. For each city, we use `lower_bound` to find the closest tower to the right of the city. If this tower is found, calculate the distance between this tower and the city. If the tower is not at the beginning of the set, the previous tower in the set will be the tower to the left of the city. Define $\texttt{dist}$ to be the minimum of the distance to the tower on the right and the distance to the tower on the left. Then, set $r$ to be the maximum of itself and $\texttt{dist}$.
+For each city, we use `lower_bound` to find the closest tower to the right of the city. If this tower is found, calculate the distance between this tower and the city. If the tower is not at the beginning of the set, the previous tower in the set will be the tower to the left of the city. Define $\texttt{dist}$ to be the minimum of the distance to the tower on the right and the distance to the tower on the left. Then, set $r$ to be the maximum of itself and $\texttt{dist}$.
```
frodakcin, line 116: I'd prefer to link all stl functions/objects

```suggestion
For each city, we use [`lower_bound`](http://www.cplusplus.com/reference/algorithm/lower_bound/) to find the closest tower to the right of the city. If this tower is found, calculate the distance between this tower and the city. If the tower is not at the beginning of the set, the previous tower in the set will be the tower to the left of the city. Define $\texttt{dist}$ to be the minimum of the distance to the tower on the right and the distance to the tower on the left. Then, set $r$ to be the maximum of itself and $\texttt{dist}$.
```

or https://en.cppreference.com/w/cpp/algorithm/lower_bound

frodakcin (changes requested):

frodakcin (approved):

PR #745, opened 2021-02-10
sorting & searching

bqi343:
- move bitwise to Bronze (doesn't belong, but it seems really random for it to be in the middle of silver, and at least it's mentioned in one of the Bronze modules). 
- Move binsearch into sorting section.


frodakcin (approved):
Looks good. interesting that psum has an entire section all to itself

PR #747, opened 2021-02-10
Added no time to paint

HarryWangATX:


bqi343:
I purposely tagged this problem in the [division table](https://github.com/cpinitiative/usaco-guide/blob/master/solutions/1_extra_usaco_probs.ts) but didn't add this to the module because I think the module has enough introductory prefix sum problems. Plus, it's good to have some additional silver probs left to go through after you go through all of the modules.

(not sure if these are good reasons)

HarryWangATX:
> I purposely tagged this problem in the [division table](https://github.com/cpinitiative/usaco-guide/blob/master/solutions/1_extra_usaco_probs.ts) but didn't add this to the module because I think the module has enough introductory prefix sum problems. Plus, it's good to have some additional silver probs left to go through after you go through all of the modules.
> 
> (not sure if these are good reasons)


Got it. I'll close.

PR #751, opened 2021-02-11
move rectangle geometry into ad hoc

thecodingwizard:
oops still not sure whether we should do this or keep rect geo as its own module; just ignore this PR for now


thecodingwizard:
nvm, seems like consensus is to keep rect geo as its own module

PR #752, opened 2021-02-11
move rectangle geometry to additional category

thecodingwizard:
I don't think rectangle geometry belongs under "fundamentals," but if other people feel otherwise please feel free to say so :O

### DIFF
--- a/content/ordering.ts
+++ b/content/ordering.ts
@@ -71,7 +71,7 @@ const MODULE_ORDERING: { [key in SectionID]: Chapter[] } = {
   bronze: [
     {
       name: 'Fundamentals',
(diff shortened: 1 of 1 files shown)


HarryWangATX (approved):

HarryWangATX:
yea i do feel like rectgeo ahouldnt be in fundamentals at least from the things on fundamentals. im fine with this tho

bqi343:
I still think rect geo should be the second module people do.

frodakcin:
Surely it must come before ad-hoc

thecodingwizard:
okay sure (ig rect geo problems are good intro problems anyway)

PR #754, opened 2021-02-11
Ad Hoc + Update Times

bqi343:
- better ad hoc definition
 - capitalize Bronze
 - update times are displayed for longer

### DIFF
--- a/content/1_General/Notes_CPP_DS.mdx
+++ b/content/1_General/Notes_CPP_DS.mdx
@@ -86,15 +86,11 @@ int addElement() {
     return res.size()-1;
 }
 
-void rec(int x) {
+int main() {
     for (int i = 0; i < 10; ++i) {
         res[i] = addElement();
         cout << i << " " << res[i] << "\n";
     }
-}
-
-int main() {
-    rec(0);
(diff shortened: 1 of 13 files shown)


bqi343:
Also, there's nothing that implies that ad hoc problems must be difficult (ex. [this](https://usaco.guide/general/expected#problem-http://www.usaco.org/index.php?page=viewproblem2&cpid=591) would be considered ad hoc), though the problems given in the module are definitely on the harder end of Bronze. Should be noted within the module.

content/1_General/Notes_CPP_DS.mdx
```diff
@@ -0,0 +1,165 @@
...
+#include <bits/stdc++.h>
+using namespace std;
+
+vector<int> res{-1};
+int addElement() {
+    res.push_back(-1);
+    return res.size()-1;
+}
+
+void rec(int x) {
```
thecodingwizard, line 89: is `x` needed?

thecodingwizard (changes requested):

thecodingwizard (approved):

PR #755, opened 2021-02-11
Splitting Up Psums section and rearranging it

HarryWangATX:


frodakcin (commented):
Maybe it'd be better to call the first one "Introduction to Prefix Sums" (like the way many of the bronze modules are "Introduction to ..."

The second one is probably fine. Maybe possibly one of "More Prefix Sums" "Extending Prefix Sums" "Additional Applications of Prefix Sums" or something... I'm not sure what would be best to call it

bqi343 (approved):

Farrius (approved):

PR #735, opened 2021-02-09
Broken profile DP tutorial

dolphingarlic:
I basically introduced the concept and then moved CSES Count Tilings into the module

Closes https://github.com/cpinitiative/usaco-guide/issues/366

PR #771, opened 2021-02-14
Fix CSES Counting Tilings

hieplpvip:


dolphingarlic:
Nice catch! I'm going to close this PR because there's already another PR where I moved this solution into a module (https://github.com/cpinitiative/usaco-guide/pull/735). However, I did use your fixes :)

PR #775, opened 2021-02-15
replace four spaces with tab in code block

### LINKED ISSUE #768: Copy paste code w/ tabs instead of spaces
ex. https://github.com/cpinitiative/usaco-guide/blob/master/content/1_General/IO.mdx#java is formatted w/ tabs but https://usaco.guide/general/io?lang=java#java copy-pastes as spaces. Should be an option to copy as tabs (and maybe specify the tab width?)

thecodingwizard:
note: this replaces *all* instances of four spaces with a tab, regardless of whether the .mdx file was using four spaces or a tab.

this probably will create some issues:

- If the code is supposed to use a mixture of tabs and spaces, then all the spaces will be converted to tabs.
- If the code is poorly formatted (with two/three spaces), the tabs will get messed up.

closes #768

### DIFF
--- a/src/components/markdown/CodeBlock/CodeBlock.tsx
+++ b/src/components/markdown/CodeBlock/CodeBlock.tsx
@@ -34,9 +34,10 @@ const renderTokens = (tokens, getLineProps, getTokenProps) => {
       <Line key={i} {...getLineProps({ line, key: i })}>
         <LineNo data-line-number={i + 1} />
         <LineContent>
-          {line.map((token, key) => (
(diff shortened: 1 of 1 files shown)


bqi343 (approved):

thecodingwizard:
Currently set as draft because:

- Some implementations (especially Java - Gold) are messed up (ex. https://usaco.guide/gold/sp?lang=java#mathcalomlog-n-implementation)
- There's currently no option to configure this in the settings modal

But technically it can be merged

PR #776, opened 2021-02-15
Update IO

bqi343:
- replace fake variables with real example
 - merge methods (`Kattio` is just a wrapper around `BufferedReader` + `PrintWriter`, similarly for C++)
 - `cin` is not a function

### DIFF
--- a/content/1_General/IO.mdx
+++ b/content/1_General/IO.mdx
@@ -4,12 +4,12 @@ title: Input & Output
 author: Darren Yao, Benjamin Qi, Nathan Wang, Allen Li
 description: Demonstrates how to read input and print output for USACO contests, including an example problem.
 prerequisites:
- - data-types
+  - data-types
 ---
 
 export const problems = {
-	cses: [new Problem('CSES', 'Weird Algorithm', '1068', 'Very Easy')],
-	fence: [new Problem('Bronze', 'Fence Painting', '567', 'Very Easy')],
+	weird: new Problem('CSES', 'Weird Algorithm', '1068', 'Very Easy'),
+	fence: new Problem('Bronze', 'Fence Painting', '567', 'Very Easy'),
 };
 
 <Info title="Update">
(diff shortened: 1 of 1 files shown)


content/1_General/IO.mdx
```diff
@@ -547,9 +555,9 @@ You must rename the `.in` and `.out` files depending on the problem. For example
 
 <CPPSection>
 
-### Method 1 - [`freopen`](https://www.cplusplus.com/reference/cstdio/freopen/)
+### Method 1 - `freopen`
```
thecodingwizard, line 559: no link to cplusplus.com? (it's fine either way; just wanted to make sure it was intentional)

same thing for <cstdio> below

content/1_General/IO.mdx
```diff
@@ -109,23 +123,71 @@ int main() {
...
+		System.out.println(a + b + c);
+	}
+}
+```
+
+This works, but `Scanner` and `System.out.print` are slow when we have to handle inputting and outputting tens of thousands of lines.
+
+### Method 2 - [`BufferedReader`](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/io/BufferedReader.html) and [`PrintWriter`](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/io/PrintWriter.html)
+
+These are faster because they buffer the input and output and handle it all at once as opposed to parsing each line individually. However, `BufferedReader` is harder to use than `Scanner`. It has quite a few more methods and the `io` library must be imported for its use as well.
```
frodakcin, line 159: Possibly explain `StringTokenizer` (https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/StringTokenizer.html)?

frodakcin (approved):
Looks good. Maybe address comment but that's a separate issue entirely.

frodakcin (approved):
I think it's good.

PR #785, opened 2021-02-17
Update Sorting_Custom.mdx

jessechoe10:


frodakcin (changes requested):
From Andrew: Make "Meetings" Hard instead of changing Triangles

jessechoe10:
> From Andrew: Make "Meetings" Hard instead of changing Triangles

Sure, but the reason I made Triangles as "Very Hard" is because the official solution makes no sense.

The part that says s(i+1)=s(i) + (2i-1) * (x(i+1)-x(i)) provided no intuition as to why it's (2i-1)...

I think until that part is explained this should be considered "very hard".

dolphingarlic:
Is the (2i - N) part not just from rearranging and simplifying the algebra? You want to add i of (x_i - x_{i - 1}) and subtract (N-i) of (x_i - x_{i - 1}) to go from s_{i - 1} to s_i, so that adds up to (2i - N)(x_i - x_{i - 1})

jessechoe10:
> Is the (2i - N) part not just from rearranging and simplifying the algebra? You want to add i of (x_i - x_{i - 1}) and subtract (N-i) of (x_i - x_{i - 1}) to go from s_{i - 1} to s_i, so that adds up to (2i - N)(x_i - x_{i - 1})

I think this should have an internal sol. Many people on reddit are confused about this.

dolphingarlic:
Imo this one line isn't enough to justify a whole new internal solution

jessechoe10:
> Imo this one line isn't enough to justify a whole new internal solution

I mean the code is kinda weird as well.

andrewwangva:
> > Imo this one line isn't enough to justify a whole new internal solution
> 
> I mean the code is kinda weird as well.

I think once you get that observation, everything else is pretty trivial. The intuition to find a relation between each point is also pretty logical since it's not hard to realize you need to find a relation between the points to improve it from N^2. Compared to Meetings, the observations are much easier to find In my opinion.

HarryWangATX:
yea, observations for meetings were much harder to find because the problem also tricks you to think in the other direction. For triangles, it was more obvious to find the observations needed imo,

jessechoe10:
> yea, observations for meetings were much harder to find because the problem also tricks you to think in the other direction. For triangles, it was more obvious to find the observations needed imo. I think hard is a good indicator of triangles, making it very hard may scare people away from doing a great problem

Okay, I'd agree with this. Will Meetings be categorized as hard or very hard. If it's hard, then I'll close this PR.

PR #788, opened 2021-02-17
added java solution to cses forest queries problem

JyVenom:
I added my solution to the Forest Query Problem. My submission can be found at https://cses.fi/problemset/hack/1652/entry/1469440/. 
If you want, I can use Buffered Reader for input instead. 
Also, in my Reader class, the buffer size can (and prob should) be shrunk to 2^16 or 1<<16, and the neg condition and checks can be removed from the nextInt method. 
 
Random sidenote, its currently the fastest JAVA sol.

### DIFF
--- a/content/3_Silver/Prefix_Sums_2.mdx
+++ b/content/3_Silver/Prefix_Sums_2.mdx
@@ -447,7 +447,6 @@ import java.util.StringTokenizer;
 
 public class ForestQueries {
 	static Kattio io = new Kattio();
-
 	public static void main(String[] args) throws IOException {
 		int N = io.nextInt() + 1;
 		int q = io.nextInt();
@@ -472,7 +471,7 @@ public class ForestQueries {
 			}
 		}
 		StringBuilder sb = new StringBuilder();
-		while (q-- > 0) {
+		for(int i = 0; i < q; i++) {
 			int y1 = io.nextInt();
 			int x1 = io.nextInt();
 			int y2 = io.nextInt();
--- a/content/3_Silver/Prefix_Sums_2.mdx
+++ b/content/3_Silver/Prefix_Sums_2.mdx
@@ -456,8 +456,9 @@ public class ForestQueries {
 		int N = n + 1;
 		int[][] forest = new int[N][N];
 		for (int i = 1; i < N; i++) {
+			String row = io.nextLine();
 			for (int j = 1; j < N; j++) {
-				if (io.next().equals("*")) {
+				if (row.charAt(j - 1) == '*') {
 					forest[i][j] = 1;
 				}
 			}
@@ -509,6 +510,14 @@ public class ForestQueries {
 				while (st == null || !st.hasMoreTokens())
 					st = new StringTokenizer(r.readLine());
 				return st.nextToken();
+			} catch (Exception ignored) {
+			}
+			return null;
+		}
+
+		public String nextLine() {
+			try {
+				return r.readLine();
 			} catch (Exception ignored) {
 			}
 			return null;
--- a/content/3_Silver/Prefix_Sums_2.mdx
+++ b/content/3_Silver/Prefix_Sums_2.mdx
@@ -487,7 +487,7 @@ public class ForestQueries {
 		}
 
 		io.println(sb);
-		pw.close();
+		io.close();
 	}
 
 	static class Kattio extends PrintWriter {
--- a/content/3_Silver/Prefix_Sums_2.mdx
+++ b/content/3_Silver/Prefix_Sums_2.mdx
@@ -449,7 +449,7 @@ public class ForestQueries {
 	static Kattio io = new Kattio();
 
 	public static void main(String[] args) throws IOException {
-		PrintWriter pw = new PrintWriter(System.out);
+		
 
 		int n = io.nextInt();
 		int q = io.nextInt();
--- a/content/3_Silver/Prefix_Sums_2.mdx
+++ b/content/3_Silver/Prefix_Sums_2.mdx
@@ -442,101 +442,86 @@ int main() {
 <JavaSection>
 
 ```java
-import java.io.DataInputStream;
-import java.io.IOException;
-import java.io.PrintWriter;
- 
-public class ForestQueries3 {
-    public static void main(String[] args) throws IOException {
-        Reader r = new Reader();
-        PrintWriter pw = new PrintWriter(System.out);
- 
-        int n = r.nextInt();
-        int q = r.nextInt();
-        int N = n + 1;
-        int[][] forest = new int[N][N];
-        for (int i = 1; i < N; i++) {
-            for (int j = 1; j < N; j++) {
-                if (r.isTree()) {
-                    forest[i][j] = 1;
-                }
-            }
-        }
- 
-        for (int i = 1; i < N; i++) {
-            for (int j = 1; j < N; j++) {
-                forest[i][j] += forest[i][j - 1];
-            }
-        }
-        for (int j = 1; j < N; j++) {
-            for (int i = 1; i < N; i++) {
-                forest[i][j] += forest[i - 1][j];
-            }
-        }
-        StringBuilder sb = new StringBuilder();
-        while (q-- > 0) {
-            int y1 = r.nextInt();
-            int x1 = r.nextInt();
-            int y2 = r.nextInt();
-            int x2 = r.nextInt();
-            sb.append(forest[y2][x2] - forest[y1 - 1][x2] - forest[y2][x1 - 1] + forest[y1 - 1][x1 - 1]).append("\n");
-        }
- 
-        pw.println(sb);
-        pw.close();
-    }
- 
-    private static class Reader {
-        final private int BUFFER_SIZE = 1 << 24;
-        private final DataInputStream dis;
-        private final byte[] buffer;
-        private int bufferPointer, bytesRead;
- 
-        public Reader() {
-            dis = new DataInputStream(System.in);
-            buffer = new byte[BUFFER_SIZE];
-            bufferPointer = bytesRead = 0;
-        }
- 
-        private int nextInt() throws IOException {
-            int ret = 0;
-            byte c = read();
-            while (c <= ' ')
-                c = read();
-            boolean neg = (c == '-');
-            if (neg)
-                c = read();
-            do {
-                ret = ret * 10 + c - '0';
-            } while ((c = read()) >= '0' && c <= '9');
-            if (neg)
-                return -ret;
-            return ret;
-        }
- 
-        private boolean isTree() throws IOException {
-            byte c = read();
-            while (c == '\n' || c == '\r')
-                c = read();
-            return c == '*';
-        }
- 
-        private void fillBuffer() throws IOException {
-            bytesRead = dis.read(buffer, bufferPointer = 0, BUFFER_SIZE);
-            if (bytesRead == -1)
-                buffer[0] = -1;
-        }
- 
-        private byte read() throws IOException {
-            if (bufferPointer == bytesRead)
-                fillBuffer();
-            return buffer[bufferPointer++];
-        }
- 
-        public void close() throws IOException {
-            dis.close();
-        }
-    }
+import java.io.*;
+import java.util.StringTokenizer;
+
+public class ForestQueries {
+	static Kattio io = new Kattio();
+
+	public static void main(String[] args) throws IOException {
+		PrintWriter pw = new PrintWriter(System.out);
+
+		int n = io.nextInt();
+		int q = io.nextInt();
+		int N = n + 1;
+		int[][] forest = new int[N][N];
+		for (int i = 1; i < N; i++) {
+			for (int j = 1; j < N; j++) {
+				if (io.next().equals("*")) {
+					forest[i][j] = 1;
+				}
+			}
+		}
+
+		for (int i = 1; i < N; i++) {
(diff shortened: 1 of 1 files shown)


content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,108 @@ int main() {
...
+ 
+public class ForestQueries3 {
+    public static void main(String[] args) throws IOException {
+        Reader r = new Reader();
+        PrintWriter pw = new PrintWriter(System.out);
+ 
+        int n = r.nextInt();
+        int q = r.nextInt();
+        int N = n + 1;
+        int[][] forest = new int[N][N];
```
frodakcin, line 457: Is this well defined behavior (I found out today it's not in cpp, but IDK about java)

frodakcin (commented):
Otherwise, OK

content/3_Silver/Prefix_Sums_2.mdx
mizlan: Just wandered across this thread, and the answer is [yes, it's well defined in Java](https://stackoverflow.com/a/3426854/10232025) :)

frodakcin: ok, thanks

frodakcin (approved):
OK

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,108 @@ int main() {
...
+                forest[i][j] += forest[i - 1][j];
+            }
+        }
+        StringBuilder sb = new StringBuilder();
+        while (q-- > 0) {
+            int y1 = r.nextInt();
+            int x1 = r.nextInt();
+            int y2 = r.nextInt();
+            int x2 = r.nextInt();
+            sb.append(forest[y2][x2] - forest[y1 - 1][x2] - forest[y2][x1 - 1] + forest[y1 - 1][x1 - 1]).append("\n");
```
dongliu0426, line 482: over 80 characters?

dongliu0426: https://usaco.guide/general/contributing#code-conventions

dongliu0426: also use tabs instead of spaces to indent

andrewwangva:
@bqi343 do we wang to start using kattio for input in java?

bqi343:
preferably, but this seems fine

content/3_Silver/Prefix_Sums_2.mdx
JyVenom: My editor is set up such that it auto replaces tabs with spaces. If you want, I can disable this or just find and replace spaces with the tab literal.

JyVenom:
Ok, i made a version using kattio. I have a question though. I don't use the "nextDouble" method, so is it okay if I remove it?. Also, the catch parameter in the "next"  is set to "e" but the catch itself is empty. is it okay if i rename "e" to "ignored?" so like from "catch (Exception e) {" to "catch (Exception ignored) {."

JyVenom:
wait no, nevermind. i cant use kattio io yet. this version doesn't work.

HarryWangATX:
use the newest???

JyVenom:
Ok, it works now. Using kattio seems to be twice as slow as my first version (59ms vs 28ms). I used tabs for indentation and I split up the long line Andrew Wang mentioned.

JyVenom:
> use the newest???

I literally just copy pasted this from the usaco guide website. So, as far as I know, this is the newest one (on usaco guide).

HarryWangATX:
> Ok, it works now. Using kattio seems to be twice as slow as my first version (59ms vs 28ms). I used tabs for indentation and I split up the long line Andrew Wang mentioned. 

i think it might be more standard because kattio was the IO on guides io section

HarryWangATX:
> Ok, i made a version using kattio. I have a question though. I don't use the "nextDouble" method, so is it okay if I remove it?. Also, the catch parameter in the "next"  is set to "e" but the catch itself is empty. is it okay if i rename "e" to "ignored?" so like from "catch (Exception e) {" to "catch (Exception ignored) {." 

its fine, just leave as is is good i think

HarryWangATX:
ur kattio seems different from the one im seeing on guide???

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,102 @@ int main() {
...
+			int x1 = io.nextInt();
+			int y2 = io.nextInt();
+			int x2 = io.nextInt();
+			sb.append(forest[y2][x2] -
+					forest[y1 - 1][x2] -
+					forest[y2][x1 - 1] +
+					forest[y1 - 1][x1 - 1]).append("\n");
+		}
+
+		pw.println(sb);
```
HarryWangATX, line 489:
```suggestion
		io.println(sb);
```

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,102 @@ int main() {
...
+		pw.close();
```
HarryWangATX, line 490:
```suggestion
		io.close();
```

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,102 @@ int main() {
...
+
+```java
+import java.io.*;
+import java.util.StringTokenizer;
+
+public class ForestQueries {
+	static Kattio io = new Kattio();
+
+	public static void main(String[] args) throws IOException {
+		PrintWriter pw = new PrintWriter(System.out);
```
HarryWangATX, line 452:
```suggestion
		
```

HarryWangATX (changes requested):

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,102 @@ int main() {
...
+
+public class ForestQueries {
+	static Kattio io = new Kattio();
+
+	public static void main(String[] args) throws IOException {
+		
+
+		int n = io.nextInt();
+		int q = io.nextInt();
+		int N = n + 1;
```
HarryWangATX, line 456:
```suggestion
		int N = io.nextInt()+1;
		int q = io.nextInt();
```

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,102 @@ int main() {
...
+```java
+import java.io.*;
+import java.util.StringTokenizer;
+
+public class ForestQueries {
+	static Kattio io = new Kattio();
+
+	public static void main(String[] args) throws IOException {
+		
+
```
HarryWangATX, line 453:
```suggestion
```

HarryWangATX (changes requested):

content/3_Silver/Prefix_Sums_2.mdx
JyVenom: add a space before and after the "+"
also, we can just do <=n instead of <N. i think that may make more sense.

HarryWangATX: i think its fine, changing to to <= n is the same as < N anyways.

bqi343 (changes requested):
see comment above

mizlan:
We were having a pretty lengthy discussion about using boilerplate in USACO server, and our conclusion was generally to avoid templating and boilerplate as that will inevitably get in the way of readability. Mind I ask why including Kattio would make the code more readable?

The question of whether to use fast IO boilerplate is not related to the algorithmic solution of a problem. Kattio should be presented in the section in the guide on tips for Java, not implanted into every editorial, detracting from the focus of the solution. I believe allowing readers to apply Kattio on their own as they see fit is more beneficial than including it in the editorial.

skittles1412:
IMO an IO boilerplate for Java (basically a `Scanner` class but with `BufferedReader` and `StringTokenizer`) is fine. `in.nextInt()` is pretty readable and self explanatory, and this itself is similar to the syntax `Scanner` uses, which most people already are familiar with. However I would say that perhaps titling it something like `FastScanner` and not extending `PrintWriter` (which I find to be an abuse of OOP) would be better.

bqi343:
> avoid templating and boilerplate as that will inevitably get in the way of readability. 

I don't see how using the same (relatively short) method of I/O would get in the way of readability? I think it beats having to manually initialize a `StringTokenizer` every time you want to read in a line.

> IMO an IO boilerplate for Java (basically a Scanner class but with BufferedReader and StringTokenizer) is fine. in.nextInt() is pretty readable and self explanatory, and this itself is similar to the syntax Scanner uses, which most people already are familiar with. 

I agree.

> Kattio should be presented in the section in the guide on tips for Java, not implanted into every editorial, detracting from the focus of the solution

you don't have to read it every time ... (plus @thecodingwizard suggested adding some method of collapsing it?)

> which I find to be an abuse of OOP

I don't particularly care about this :P

mizlan:
Is `Scanner` unacceptable for an editorial?

content/3_Silver/Prefix_Sums_2.mdx
bqi343: can we have <= n?

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,87 @@ int main() {
...
+<JavaSection>
+
+```java
+import java.io.*;
+import java.util.StringTokenizer;
+
+public class ForestQueries {
+	static Kattio io = new Kattio();
+	public static void main(String[] args) throws IOException {
+		int N = io.nextInt() + 1;
```
bqi343, line 451: I think `int n = io.nextInt()` made more sense.

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -439,6 +439,87 @@ int main() {
...
+		int q = io.nextInt();
+		int[][] forest = new int[N][N];
+		for (int i = 1; i < N; i++) {
+			String row = io.nextLine();
```
bqi343, line 455: doesn't compile anymore I think?

bqi343 (changes requested):

skittles1412:
For obvious reasons `Scanner` shouldn't be used, and I feel like a `FastScanner` class wouldn't affect readability at all. My main concern is `Kattio`. The biggest problem with it is that beginners are forced to either treat it as a black box or attempt to understand it, if they don't know OOP. In contrast just a `FastScanner` class which is much easier to understand (since it doesn't involve OOP), and I don't see any disadvantages of separating input class and output class.

bqi343:
> For obvious reasons `Scanner` shouldn't be used, and I feel like a `FastScanner` class wouldn't affect readability at all. My main concern is `Kattio`. The biggest problem with it is that beginners are forced to either treat it as a black box or attempt to understand it, if they don't know OOP. In contrast just a `FastScanner` class which is much easier to understand (since it doesn't involve OOP), and I don't see any disadvantages of separating input class and output class.

I agree that OOP would be harder to understand for a complete beginner, but I don't think that's sufficient reason to avoid it? Darren's book assumes knowledge of APCS, not sure if this should apply to the guide as well.

I agree that separating the two wouldn't affect readability, but I think it's more convenient to have all I/O managed through the same class. @HarryWangATX @AndrewWangva thoughts?

skittles1412:
What you're doing is the equivalent of merging cin and cout and then doing `cio >> ` `cio << `. This is personal preference, but you also have to consider contestants typing it in contest. A simple input class is much easier to write during the contest than a confusing subclass. Also I don't see how managing IO through the same/different class(es) change(s) much when you are writing code. `io.nextInt()` vs `in.nextInt()` and `io.println()` vs `pw.println()` are pretty similar. If you want a method for initializing io classes make the classes global and then write your corresponding messages.

mizlan:
@skittles1412 You should probably open a separate issue as this discussion is largely concerning something with greater scope that the PR at hand.

HarryWangATX:
> > For obvious reasons `Scanner` shouldn't be used, and I feel like a `FastScanner` class wouldn't affect readability at all. My main concern is `Kattio`. The biggest problem with it is that beginners are forced to either treat it as a black box or attempt to understand it, if they don't know OOP. In contrast just a `FastScanner` class which is much easier to understand (since it doesn't involve OOP), and I don't see any disadvantages of separating input class and output class.
> 
> 
> 
> I agree that OOP would be harder to understand for a complete beginner, but I don't think that's sufficient reason to avoid it? Darren's book assumes knowledge of APCS, not sure if this should apply to the guide as well.
> 
> 
> 
> I agree that separating the two wouldn't affect readability, but I think it's more convenient to have all I/O managed through the same class. @HarryWangATX @AndrewWangva thoughts?

i wasnt a user of Kattio at first, but i found it to be way easier than common input plus output seperated

HarryWangATX:
also Kattio is simply extending another class, googling extend in java is sufficient

frodakcin:
tbh I quite liked the use of fast IO. I actually don't know anywhere guide features it

bqi343:
> tbh I quite liked the use of fast IO. I actually don't know anywhere guide features it

the fast i/o module contains a similar input template

frodakcin:
Oh ok.

bqi343:
Closing as requested changes have not been made. Make a new PR when ready.

mizlan:
That's not a valid reason to close a PR.

Issue #803, opened 2021-02-18
Add editorial code requirements


skittles1412:
This applies to USACO editorials as well, really all editorials.

The point of editorial code is to provide a reference for an implementation of the ideas mentioned in the editorial, not for beginners to have to ask what it's doing on a discord server, and not for readers to have to split screen so they can view the macros on one side and the code on the other. Editorial code should be **as readable as possible**, which means at least no macros, templating to a minimum, and preferably comments. IMO **hard requirements should be set to improve the quality of editorials**. Editorial code readability shouldn't be compromised but rather be a required part of any editorial. This means that **you should spend the 10 extra minutes improving the readability of your code**, rather than give that responsibility to the hundreds of readers which will spend a lot more wasted time trying to understand your editorial code that should've been clear in the first place.

dolphingarlic:
What requirements do you propose?

skittles1412:
No 200 line templates. Macros should be avoided; maybe `using ll = long long;` but other stuff like loop macros and the shortening of 50 different combination of primitives, vectors, and pairs should be removed. Don't use separate methods for initializing IO; just use three lines at the beginning of main. Really, just use this as your template and don't add other stuff.
```cpp
#include <bits/stdc++.h>

using namespace std;

int main() {
    cin.tie(0) -> sync_with_stdio(0);
    //maybe freopen here
}
```

Try your best to make your code readable. Use meaningful variable names and function names (meaning that they may get long) and try your best to add comments to any code you think readers might find confusing. Comment what a variable/loop/function is doing and add implementation notes.

Commenting and readability is obviously subjective, but I think a hard requirement should be little to zero macros and no templates.

dolphingarlic:
I agree with you about the templates (after all, they're against USACO rules now :P). However, I don't think that comments are necessary unless the problem a) involves casework or b) uses "complicated" maths (since the written solution above the code almost always already has the variable definitions, I think)

What do you guys think about requirements for spacing in code? (i.e. `for(int i=0i<n;i++)` vs `for (int i = 0; i < n; i++)`) Are they also necessary?

skittles1412:
Comments are subjective. I think just add it for any code that you think readers might find confusing or use it as implementation notes that aren't mentioned in the editorial. I think I'll open a separate issue about formatting.

bqi343:
I agree that a lot of code needs to be replaced.

> No 200 line templates. 

:eyes:

frodakcin:
[...]

PR #805, opened 2021-02-18
add note about kattio, upd contributing

bqi343:
editorial code reqs are still TBD? #803

### DIFF
--- a/content/1_General/Contributing.mdx
+++ b/content/1_General/Contributing.mdx
@@ -183,7 +183,6 @@ Please read the following resources before contributing code (although you do no
 Additional Notes:
 
 - Use _tabs_ to indent.
-- [Macros](/general/generic?lang=cpp#macros) and "templates" longer than a few lines should be avoided.
 - Lines should not exceed 80 characters.
 - Be *consistent* about formatting your code, unlike the following snippet:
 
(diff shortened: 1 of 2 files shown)


bqi343:
Should there be a distinction before "referring" to `Kattio` and copy pasting it?

thecodingwizard:
No :P

(I really don't think the intent behind banning templates was to stop people from using fast input/output for Java...)

content/1_General/Contributing.mdx
```diff
@@ -195,4 +195,24 @@ Additional Notes:
...
+
+`Scanner` is [significantly slower](/general/fast-io?lang=java#fast-input) than `BufferedReader` and should be avoided.
+
+To improve readability, Java solutions using `BufferedReader` should use a wrapper class around input (preferably [`Kattio`](/general/io?lang=java#io-template) for consistency, or use your own variant if you really want). 
+
+</Info>
+
+</JavaSection>
+
+<PySection />
```
thecodingwizard, line 216: Maybe mention pep-8 for python?

bqi343: you can, I haven't actually looked at it

thecodingwizard: actually, on second thought, let's just not mention it in case there's some rule that doesn't make sense for competitive programming. We don't get many (any) python PR's anyway :P

thecodingwizard:
Should we specify that the following template is recommended for C++? (ie. macros other than the ones below should be avoided for Bronze - Gold)

```cpp
#include <bits/stdc++.h>
using namespace std;

using ll = long long;
using vi = vector<int>;
using pi = pair<int,int>;

#define sz(x) (int)(x).size()
#define pb push_back
#define all(x) begin(x), end(x)
#define f first
#define s second

int main() {
    cin.tie(0)->sync_with_stdio(0); 
    // freopen("name.in", "r", stdin); 
    // freopen("name.out", "w", stdout);

}
```

content/1_General/Contributing.mdx
```diff
@@ -195,4 +194,54 @@ Additional Notes:
...
+
+int main() {
+    cin.tie(0)->sync_with_stdio(0); 
+    // freopen("name.in", "r", stdin); 
+    // freopen("name.out", "w", stdout);
+
+}
+```
+
+You don't have to use the template, but avoid using additional macros beyond the ones listed above.
```
thecodingwizard, line 229: @bqi343 is this fine?

bqi343: https://github.com/cpinitiative/usaco-guide/issues/806#issuecomment-782051462 maybe?

bqi343:
seems fine to me

PR #815, opened 2021-02-20
Replaced template solutions for five problems

dolphingarlic:


bqi343 (changes requested):

solutions/IOI 04-empodia.mdx
```diff
@@ -2,11 +2,13 @@
 id: ioi-04-empodia
 source: IOI 2004
 title: Empodia
-author: Benjamin Qi
+author: Benjamin Qi, Andi Qu
 ---
 
 [Official Editorial](https://ioinformatics.org/page/ioi-2004/30)
 
+**Time Complexity:** $\mathcal O(N)$.
```
bqi343, line 10: your implementation isn't O(N)? Also, no period

bqi343 (changes requested):

solutions/IOI 04-empodia.mdx
dolphingarlic: Oops should I change the map into an unordered_map? If not, should I just use @tmwilliamlin168's implementation? (https://github.com/tmwilliamlin168/CompetitiveProgramming/blob/master/IOI/04-Empodia.cpp)

bqi343: wasn't my solution ok?

dolphingarlic: The only issue with your solution was that it used templates, but since it's an IOI problem, that's probably ok to have. I've reverted back to your code (with the new CodeSnip feature)

dolphingarlic:
I'll actually just close this PR for now because we have the new CodeSnip feature now

PR #817, opened 2021-02-20
capitalize each part of tag, remove some duplicates

bqi343:


bqi343:
Would be good if there was some sort of hierarchy to tags. 

Also, how specific do we want to be regarding tags (ex. do we want to tag a problem involving a lazy segment tree as lazy segment tree, segment tree, or just 1DRQ)?

frodakcin (approved):
I think `1DRQ` is preferable, but I think there should be a way to differentiate between log and sqrt based 1DRQ

maybe we consider sqrt based range query to be "sqrt decomp" and not a "range query"

PR #793, opened 2021-02-17
General - Silver Content Videos

jessechoe10:

PR #835, opened 2021-02-21
Videos + Add Sorting Problem + Change Difficulty

jessechoe10:


thecodingwizard (changes requested):
As per the discussion on #793, can we not use Cararra's videos? I don't think the comedy style fits well with the USACO Guide. Instead [williamfiset](https://www.youtube.com/user/purpongie) (and other channels) have some videos that better match the tone of the Guide

content/3_Silver/Sorting_Custom.mdx
```diff
@@ -68,21 +68,22 @@ export const problems = {
...
 			''
 		),
+		new Problem('Silver', 'Moo Particle', '1040', 'Normal', false, ['Sorting, Graphs'], ''),
+		new Problem('Silver', 'Triangles', '1015', 'Hard', false, ['Sorting'], ''),
 		new Problem(
 			'Silver',
 			'Meetings',
 			'967',
-			'Very Hard',
+			'Hard',
```
Farrius, line 86: From my point of view Meetings is very hard for silver.

PR #836, opened 2021-02-21
Update LanguageSection.tsx

### LINKED ISSUE #834: Show C++ code if Java/Python code isn't available?
From feedback on Discord:
> i feel like you should be able to choose multiple languages instead of just 1
> because if i prefer to see sols in java, but a solution is not available in java, i have to go and change my language to c++, then afterwards change in back

This might be a problem if people don't expect the code not to be in Java/Python though (e.g. if they want to copy and paste), so how do you guys think we should approach this?

bqi343:
Closes #834 

Displays another lang in addition to the incomplete message.

### DIFF
--- a/src/components/markdown/LanguageSection.tsx
+++ b/src/components/markdown/LanguageSection.tsx
@@ -4,17 +4,20 @@ import { useContext } from 'react';
 import { LANGUAGE_LABELS } from '../../context/UserDataContext/properties/userLang';
 
 export const LanguageSection = props => {
-  const { lang } = useContext(UserDataContext);
+  const { lang: userLang } = useContext(UserDataContext);
 
-  let sections = {};
+  const sections = {};
   React.Children.map(props.children, child => {
-    let type = child.props.mdxType;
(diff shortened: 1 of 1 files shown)


src/components/markdown/LanguageSection.tsx
```diff
@@ -47,26 +57,41 @@ export const LanguageSection = props => {
...
           </div>
         </div>
       </div>
     );
+    return (
+      <>
+        {notAvailable}
+        {fallbackLang && (
+          <>
+            <br />
```
thecodingwizard, line 87: we probably don't need the `<br>`; otherwise the spacing looks kind of weird

thecodingwizard (approved):

bqi343:
without the `<br />` there's no space between the incomplete message and the code in the solution spoiler here: https://usaco.guide/silver/binary-search?lang=cpp#example---maximum-median

thecodingwizard:
is this better?

bqi343:
Looks good.

PR #838, opened 2021-02-22
CodeBlock: Add Code Snips (excuse the spaghetti)

frodakcin:
A few problems I can fix if they need to be fixed:
- The "replacement text" (aka stuff in curly braces that is placed on the snipped line) is not formatted.
- Directly used style instead of a class

A few problems someone better than me needs to fix if they cause problems:
- If the majority of the code is snipped off, then the collapse detection fails.
- Did not use modern react stuff (I had to resort to `class ... extends React.component`)
- The part that detects BeginCodeSnip maybe hinders runtime

Use:
Put "`BeginCodeSnip{any random replacement text}`" somewhere in the code block and "`EndCodeSnip`" somewhere else. Everything on these 2 lines are deleted, and everything between them are code snips that are automatically minimized. There's buttons that can maximize it. I found the liveedit tool to be helpful for testing this.

^^ BTW if there is a better UI or format for the user to specify where to snip code, then we can probably change it without too much difficulty.

<details>
<summary> example </summary>

paste

````
```cpp
//BeginCodeSnip{}
#include <bits/stdc++.h>
using namespace std;
 
using ll = long long;
using db = long double; // or double, if TL is tight
using str = string; // yay python!

using pi = pair<int,int>;
using pl = pair<ll,ll>;
using pd = pair<db,db>;

using vi = vector<int>;
using vb = vector<bool>;
using vl = vector<ll>;
using vd = vector<db>; 
using vs = vector<str>;
using vpi = vector<pi>;
using vpl = vector<pl>; 
using vpd = vector<pd>;

#define tcT template<class T
#define tcTU tcT, class U
// ^ lol this makes everything look weird but I'll try it
tcT> using V = vector<T>; 
tcT, size_t SZ> using AR = array<T,SZ>; 
tcT> using PR = pair<T,T>;

// pairs
#define mp make_pair
#define f first
#define s second

// vectors
// oops size(x), rbegin(x), rend(x) need C++17
#define sz(x) int((x).size())
#define bg(x) begin(x)
#define all(x) bg(x), end(x)
#define rall(x) x.rbegin(), x.rend() 
#define sor(x) sort(all(x)) 
#define rsz resize
#define ins insert 
#define ft front()
#define bk back()
#define pb push_back
#define eb emplace_back 
#define pf push_front
#define rtn return

#define lb lower_bound
#define ub upper_bound 
tcT> int lwb(V<T>& a, const T& b) { return int(lb(all(a),b)-bg(a)); }

// loops
#define FOR(i,a,b) for (int i = (a); i < (b); ++i)
#define F0R(i,a) FOR(i,0,a)
#define ROF(i,a,b) for (int i = (b)-1; i >= (a); --i)
#define R0F(i,a) ROF(i,0,a)
#define rep(a) F0R(_,a)
#define each(a,x) for (auto& a: x)

const int MOD = 1e9+7;
const int MX = 2e5+5;
const ll INF = 1e18; // not too close to LLONG_MAX
const db PI = acos((db)-1);
const int dx[4] = {1,0,-1,0}, dy[4] = {0,1,0,-1}; // for every grid problem!!
mt19937 rng((uint32_t)chrono::steady_clock::now().time_since_epoch().count()); 
template<class T> using pqg = priority_queue<T,vector<T>,greater<T>>;

// bitwise ops
// also see https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
constexpr int pct(int x) { return __builtin_popcount(x); } // # of bits set
constexpr int bits(int x) { // assert(x >= 0); // make C++11 compatible until USACO updates ...
	return x == 0 ? 0 : 31-__builtin_clz(x); } // floor(log2(x)) 
constexpr int p2(int x) { return 1<<x; }
constexpr int msk2(int x) { return p2(x)-1; }

ll cdiv(ll a, ll b) { return a/b+((a^b)>0&&a%b); } // divide a by b rounded up
ll fdiv(ll a, ll b) { return a/b-((a^b)<0&&a%b); } // divide a by b rounded down

tcT> bool ckmin(T& a, const T& b) {
	return b < a ? a = b, 1 : 0; } // set a = min(a,b)
tcT> bool ckmax(T& a, const T& b) {
	return a < b ? a = b, 1 : 0; }

tcTU> T fstTrue(T lo, T hi, U f) {
	hi ++; assert(lo <= hi); // assuming f is increasing
	while (lo < hi) { // find first index such that f is true 
		T mid = lo+(hi-lo)/2;
		f(mid) ? hi = mid : lo = mid+1; 
	} 
	return lo;
}
tcTU> T lstTrue(T lo, T hi, U f) {
	lo --; assert(lo <= hi); // assuming f is decreasing
	while (lo < hi) { // find first index such that f is true 
		T mid = lo+(hi-lo+1)/2;
		f(mid) ? lo = mid : hi = mid-1;
	} 
	return lo;
}
tcT> void remDup(vector<T>& v) { // sort and remove duplicates
	sort(all(v)); v.erase(unique(all(v)),end(v)); }
tcTU> void erase(T& t, const U& u) { // don't erase
	auto it = t.find(u); assert(it != end(t));
	t.erase(it); } // element that doesn't exist from (multi)set


void DBG() { cerr << "]" << endl; }
template<class H, class... T> void DBG(H h, T... t) {
	cerr << h; if (sizeof...(t)) cerr << ", ";
	DBG(t...); }
#ifdef LOCAL // compile with -DLOCAL
#define dbg(...) cerr << "LINE(" << __LINE__ << ") -> [" << #__VA_ARGS__ << "]: [", DBG(__VA_ARGS__)
#else
#define dbg(...) 0
#endif
//EndCodeSnip

int N, Q, a[MX];
ll p[MX];

int main()
{
	scanf("%d%d", &N, &Q);
	for(int i=0;i<N;++i)
		scanf("%d", a+i);
	for(int i=0;i<N;++i)
		p[i+1]=p[i]+a[i];
	for(int i=0;i<Q;++i)
	{
		int l,r;
		scanf("%d%d", &l, &r);
		printf("%lld\n", p[r]-p[l]);
	}
	return 0;
}
```
````

into liveedit

</details>

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -75,7 +75,10 @@ export const plugins = [
         // },
       ],
       remarkPlugins: [require(`remark-external-links`), require('remark-math')],
-      rehypePlugins: [require('./src/mdx-plugins/rehype-math.js')],
+      rehypePlugins: [
+        require('./src/mdx-plugins/rehype-math.js'),
+        require('./src/mdx-plugins/rehype-snippets.js'),
+      ],
       plugins: [
         {
           resolve: `gatsby-remark-autolink-headers`,
--- a/src/components/DynamicMarkdownRenderer.tsx
+++ b/src/components/DynamicMarkdownRenderer.tsx
@@ -12,6 +12,7 @@ import grayMatter from 'gray-matter';
 import { components } from './markdown/MDXProvider';
 import { Problem } from '../models/problem';
 import customRehypeKatex from '../mdx-plugins/rehype-math.js';
+import rehypeSnippets from '../mdx-plugins/rehype-snippets.js';
 
 class ErrorBoundary extends React.Component {
   state: {
@@ -75,7 +76,7 @@ export const _frontmatter = ${JSON.stringify(data)}`;
         const jsx = (
           await mdx(content, {
             remarkPlugins: [remarkExternalLinks, remarkMath],
-            rehypePlugins: [customRehypeKatex],
+            rehypePlugins: [customRehypeKatex, rehypeSnippets],
             skipExport: true,
           })
         ).trim();
--- a/src/components/markdown/CodeBlock/CodeBlock.tsx
+++ b/src/components/markdown/CodeBlock/CodeBlock.tsx
@@ -27,7 +27,14 @@ const LineSnip = styled.span`
   white-space: nowrap;
   user-select: none;
   width: 1em;
-  padding-right: 1em;
+  padding-right: 0.5em;
+`;
+
+const CodeSnipButtonIcon = styled.svg`
+  color: rgba(156, 220, 254, 0.3);
+  &:hover {
+    color: rgba(156, 220, 254, 0.7);
+  }
 `;
 
 const LineContent = styled.span`
@@ -35,9 +42,9 @@ const LineContent = styled.span`
 `;
 
 const CodeSnippetLineContent = styled(LineContent)`
-  color: #00bb00; // any better color?
+  color: #00bb00a0; // any better color?
   &:hover > span {
-    text-decoration: underline;
+    color: #00bb00c0;
     cursor: pointer;
   }
 `;
@@ -52,11 +59,10 @@ const CodeSnipButton = ({
   onShowSnipChange: (snipID: number, showSnip: boolean) => void;
 }) => {
   return (
-    <svg
+    <CodeSnipButtonIcon
       fill="none"
       viewBox="0 0 24 24"
       stroke="currentColor"
-      opacity="0.3"
       className={
         'transform transition translate-y-0.5 h-4 cursor-pointer' +
         (!showSnip ? ' -rotate-90' : '')
@@ -69,7 +75,7 @@ const CodeSnipButton = ({
         strokeWidth={3}
         d="M17 10l-5 5-5-5"
       />
-    </svg>
+    </CodeSnipButtonIcon>
   );
 };
 
--- a/src/components/markdown/CodeBlock/CodeBlock.tsx
+++ b/src/components/markdown/CodeBlock/CodeBlock.tsx
@@ -34,42 +34,55 @@ const LineContent = styled.span`
   display: table-cell;
 `;
 
-class CodeSnipButton extends React.Component {
-  constructor(props) {
-    super(props);
-    this.clickEvent = this.clickEvent.bind(this);
+const CodeSnippetLineContent = styled(LineContent)`
+  color: #00bb00; // any better color?
+  &:hover {
+    text-decoration: underline;
+    cursor: pointer;
   }
+`;
 
-  clickEvent() {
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by thecodingwizard** [comment]
nice! I'm down to merge as is; possible remaining steps include:

- changing the color to something better
- automatically mapping certain labels (ex. "Benq's Template") to some code so the .mdx file isn't so long
- cleaning up the code
- (maybe) consider converting into functional component

But I think it's fine to merge as is too.

**Review by thecodingwizard — APPROVED** [review]

**Comment by frodakcin** [comment]
> automatically mapping certain labels

I think this in particular would be useful, specifically for benq's template and Kattio because I think these would be the primary use for code snips.

In terms of implementation, here are my ideas:

UI:
type
```
BeginCodeSnip{Benq Template}
content (which probably would be nothing)
EndCodeSnip
```
and `content` can be deleted for ben's template to be inserted, or we can prepend ben's template to `content`. 

Backend:

1. Ideally, I would like to modify the "children" variable but it turns out to be read-only (actually this was one of the reasons for the messy code in renderTokens -- I would have preferred to directly delete the Begin/End lines from `children` but instead I had to if-statement them out in renderTokens)
2. An alternative is to go in renderTokens and run a check whenever a "begin" line is encountered. If it's a special template, then display it. However, this would lead to even messier code. Also, I think you also have to pre-render the template through prism to know what to paste and that seems really excessive
3. Workaround to solution 1 -- when mdx is converted to html (idk when this happens, however) run a find-replace for all instances of the template keyword and replace it with the Begin/End syntax + pasted template. One benefit is that this solution is flexible to use any keyword (like `CodeSnip{Benq Template}` instead of having a contentless begin/end)

Idk how to implement any of the 3 above though...

thecodingwizard:
Something like this may theoretically be possible with Gatsby; during build-time, we can use regex to replace all instances of `BeginCodeSnip{Benq Template}` with the actual template in any code block (though I'll have to figure out how to do this). In this case, we wouldn't have to worry about performance since it's done during build-time.

thecodingwizard:
Added replacement functionality; `//CodeSnip{Benq Template}` is replaced with the full thing

frodakcin:
Awesome! I think `Kattio` would be very useful too

I think by default it should have 1 extra tab indent? So probably the regex should be more like `/\t*//CodeSnip{..}/` or even possibly `/.*CodeSnip{..}.*/g`

thecodingwizard:
oH yeah, we should intelligently determine indentation...

thecodingwizard:
okay this prepends whatever indentation there is before `//CodeSnip{}` to the replacement. I modified a bit of the regex though, @frodakcin you might want to take a look to make sure I didn't break anything...

thecodingwizard (approved):

frodakcin:
I think this is ready to merge.

PR #871, opened 2021-02-26
Detemplate for 2nd solution of Splitting the Field

tachithanhdanh:


solutions/USACO open16-splitting-the-field.mdx
```diff
@@ -92,238 +92,65 @@ From the analysis:
 > Note that it is also possible to dispense with the binary search trees entirely and just keep running mins and maxes in $y$.
```
tachithanhdanh, line 93: Maybe we should add some more explanation for this solution?

tachithanhdanh:
I have [another implementation](https://github.com/tachithanhdanh/CompetitiveProgramming/blob/main/USACO/Silver/Sorting%20with%20custom%20comparators/split_no_macros.cpp) for the 2nd solution which uses literally no macros, should I replace the current implementation with this?

solutions/USACO open16-splitting-the-field.mdx
SansPapyrus683: We probably should, but that's something for another PR(?).

solutions/USACO open16-splitting-the-field.mdx
```diff
@@ -92,238 +92,65 @@ From the analysis:
...
-	F0R(i,N-1) if (v[i].f != v[i+1].f)
-		ckmin(bes,(ll)(v[i].f-v[0].f)*(lef[i].s-lef[i].f)
-			+(ll)(v.bk.f-v[i+1].f)*(rig[i+1].s-rig[i+1].f));
-	ckmax(ans,area-bes);
+	for (int i = 1; i < N; ++i) lef[i] = comb(lef[i-1],v[i].s);
+	rig[N-1] = mp(v[N-1].s,v[N-1].s);
+	for (int i = N - 2; i >= 0; --i) rig[i] = comb(rig[i+1],v[i].s);
+	ll area = (ll)(v.back().f-v.front().f)*(lef.back().s-lef.back().f);
+	ll best = INF;
+	for (int i = 0; i < N - 1; ++i) if (v[i].f != v[i+1].f)
```
bqi343, line 138: you put spaces in subtraction here, but not above

bqi343 (changes requested):

solutions/USACO open16-splitting-the-field.mdx
tachithanhdanh: so shall I remove the spaces for consistency?

tachithanhdanh:
> I have [another implementation](https://github.com/tachithanhdanh/CompetitiveProgramming/blob/main/USACO/Silver/Sorting%20with%20custom%20comparators/split_no_macros.cpp) for the 2nd solution which uses literally no macros, should I replace the current implementation with this?

@bqi343 what do you think?

bqi343:
I love macros!!

tachithanhdanh:
> I love macros!!

But we are writing editorials which prioritize the code readability 😀

bqi343 (approved):

frodakcin (approved):
Changes are good.

Side note: why do the 2 solutions have different outputs on the below test case?

tachithanhdanh:
> Changes are good.
> 
> Side note: why do the 2 solutions have different outputs on the below test case?

I don't know. I'm just removing the template for Benq.

tachithanhdanh:
> Changes are good.
> 
> Side note: why do the 2 solutions have different outputs on the below test case?

I think the second implementation has some flaws because [my alternative implementation for the second approach](https://github.com/tachithanhdanh/CompetitiveProgramming/blob/main/USACO/Silver/Sorting%20with%20custom%20comparators/split_using_brute_force.cpp) works fine for that test case. 
@frodakcin

frodakcin:
wut... then you should fix it?

tachithanhdanh:
> wut... then you should fix it?

I have just tested again Benq's solution and it seems that there is a mistake.

solutions/USACO open16-splitting-the-field.mdx
```diff
@@ -92,238 +92,65 @@ From the analysis:
...
-	F0R(i,N-1) if (v[i].f != v[i+1].f)
-		ckmin(bes,(ll)(v[i].f-v[0].f)*(lef[i].s-lef[i].f)
-			+(ll)(v.bk.f-v[i+1].f)*(rig[i+1].s-rig[i+1].f));
-	ckmax(ans,area-bes);
+	for (int i = 1; i < N; ++i) lef[i] = comb(lef[i-1],v[i].s);
+	rig[N-1] = mp(v[N-1].s,v[N-1].s);
+	for (int i = N-2; i >= 0; --i) rig[i] = comb(rig[i+1],v[i].s);
+	ll area = (ll)(v.back().f-v.front().f)*(lef.back().s-lef.back().f);
+	ll best = INF;
+	for (int i = 0; i < N-1; ++i) if (v[i].f != v[i+1].f)
```
tachithanhdanh, line 138: Remove the ```if``` condition and everything will be fine

tachithanhdanh: when I remove this line, everything will be done.
@frodakcin

tachithanhdanh:
Both solutions are now correct, maybe someone should add some comments to explain the code from the 2nd implementation.

solutions/USACO open16-splitting-the-field.mdx
tachithanhdanh: > Remove the `if` condition and everything will be fine
Basically, that condition skips some of the cases which potentially are the answers.

bqi343:
not sure why you think my solution is the one that's wrong :eyes:

HarryWangATX (commented):
might as well get rid of some one liners

tachithanhdanh:
> not sure why you think my solution is the one that's wrong :eyes:
> 
> pretty sure two enclosures that share a corner are considered to overlap

The wrong one is the `if` condition in the last loop, it skips some correct answers, so it's totally correct to consider all cases rather than omit some of them.

bqi343:
Isn't the answer for that test case 6?

tachithanhdanh:
> Isn't the correct answer for that test case 6?

It's 8?

tachithanhdanh:
> Isn't the correct answer for that test case 6?

I did remember that the optimal answer for that test case is to divide into 2 enclosures whose areas are both equal to 4, so we can save 8.

solutions/USACO open16-splitting-the-field.mdx
```diff
@@ -92,262 +92,64 @@ From the analysis:
...
-17
-0 0
-0 1
-0 2
-1 0
-1 1
-1 2
-2 0
-2 1
-2 2
```
tachithanhdanh, line 344: If we take 9 first points and 8 last points to 2 enclosure, we will get the optimal answer @bqi343

bqi343:
the enclosures can't overlap

tachithanhdanh:
> the enclosures can't overlap

Oh so that's why we must include that `if` statement.

I think what we have to do now is to modify the first implementation to make it works the same as the second one.

solutions/USACO open16-splitting-the-field.mdx
tachithanhdanh: I thought it was wrong (?). I will revert it soon.

bqi343:
I think it's fine to just leave it as is (and add the correct answer for the test case :P)

tachithanhdanh:
> I think it's fine to just leave it as is (and add the correct answer for the test case :P)

You mean revert to that test case and keep that `if` statement removed?

bqi343:
I mean keep my solution the way it originally was.

PR #880, opened 2021-02-27
Add C++ Implementation for CF - Kayaking (Intro to Sorting)

tachithanhdanh:


solutions/CF Kayaking.mdx
```diff
@@ -0,0 +1,52 @@
+---
+id: kayaking
```
Farrius, line 2: I think is better to put kayaking-CF or something like that, since there could be more problems with the same name.

solutions/CF Kayaking.mdx
```diff
@@ -0,0 +1,52 @@
...
+			min_instability = min(min_instability, total_instability);
+		}
+	}
+	cout << min_instability;
+	return 0;
+}
+
+```
+
+
```
Farrius, line 49: The code looks good to me.

Farrius (changes requested):

solutions/CF Kayaking.mdx
```diff
@@ -0,0 +1,52 @@
...
+using namespace std;
+
+int main() {
+	ios_base::sync_with_stdio(0);
+	cin.tie(0);
+	int N;
+	cin >> N;
+	N *= 2;
+	vector<int> W(N);
+	for (int& x : W) cin >> x;
```
jessechoe10, line 29:
```cpp 
for(int i=0; i<N; i++){
cin >> W[i];
}
```

solutions/CF Kayaking.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## Implementation
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int main() {
+	ios_base::sync_with_stdio(0);
+	cin.tie(0);
```
jessechoe10, line 24: Optional: maybe try putting `cin.tie` together with `ios_base::sync_with_stdio(0);`...

jessechoe10 (changes requested):

solutions/CF Kayaking.mdx
Farrius: I think there is no need, everyone has personal preference writing code. Anyways, @darren-yao  decides.

Farrius (approved):

solutions/CF Kayaking.mdx
tachithanhdanh: I think it is fine so no need to change this.

andrewwangva (approved):

PR #886, opened 2021-03-02
CPP solution for revegetate

aldew5:


maggie-j-liu:
Isn't this the same as the [official solution](http://www.usaco.org/current/data/sol_revegetate_silver_feb19.html)? I haven't done this problem, so let me know if I'm missing something.

aldew5:
@maggie-j-liu oh wow you're right.  I didn't use the editorial, but my solution is very similar.  It's basically the same with different variable names.  Sorry about that

jessechoe10:
This solution is redundant to the official solution, so we're closing it @aldew5 .

PR #891, opened 2021-03-03
add warning about dp on bitmasks/ranges for gold

thecodingwizard:


bqi343:
What about LCA / HLD :eyes:

bqi343 (approved):

thecodingwizard:
I still hold out hope that HLD as part of gold was a mistake that won't be repeated D:

(But if we want to add HLD I don't particularly mind :P)

Issue #789, opened 2021-02-17
Code blocks should not collapse if they are the sole element inside a spoiler


jeffreymeng:
If a code block is the only thing inside a spoiler tag, then the code block should not be collapsed or collapsable. For example, see solution 2 of [this page](https://usaco.guide/gold/dp-trees?lang=java). When expanding the solution, there's no need to have to spend another click to expand the code block.

(how to solve)
Inside spoiler blocks, iterate through its children. If the only child is a code block, then pass a prop `expandable={false}` to the code block.
Inside the code block component, accept the prop mentioned above and render the entire code block, without any collapsing functionality.

MrinallU:
@jeffreymeng I'll do this.

MrinallU:
Closed by #1008

thecodingwizard:
I'm reopening until the PR is merged (and it should auto-close when it's merged anyway)

PR #895, opened 2021-03-04
Spoilers

MrinallU:


MrinallU:
Proof of concept/intial logic seems to work fine.

MrinallU:
I think this should work.

MrinallU:
@thecodingwizard Resolved Issue #789

thecodingwizard (changes requested):
Wait what we want is that if a code block is the only element of a spoiler, then the *code block* should not be expandable (it should just always be expanded). The *spoiler* should still function as normal, defaulting to collapsed and being able to expand/collapse when the user clicks on it. So you'll probably have to modify both the spoiler and the code block.

Also, there's two ways that there will only be a singular code block inside the spoiler:
- the child is a singular code block
- the child is a singular language section and each language section only has a singular child that's a code block

MrinallU:
Where can I find the code for the individual language section?

thecodingwizard:
https://github.com/cpinitiative/usaco-guide/blob/master/src/components/markdown/LanguageSection.tsx

MrinallU:
Actually I kinda meant the code for the code blocks inside the language
section so I can make the code block non expandable.

thecodingwizard:
https://github.com/cpinitiative/usaco-guide/blob/master/src/components/markdown/CodeBlock/CodeBlock.tsx

MrinallU:
> Wait what we want is that if a code block is the only element of a spoiler, then the _code block_ should not be expandable (it should just always be expanded). The _spoiler_ should still function as normal, defaulting to collapsed and being able to expand/collapse when the user clicks on it. So you'll probably have to modify both the spoiler and the code block.
> 
> Also, there's two ways that there will only be a singular code block inside the spoiler:
> 
> * the child is a singular code block
> * the child is a singular language section and each language section only has a singular child that's a code block

Will do.

dongliu0426:
closes #789

MrinallU:
@thecodingwizard So I'm able to pass the expandable prop from the Spoiler to the language section pretty fine but then the new prop does not seem to be transferred to the code block. As a result, I'm fairly sure that the children of the LanguageSection.tsx don't directly go to the CodeBlock.tsx component. As such,  do you know where the children and props of the LanguageSection.tsx file go to next?

thecodingwizard:
probably here: https://github.com/cpinitiative/usaco-guide/pull/895/files#diff-8996b64bdddc8b53b263fe009ad522143714ec06c36b45f17bf1b78d452d3076R111-R121

PR #896, opened 2021-03-04
add cpp solution (Grass Planting) (Bronze - Introduction to Graphs)

Ching367436:


maggie-j-liu:
Isn't this the same as the [official solution](http://www.usaco.org/current/data/sol_planting_silver_jan19.html)?

dongliu0426:
Thanks for your contribution, but there's already an official solution that uses the same logic + language, so we're closing this PR.

PR #899, opened 2021-03-05
C++ Solution to Triangles (Bronze, February 2020)

KartavyaSharma:


maggie-j-liu:
The [official solution](http://www.usaco.org/current/data/sol_triangles_bronze_feb20.html) has a C++ solution that is the same as yours...

Also, you have a typo here: `for(int i=0; i<0; i++) {`

KartavyaSharma:
@maggie-j-liu I didn't realise that, my bad. However, the USACO guide didn't seem to have an internal solution to this problem. If this solution is not required, should I just close this pull request?

maggie-j-liu:
Yep, you can close this; thanks!

Issue #890, opened 2021-03-03
Lex Min TopoSort


bqi343:
https://usaco.guide/gold/toposort?lang=cpp#bfs

Clarify that this code doesn't actually extract the lexicographically minimum topological sort ...

mizlan:
Git blame time! Maybe we would've been able to inspect [what happened there](https://github.com/cpinitiative/usaco-guide/blame/master/content/4_Gold/TopoSort.mdx#L375) if we had better commit messages...

bqi343:
Not sure what you mean. The module isn't incorrect per se (the code was never intended to extract it).

dolphingarlic:
Should we just move the lexographically minimum paragraph to an "optional" block to avoid the confusion?

PR #901, opened 2021-03-05
Clarify that BFS can be used to get the lexographically smallest toposort

dolphingarlic:
Also noted that the code doesn't do it

Closes https://github.com/cpinitiative/usaco-guide/issues/890

### DIFF
--- a/content/4_Gold/TopoSort.mdx
+++ b/content/4_Gold/TopoSort.mdx
@@ -374,7 +374,7 @@ This code assumes that there are no self-loops.
 
 ## BFS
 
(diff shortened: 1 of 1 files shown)


dongliu0426 (commented):
i think we should add code

dolphingarlic:
@bqi343 thoughts?

bqi343:
Just say that you can replace the queue with a priority queue.

dongliu0426 (approved):

bqi343 (approved):

PR #914, opened 2021-03-07
Add alternative solution for Family Tree - Bronze Open18

tachithanhdanh:


solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,125 @@
...
+				for (int i = 0; i < d2 - 2; ++i) cout << "great-";
+				cout << "aunt of " << c[2];
+			}
+		}
+		else cout << "COUSINS";
+	}
+
+	return 0;
+}
+```
```
Farrius, line 121: I liked the code and the comments. But I think the difficult of the problem, at least for me, is just know if is siblings or great/grand mother, and just simulate with O(n^4). In addition the use of BFS is not interesting and it can cause the sensation that you need to learn BFS to pass bronze.

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+</CPPSection>
```
Farrius, line 123: Wait for @darren-yao  he decides. But as I said I think it cause some confusion among bronze contestants although for me, the code and comments are well written.

solutions/USACO open18-familytree.mdx
frodakcin: I think he makes it pretty clearly that this is an alternative solution using concepts outside bronze curriculum.

Farrius: Sure, but personally I think it can be a little daunting if you  are bronze. But I’m probably being too subjective.

frodakcin: I see what you mean. Perhaps if we also had a "main solution" that was simple, the alternative would seem less daunting (in the sense that it's more clearly seen as an alternative)

Farrius: Yes, if we had that there would be no problem in adding this solution.

jessechoe10: Maybe we can write a "main" solution for this problem (similar to the official solution) and have this as an alternative solution that is recommended for people in Gold+...

tachithanhdanh: And maybe we can include this problem in Gold - BFS Module as a `very easy` problem

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,125 @@
...
+id: usaco-833
+source: USACO Bronze 2018 Open
+title: Family Tree
+author: Danh Ta Chi Thanh
+---
+
+[Official Analysis](http://www.usaco.org/current/data/sol_family_bronze_open18.html)
+
+## Alternative Solution - BFS
+
```
frodakcin, line 11:
```suggestion

## Solution

<IncompleteSection>

Add Intended Solution

</IncompleteSection>

```

we can do this as a cop-out

tachithanhdanh: I thought the main solution is in the **official analysis**?

frodakcin: Yes. Maybe just have alternate code or smth? I mean ofc we can just leave it blank and only have alternate.

frodakcin: I cannot english. What I mean:
1) We can either mention the main solution (and have some content under it, perhaps just an implementation), or
2) We can not mention the main solution and only have the alternate

tachithanhdanh: I think we can have something like this:

```

## Main solution (for Silver, Bronze)

paste the official analysis link here

## Alternative solution (for Gold+)

Insert my solution here

```

frodakcin:
```
## Solution
```
and
```
## Alternate Solution
```
are sufficient

Typically we separate main solution and editorial even if they are the same though... Maybe we can just say "check official editorial above" or smth

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,127 @@
...
+---
+
+## Solution
+
+[Official Analysis (C++)](http://www.usaco.org/current/data/sol_family_bronze_open18.html)
+
+## Alternative Solution - BFS
+
+This problem can also be solved by using BFS to find the common ancestors and the distance between the two cows and their nearest common ancestor.
+
```
jessechoe10, line 15: Provide more explanation to your approach...

jessechoe10: Since you're using BFS (a gold topic), I recommend you copy this problem on the BFS module for gold and make the difficulty `very easy`

jessechoe10 (approved):

solutions/USACO open18-familytree.mdx
thecodingwizard: Oops I don't think this is a good problem to put in gold BFS considering the intended solution isn't bfs :P

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,127 @@
...
+	for (int k = 1; k <= 2; ++k) {
+		queue<string> q;
+		q.push(c[k]);
+		// BFS on directed tree
+		// to find the ancestors of both cows and the distance to them
+		while (!q.empty()) {
+			string node = q.front();
+			q.pop();
+			for (auto& E : edge) if (E.f == node) {
```
frodakcin, line 76: This is potentially bad code practice... You should use something that points E.f to E.s instead of looping through all edges. Pretend the constraints are actually high

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,127 @@
...
+	dist[1].pb(mp(c[1],0));
+	dist[2].pb(mp(c[2],0));
+
+	vector<pair<string,string>> edge(N); //edge list
+
+	for (int i = 0; i < N; ++i) {
+		string u, v;
+		cin >> v >> u;
+		// If v is a mother of u, then there is an edge from node u to node v.
+		edge[i] = mp(u,v);
```
frodakcin, line 65: Ugh I hate this problem... typically you would map strings to integers and work with integers. It's not necessary here and makes it a lot more complicated, but it's important for more typical constraints.

frodakcin (changes requested):
Since this solution is already overkill, just go all out. Solve it as if it has max constraints.

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,127 @@
...
+source: USACO Bronze 2018 Open
+title: Family Tree
+author: Danh Ta Chi Thanh
+---
+
+## Solution
+
+[Official Analysis (C++)](http://www.usaco.org/current/data/sol_family_bronze_open18.html)
+
+## Alternative Solution - BFS
```
frodakcin, line 12: I forgot to mention.. what is runtime?

frodakcin (changes requested):
(in terms of N and $\sum{|s_i|}$)

solutions/USACO open18-familytree.mdx
tachithanhdanh: maybe 'map<string,string>`?

solutions/USACO open18-familytree.mdx
tachithanhdanh: It's O(N^2), but can be faster by using `map` to store the value.

solutions/USACO open18-familytree.mdx
frodakcin: Yes, I suppose that's fine.

solutions/USACO open18-familytree.mdx
frodakcin: Yea, I'd like to see it faster -- it feels odd for a bottleneck to be so easily avoidable...

solutions/USACO open18-familytree.mdx
frodakcin: Wouldn't it be simpler to match each string to an integer and just stick to arrays?

solutions/USACO open18-familytree.mdx
tachithanhdanh: The constraints are low, so I don't know if the constant factors from inserting and finding elements in a `std::map` would be an issue or not.

frodakcin: The point is that if we want a overkill solution, why have it only be partly overkill? The fact that it's kind of in the middle makes it un-understandable for beginners (which is intended) but also not as useful to those more familiar w the concept.

tachithanhdanh: I think the BFS solution should be just for those who want to use a different approach for this problem.

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,129 @@
...
+			}
+		}
+	}
+
+	string ancestor;
+	int d1 = 105, d2 = 105;
+	// Finding the common ancestor with shortest distance
+	for (auto& D1 : dist[1]) {
+		auto D2 = dist[2].find(D1.f);
+		if (D2 != dist[2].end() && (d1 > D1.s || d2 > D2->s)) {
```
frodakcin, line 91: This can be made linear (just by keeping distance to root and sliding them up, like a typical O(N) lca)

frodakcin (changes requested):

solutions/USACO open18-familytree.mdx
tachithanhdanh: How to make them linear?
I haven't known of the LCA Algorithm :(((

frodakcin: Ok, so here's the general idea:

Given two nodes $a$ and $b$ are distances $d_a$ and $d_b$ away from the root (great ... great grandfather of them all), we try to find their LCA. WLOG suppose $d_a>d_b$. Slide $a$ up (move to its parent) until $d_a$ is equal to $d_b$. Then keep sliding $a$ and $b$ up until they are equal. The first number for which they are equal is the lca.

ack bad explanation but i cannot think rn sry

frodakcin: You know what it's fine as is... it's N log N, and a way to find lca i've never seen before

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,129 @@
...
+		// If v is a mother of u, then there is an edge from node u to node v.
+		edge[u] = v;
+	}
+
+	for (int k = 1; k <= 2; ++k) {
+		queue<string> q;
+		q.push(c[k]);
+		// BFS on directed tree
+		// to find the ancestors of both cows and the distance to them
+		while (!q.empty()) {
```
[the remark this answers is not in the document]
frodakcin: I think you should just make it a for loop... this edges on both bad code practice and confusion (because it's completely unnecessary to use a queue)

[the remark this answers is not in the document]
tachithanhdanh: Yeah, I'll remove the queue.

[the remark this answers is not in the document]
tachithanhdanh: Wait but how am I supposed to do to make sure that I'm visiting the node in the right order as the queue?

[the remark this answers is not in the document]
tachithanhdanh: Okay I knew the way to change it.

solutions/USACO open18-familytree.mdx
```diff
@@ -0,0 +1,127 @@
...
+		// BFS on directed tree
+		// to find the ancestors of both cows and the distance to them
+		string node = c[k];
+		while (sz(node)) {
+			auto E = edge.find(node);
+			if (E != edge.end()) {
+				dist[k][E->s] = dist[k][node] + 1;
+				node = E->s;
+			}
+			else node = "";
```
frodakcin, line 80:
```suggestion
			else break;

```

is better practice I think

frodakcin (changes requested):

solutions/USACO open18-familytree.mdx
darren-yao: I think it's fine to put the BFS solution in bronze as long as it's clearly indicated that BFS is not necessary for bronze. However, I think putting a bronze question in a gold module is kind of ridiculous.

darren-yao (approved):

PR #927, opened 2021-03-08
Update Func_Graphs.mdx

### LINKED ISSUE #882: Todo: Remove Dance Mooves, Swapity Swapity Swap from Func Graph
https://usaco.guide/silver/func-graphs?lang=cpp

Material discussed in the module isn't relevant for these problems.

bqi343:
Closes #882

### DIFF
--- a/content/3_Silver/Func_Graphs.mdx
+++ b/content/3_Silver/Func_Graphs.mdx
@@ -27,8 +27,8 @@ export const problems = {
 			'https://codeforces.com/contest/1137/problem/D',
 			'Hard',
 			false,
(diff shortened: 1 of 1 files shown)


frodakcin (approved):
Any reason double quotes are preferred over `\'`? (I had changed it like that to be consistent with the rest of the single quotes)

frodakcin:
Actually why are we not always using double quotes?

bqi343:
Whatever is fine, I think the autoformatter defaults to single quotes. unless single quote is contained in between.

frodakcin:
I don't like autoformatter :(

I think it should default to double quote for all strings and only permit single quote for chars.

Issue #874, opened 2021-02-26
Update USACO Langs


bqi343:
Change mention of C++11 here:

https://github.com/cpinitiative/usaco-guide/blob/369f46701fca844483d6ce28aed27caa842afadf/content/1_General/Cpp_Command.mdx

Also

https://github.com/cpinitiative/usaco-guide/pull/409#issuecomment-786747792

thecodingwizard:
Closed with #928 (I think)

PR #928, opened 2021-03-08
F string revert

andrewwangva:
Issue #874
Not sure what else I need to change?

### DIFF
--- a/content/1_General/Cpp_Command.mdx
+++ b/content/1_General/Cpp_Command.mdx
@@ -259,7 +259,7 @@ g++ -std=c++17 -O2 name.cpp -o name -Wall
 ```
 
 - `-O2` tells `g++` to compile your code to run more quickly while increasing compilation time (see [here](https://www.rapidtables.com/code/linux/gcc/gcc-o.html)).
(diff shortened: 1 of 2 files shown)


content/1_General/Cpp_Command.mdx
```diff
@@ -259,7 +259,7 @@ g++ -std=c++17 -O2 name.cpp -o name -Wall
 ```
 
 - `-O2` tells `g++` to compile your code to run more quickly while increasing compilation time (see [here](https://www.rapidtables.com/code/linux/gcc/gcc-o.html)).
-- `-std=c++17` allows you to use features that were added to C++ in 2017. USACO currently uses `-std=c++0x` ([aka C++11](https://stackoverflow.com/questions/9538701/what-is-the-difference-between-c0x-and-c11)).
+- `-std=c++17` allows you to use features that were added to C++ in 2017. USACO has recently added C++17, which is up to date.
```
bqi343, line 262: > which is up to date.

not sure what this means, C++20 exists?

bqi343 (changes requested):

content/1_General/Cpp_Command.mdx
bqi343: maybe leave in that USACO recently upgraded to C++17 from C++11

PR #945, opened 2021-03-10
Added note in gold section

HarryWangATX:


bqi343 (commented):
ok after above is fixed

bqi343 (approved):

PR #965, opened 2021-03-14
Add segtree implementation

jeffarjeffar:
It's kinda long and a bit messy.
Please comment if there's something that needs to be fixed or cleaned up

Btw why is IOI: 2005 - Mountain in two modules?


bqi343:
Isn't there already a similar implementation in the sparse segtree module?

> Btw why is IOI: 2005 - Mountain in two modules?

I guess you can do it with or without implicit segment tree. (I rarely use implicit segment tree myself though.)

PR #823, opened 2021-02-20
Update Debugging_General.mdx

aksh1:

PR #980, opened 2021-03-16
Update Debugging_General.mdx

bqi343:
emphasize ll overflow. also comment out repetition example because as I stated in #823, I don't think it's particularly good

### DIFF
--- a/content/1_General/Debugging_General.mdx
+++ b/content/1_General/Debugging_General.mdx
@@ -28,48 +28,6 @@ This module is based on both of these resources. I've included the content that 
 
 - Don't repeat yourself!
 
-<!--
-<LanguageSection>
-
-<JavaSection>
-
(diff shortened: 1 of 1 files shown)


content/1_General/Debugging_General.mdx
```diff
@@ -6,24 +6,29 @@ description: 'General tips for identifying errors within your program.'
...
 
 ## Pre-Submit
 
- - Don't repeat yourself!
+- Don't repeat yourself!
 
+<!--
 <LanguageSection>
 
 <JavaSection>
```
thecodingwizard, line 34: might as well just delete the entire thing instead of commenting it out -- it can be found via the PR (https://github.com/cpinitiative/usaco-guide/pull/823/files) anyway

thecodingwizard (changes requested):

PR #986, opened 2021-03-17
Fix template link

dongliu0426:


thecodingwizard:
@bqi343 should we use the new one?

bqi343:
either should work. My current template is not C++11 compatible, previous one is

thecodingwizard (changes requested):
@dongliu0426 can we update to the new template then?

- update link in content documentation
- update snippet here: https://github.com/cpinitiative/usaco-guide/blob/master/src/mdx-plugins/rehype-snippets.js
- any other changes?

thecodingwizard:
seems fine I hope

PR #988, opened 2021-03-17
downgrade silver problem difficulties

bqi343:
should encourage ppl to solve these (people probably just skip over those marked "insane"?)


frodakcin:
I think Plat load balancing is very justifiably "Insane" for silver binary search module... as is Sprinklers for psum...

I don't know that atcoder problem though

HarryWangATX:
Yes, I agree with the actual difficulty but why put them there if no one wants to solve them anyway? Either remove from module or decrease the difficulty to encourage people to try is my opinion.

dongliu0426:
or maybe star the problem?

content/3_Silver/Prefix_Sums_2.mdx
```diff
@@ -90,7 +90,7 @@ export const problems = {
 			'Plat',
 			'Sprinklers',
 			'794',
-			'Insane',
+			'Very Hard',
 			false,
 			['Prefix Sums'],
```
jessechoe10, line 95: We should remove this problem altogether ngl...

jessechoe10 (changes requested):

jessechoe10:
tbh, i think we should just remove the section `**A More Complicated Example**` altogether. idt it's particularly useful for the silver division cuz there hasn't rly been a problem like this b4...

content/3_Silver/Binary_Search.mdx
```diff
@@ -82,7 +82,7 @@ export const problems = {
 			'Plat',
 			'Load Balancing',
```
jessechoe10, line 83: For this problem, we should provide a detemplified solution (the template is too long :stuck_out_tongue: ). Could you detemplify or should someone else do it?

jessechoe10 (changes requested):

bqi343:
> tbh, i think we should just remove the section **`A More Complicated Example`** altogether. idt it's particularly useful for the silver division cuz there hasn't rly been a problem like this b4...

maybe remove this part and add the first two problems to the end of the first prefix sums module?

> We should remove this problem altogether ngl...

I don't think sprinklers is harder than the other problems. Thought I guess it's not a great example of prefix sums (I just did it w/ two pointers) so it's probably fine to remove.

content/3_Silver/Binary_Search.mdx
bqi343: you can do it :)

jessechoe10:
> > tbh, i think we should just remove the section **`A More Complicated Example`** altogether. idt it's particularly useful for the silver division cuz there hasn't rly been a problem like this b4...
> 
> maybe remove this part and add the first two problems to the end of the first prefix sums module?
> 
> > We should remove this problem altogether ngl...
> 
> I don't think sprinklers is harder than the other problems. Though I guess it's not a great example of prefix sums (I just did it w/ two pointers only) so it's probably a good idea to remove.



> I don't think sprinklers is harder than the other problems. Though I guess it's not a great example of prefix sums (I just did it w/ two pointers only) so it's probably a good idea to remove.



> maybe remove this part and add the first two problems to the end of the first prefix sums module?

Yeah

jessechoe10 (approved):

jessechoe10:
@bqi343 I think we can merge. Maybe we should wait for other reviewers tho :thinking: .

content/3_Silver/Prefix_Sums.mdx
```diff
@@ -92,29 +101,37 @@ export const problems = {
...
 		new Problem(
-			'Silver',
-			'My Cow Ate My Homework',
-			'762',
-			'Easy',
+			'KS@Google KickStart',
+			'Candies (Test Set 1)',
+			'https://codingcompetitions.withgoogle.com/kickstart/round/000000000019ff43/0000000000337b4d',
+			'Hard',
 			false,
```
frodakcin, line 109: I think this trick/modification to psums is quite useful. If we remove from module should we star the problem? (or is there another starred problem that introduces this trick?)

frodakcin (changes requested):

content/3_Silver/Prefix_Sums.mdx
bqi343: wait, were you referring to "Candies" or "My Cow Ate my Homework"?

frodakcin (approved):

content/3_Silver/Prefix_Sums.mdx
frodakcin: Candies (or wherever the `a[i]*i` psum array explanation is)

PR #998, opened 2021-03-19
dp bitmasks

nwatx:
I wrote this as an elaboration for the topics presented, but it felt rather long, so I decided to make a blog and put it on there. Luckily all the mdx is cross compatible with usaco guide, so please do lmk if any sections should be merged. Not totally sure how we're going to do a pr the traditional way on this one though.

### DIFF
--- a/content/5_Plat/DP_Bitmasks.mdx
+++ b/content/5_Plat/DP_Bitmasks.mdx
@@ -232,7 +232,7 @@ where $S  \setminus \{i\}$ is the subset $S$ without city $i$.
 using namespace std;
 #define ll long long
 
-const int MX = (1 << 21);
+const int MX = (1 << 20);
 ll dp[MX][21];
 vector<int> adj[21];
 const ll MOD = (ll)1e9+7;
--- a/content/5_Plat/DP_Bitmasks.mdx
+++ b/content/5_Plat/DP_Bitmasks.mdx
@@ -1,7 +1,7 @@
 ---
 id: dp-bitmasks
 title: 'Dynamic Programming on Bitmasks'
-author: Michael Cao, Siyong Huang, Andrew Wang
+author: Michael Cao, Siyong Huang, Andrew Wang, Neo Wang
 prerequisites:
   - intro-bitwise
   - intro-dp
--- a/content/5_Plat/DP_Bitmasks.mdx
+++ b/content/5_Plat/DP_Bitmasks.mdx
@@ -232,7 +232,7 @@ where $S  \setminus \{i\}$ is the subset $S$ without city $i$.
 using namespace std;
 #define ll long long
 
-const int MX = (1 << 21) + 5;
+const int MX = (1 << 21);
 ll dp[MX][21];
 vector<int> adj[21];
 const ll MOD = (ll)1e9+7;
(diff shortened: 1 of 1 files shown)


jessechoe10 (approved):

thecodingwizard:
>  Not totally sure how we're going to do a pr the traditional way on this one though.

You can either leave it as is and just link to your blog, or just copy paste your blog MDX into the file. It should work unless you're using some fancy components we don't have :)

bqi343:
What's V in the complexity? 

Also, why do you use `\textrm` everywhere? I think $\mathcal{O}(n)$ looks a lot better than $\mathcal{O}(\textrm{n})$.

Screen Shot 2021-03-19 at 5 18 06 PM

Just use Arabic numerals?

Screen Shot 2021-03-19 at 5 20 18 PM

vertex? Missing periods in some places

Screen Shot 2021-03-19 at 5 21 01 PM

Don't like how this looks. I prefer how I styled bitmask DP [here.](http://www.usaco.org/current/data/sol_prob1_gold_jan21.html)

nwatx:
> 
> 
> What's V in the complexity?
> 
> Also, why do you use `\textrm` everywhere? I think $\mathcal{O}(n)$ looks a lot better than $\mathcal{O}(\textrm{n})$.
> 
> Screen Shot 2021-03-19 at 5 18 06 PM
> 
> Just use Arabic numerals?
> 
> Screen Shot 2021-03-19 at 5 20 18 PM
> 
> vertex? Missing periods in some places
> 
> Screen Shot 2021-03-19 at 5 21 01 PM
> 
> Don't like how this looks. I prefer how I styled bitmask DP [here.](http://www.usaco.org/current/data/sol_prob1_gold_jan21.html)

Ok changes (I'll say here but if it's more convenient I can just make the gh repo public/add u)

V => k
\textrm was not used everywhere, rather a styling sheet error. That has been fixed.
CN => arabic
vertex + periods
image
changed style

bqi343:
Do you mean m, not k? 

I think $\ll$ looks better than $<<$.

Why does your array have size `(1<<21) + 5`?

> Screen Shot 2021-03-19 at 7 57 34 PM

Typo?

> image

There should be space in between `]` and `if`. $x$ should not have the same text style as `if`.

nwatx:
> 
> 
> Do you mean m, not k?
> 
> I think $\ll$ looks better than $<<$.
> 
> Why does your array have size `(1<<21) + 5`? (referring to the `+5` specifically)
> 
> > Screen Shot 2021-03-19 at 7 57 34 PM
> 
> Typo?
> 
> > image
> 
> There should be space in between `]` and `if`. $x$ should not have the same text style as `if`.

changed k => m (oops)

ll looks cool so yeah (also less space)

no need for +5 => removed (although padding ig?)

added spacing

bqi343:
> Screen Shot 2021-03-19 at 8 24 39 PM

> Screen Shot 2021-03-19 at 8 26 04 PM

Actually, doesn't `1<<20` suffice?

bqi343 (approved):

thecodingwizard (approved):
oops the blog post is too long, I don't wanna read it D:

(all the other changes look fine though)

bqi343:
can mb merge the content of the blog later

PR #1009, opened 2021-03-21
Garland 2P solution

HarryWangATX:

PR #1015, opened 2021-03-22
Detemplated CF books and added a simpler sol

TanishT:
the current internal sol has the benq template and was kinda confusing for me. Should I have replaced his sol with mine or should I leave them both there?


bqi343:
This is not the same as the previous solution.

solutions/CF Books.mdx
```diff
@@ -14,223 +14,41 @@ When moving the left pointer one unit to the right, the right pointer should be
...
-		res += "}"; return res;
-	#else
-		bool fst = 1; str res = "";
-		for (const auto& x: v) {
-			if (!fst) res += " ";
-			fst = 0; res += ts(x);
+int main(){
+	int n, t;	
+	cin >> n >> t;
+	int arr[n];
```
bqi343, line 31: As noted [here](https://usaco.guide/bronze/intro-ds?lang=cpp), variable length arrays are not part of the C++ standard, so I think vector is preferred?

bqi343 (changes requested):

TanishT:
I added the previous solution back in and added my solution with a brief explanation at the bottom. Let me know if anything needs to be fixed.

solutions/CF Books.mdx
```diff
@@ -14,6 +14,18 @@ When moving the left pointer one unit to the right, the right pointer should be
...
 
+Explanation of Sample Cases:
+1. Valera can read the books with times 1, 2, 1.
+2. Valera can only read one of the books. 
+
+<LanguageSection>
+
+<CPPSection>
+
+# C++ Implementation
```
bqi343, line 25: > # C++ Implementation

remove this

solutions/CF Books.mdx
```diff
@@ -14,6 +14,18 @@ When moving the left pointer one unit to the right, the right pointer should be
...
+
+## Solution 1
```
bqi343, line 27: Only put code inside of LanguageSection.

bqi343 (changes requested):

solutions/CF Books.mdx
```diff
@@ -17,211 +23,21 @@ than or equal to $t$. This way, we can find for each possible left end the
...
-		}
-		res += "}"; return res;
-	#else
-		bool fst = 1; str res = "";
-		for (const auto& x: v) {
-			if (!fst) res += " ";
-			fst = 0; res += ts(x);
-		}
-		return res;
+Our goal is to find the length of the longest subarray that has a sum of greater than or equal to $t$. In the first sample case, the subarray that has [$1,2,1$] sums to $5$ and has a length of $3$. The subarray [$3,1,2$] that sums to $6$ also works. In the second sample case, any of the subarrays with length $1$ will work.
```
bqi343, line 28: > greater than or equal

less than or equal to?

> `[$1,2,1$]`

$[1,2,1]$ preferred

> The subarray that has [$1,2,1$] sums to $5$ and has a length of $3$.

1+2+1 is not 5?

> The subarray [$3,1,2$] that sums to $6$ also works.

I don't think so

bqi343: $left$

solutions/CF Books.mdx
```diff
@@ -17,211 +23,21 @@ than or equal to $t$. This way, we can find for each possible left end the
...
-	// throws exception when do smth illegal
-	// ex. try to read letter into int
-	if (sz(s)) setIn(s+".in"), setOut(s+".out"); // for USACO
-}
+
+<LanguageSection>
+
+<CPPSection>
+
+## Solution 1
```
bqi343, line 37: Please put the implementation directly below the corresponding explanation (as in https://github.com/cpinitiative/usaco-guide/pull/1009/files)

bqi343 (changes requested):

solutions/CF Books.mdx
bqi343: shouldn't this go above both solutions?

bqi343 (changes requested):

solutions/CF Books.mdx
bqi343: ^ not resolved

bqi343 (changes requested):
see unresolved comment above

solutions/CF Books.mdx
```diff
@@ -2,28 +2,42 @@
...
 
 ```cpp
 #include <bits/stdc++.h>
 using namespace std;
-
+
```
bqi343, line 37: also what's up w/ all these spaces :/

bqi343 (changes requested):

PR #1020, opened 2021-03-24
Expandable sidebar

MrinallU:


package.json
```diff
@@ -83,6 +84,7 @@
     "react-instantsearch-dom": "^6.9.0",
     "react-markdown": "^5.0.3",
     "react-notification-system": "^0.4.0",
+    "react-pro-sidebar": "^0.6.0",
```
thecodingwizard, line 87: I'd be cautious about using this library. It animates width, which is extremely expensive CPU wise, especially since we have a very large number of DOM elements. I would test performance especially on mobile devices (personally I don't think this library will work unless we override the width transitions or something).

MrinallU: That library actually didn't work so I will have to manually apply the CSS transitions.

MrinallU:
@thecodingwizard I'll close this for now and reopen with a different branch.

PR #1067, opened 2021-03-30
Added a new python solution for feb 21 bronze q3

cskitty:
Added a new solution in python for feb 21 bronze q3 using vector cross production

### DIFF
--- a/solutions/USACO feb21-clockwise-fence.mdx
+++ b/solutions/USACO feb21-clockwise-fence.mdx
@@ -13,32 +13,33 @@ The vector cross product can be used to determine if 2 vectors are rotating cloc
 
 <LanguageSection>
 <PySection>
+
 ## Implementation
 
 ```python
 def cross_product(p1, p2):
-    return p1[0] * p2[1] - p2[0] * p1[1]
+	return p1[0] * p2[1] - p2[0] * p1[1]
 
 for i in range (int(input())):
-    p0 = (0, 0)
-    sum = 0
-    direction = input()
-    for d in direction:
-        if d == 'N':
-            p1 = (p0[0], p0[1] + 1)
-        elif d == 'E':
-            p1 = (p0[0] + 1, p0[1])
-        elif d == 'S':
-            p1 = (p0[0], p0[1] - 1)
-        else:
-            p1 = (p0[0] - 1, p0[1])
-        sum += cross_product(p1, p0)
-        p0 = p1
+	p0 = (0, 0)
+	sum = 0
+	direction = input()
+	for d in direction:
+		if d == 'N':
+			p1 = (p0[0], p0[1] + 1)
+		elif d == 'E':
+			p1 = (p0[0] + 1, p0[1])
+		elif d == 'S':
+			p1 = (p0[0], p0[1] - 1)
+		else:
+			p1 = (p0[0] - 1, p0[1])
+		sum += cross_product(p1, p0)
+		p0 = p1
 
-    if sum > 0:
-        print("CW")
-    else:
-        print("CCW")
+	if sum > 0:
+		print("CW")
+	else:
+		print("CCW")
 ```
 
 </PySection>
--- a/solutions/USACO feb21-clockwise-fence.mdx
+++ b/solutions/USACO feb21-clockwise-fence.mdx
@@ -13,7 +13,7 @@ To find the area of a triangle with a vertex at (0, 0), a second vertex at (x1, 
 
 In the program, every time Farmer John lays down another fence, a new triangle is formed with his new position, the origin, and the previous position (the last fence he laid down). We can calculate the area of this new triangle. After he lays down all the fences, we can sum the areas of the multiple triangles to find if the final direction was positive (clockwise) or negative (counterclockwise).
 
-In terms of physics, [cross product](https://en.wikipedia.org/wiki/Cross_product) is twice the area of the triangle between the two vectors (positive/negative). Every time Farmer John lays down a fence, we can treat it as a vector and compute the cross product of this vector against the previous one. By adding the cross products together, we get twice the final area of the polygon formed by these vectors. Based on the positive/negative value of the area, we can know if it's counterclockwise or clockwise (vector directing up or down with the right hand rule). The process is also how the shoelace formula is proved.
+In terms of physics, the [cross product](https://en.wikipedia.org/wiki/Cross_product) of two vectors is twice the signed area of the triangle between them. Every time Farmer John lays down a fence, we can treat it as a vector and compute the cross product of this vector against the previous one. By adding the cross products together, we get twice the final area of the polygon formed by these vectors. Based on the sign of the area, we can know if it's counterclockwise or clockwise (vector directing up or down with the right hand rule). The process is also how the shoelace formula is proved.
 
 
 <LanguageSection>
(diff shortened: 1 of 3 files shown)


solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -0,0 +1,47 @@
...
+[Official Analysis (C++)](http://usaco.org/current/data/sol_prob3_bronze_feb21.html)
+
+## Solution (Vector Cross Product)
+
+Vector cross product can be used to determine if 2 vectors are rotating clockwise or anticlockwise.
+For this question, if we treat each step that Farmer John walks as a vector and do cross production of the current step against
+previous step and sum all the cross product values. We can determine the overall direction is CW or CCW based on the sign of sum.
+
+<LanguageSection>
+<PySection>
```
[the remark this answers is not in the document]
maggie-j-liu: this change was somehow reverted?

solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -0,0 +1,47 @@
...
+## Implementation
+
+```python
+def cross_product(p1, p2):
```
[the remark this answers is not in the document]
maggie-j-liu: not resolved; the code is still indented with spaces

solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -0,0 +1,46 @@
...
+<LanguageSection>
+<PySection>
+
+## Implementation
+
+```python
+def cross_product(p1, p2):
+	return p1[0] * p2[1] - p2[0] * p1[1]
+
+for i in range (int(input())):
```
maggie-j-liu, line 23:
```suggestion
for i in range(int(input())):
```

maggie-j-liu (approved):

solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -0,0 +1,46 @@
...
+source: USACO Bronze 2021 February
+title: Clockwise Fence
+author: Melody Yu
+---
+
+[Official Analysis (C++)](http://usaco.org/current/data/sol_prob3_bronze_feb21.html)
+
+## Solution (Vector Cross Product)
+
+The vector cross product can be used to determine if 2 vectors are rotating clockwise or anticlockwise. For this question, we treat each step that Farmer John walks as a vector and take the cross product of the current step and the previous step. If we sum all the cross product values, we can determine if the overall direction is CW or CCW based on the sign of the sum.
```
thecodingwizard, line 12: (oops my math is kind of bad) we're only able to sum all the cross product values and check the sign because the vector lengths are the same, right? I think

> If we sum all the cross product values, we can determine if the overall direction is CW or CCW based on the sign of the sum.

could use with a bit of elaboration as to why this works (not why cross product returns positive or negative, but why we can sum together the cross product values and still do the positive/negative check). (Or maybe I'm just really bad at math oops, does anybody else have an opinion?)

maggie-j-liu: i think taking the cross products and summing them is essentially the same thing as shoelace?

cskitty: Basically cross product is twice the area of two vectors (positive/negative). By adding them together we get 2x the final area of the polygon formed by these vectors. So, based on the positive/negative value of the area, we can know if it's counterclockwise or clockwise (vector up or vector down). The process is how the shoelace formula is proved.

Should I add that then?

maggie-j-liu: sure, *twice the area of the triangle between the two vectors

solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -9,7 +9,12 @@ author: Melody Yu
 
 ## Solution (Vector Cross Product)
 
-The vector cross product can be used to determine if 2 vectors are rotating clockwise or anticlockwise. For this question, we treat each step that Farmer John walks as a vector and take the cross product of the current step and the previous step. If we sum all the cross product values, we can determine if the overall direction is CW or CCW based on the sign of the sum.
+To find the area of a triangle with a vertex at (0, 0), a second vertex at (x1, y1) and third vertex at (x2, y2), we can use the formula (x1y2 - y1x2)/2. This value is positive if x1/y1 is greater than x2/y2, since the slope becomes smaller, making it clockwise. Likewise, a negative area means the direction is counterclockwise.
```
maggie-j-liu, line 12:
```suggestion
To find the area of a triangle with a vertex at $(0, 0)$, a second vertex at $(x_1, y_1)$ and third vertex at $(x_2, y_2)$, we can use the formula $\frac{x_1y_2 - y_1x_2}{2}$. This value is positive if $\frac{x_1}{y_1}$ is greater than $\frac{x_2}{y_2}$, since the slope becomes smaller, making it clockwise. Likewise, a negative area means the direction is counterclockwise.
```

solutions/USACO feb21-clockwise-fence.mdx
```diff
@@ -9,7 +9,12 @@ author: Melody Yu
...
+
+In the program, every time Farmer John lays down another fence, a new triangle is formed with his new position, the origin, and the previous position (the last fence he laid down). We can calculate the area of this new triangle. After he lays down all the fences, we can sum the areas of the multiple triangles to find if the final direction was positive (clockwise) or negative (counterclockwise).
+
+In terms of physics, [cross product](https://en.wikipedia.org/wiki/Cross_product) is twice the area of the triangle between the two vectors (positive/negative). Every time Farmer John lays down a fence, we can treat it as a vector and compute the cross product of this vector against the previous one. By adding the cross products together, we get twice the final area of the polygon formed by these vectors. Based on the positive/negative value of the area, we can know if it's counterclockwise or clockwise (vector directing up or down with the right hand rule). The process is also how the shoelace formula is proved.
```
maggie-j-liu, line 16: not just physics... cross product is also math

maggie-j-liu, line 16:
```suggestion
In terms of physics, the [cross product](https://en.wikipedia.org/wiki/Cross_product) of two vectors is twice the signed area of the triangle between them. Every time Farmer John lays down a fence, we can treat it as a vector and compute the cross product of this vector against the previous one. By adding the cross products together, we get twice the final area of the polygon formed by these vectors. Based on the sign of the area, we can know if it's counterclockwise or clockwise (vector directing up or down with the right hand rule). The process is also how the shoelace formula is proved.
```

SansPapyrus683: Maybe just remove the "In terms of physics" entirely.

solutions/USACO feb21-clockwise-fence.mdx
maggie-j-liu: this hasn't been fixed yet...

solutions/USACO feb21-clockwise-fence.mdx
maggie-j-liu: Yep, removing "In terms of physics" is fine

cskitty: got it, thanks

maggie-j-liu (approved):

PR #1103, opened 2021-04-04
Update LIS.mdx With Alternative RMQ Solution

Aryansh-S:
Alternative RMQ Solution added

### DIFF
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -200,15 +200,15 @@ def find_lis(arr: List[int]) -> int:
 
 Note that the following solution assumes knowledge of a basic PURQ data structure for RMQ. Suitable implementations include a segment tree or a modified Fenwick tree, but both of these are generally considered Platinum level topics. However, this solution has the advantage of being more intuitive if you're deriving LIS on the spot. 
 
-Let $\text{dp}[k]$ be the length of the LIS of $\text{a}$ that ends with the element $k \in \text{a}.$  Our base case is obviously $\text{dp}[a[0]] = 1$ (we have an LIS containing only $\text{a}[0],$ which has length $1$). Since LIS exhibits optimal substructure, we transition as follows, processing $\forall t \in \text{a}[1 \dots n-1]$ from left to right (which we must do in order to maintain the property that our subsequence strictly increases in this direction):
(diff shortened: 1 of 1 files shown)


dongliu0426 (changes requested):
wrap `dp` with \texttt 

no need to textify a, just put $a$

dongliu0426:
this could also be done w/o coordinate compression:

1. sort by value
2. for each element, `dp[i] = query(0, i - 1)` where i is the position of the element
3. update `dp[i]` into the DS

Aryansh-S:
> this could also be done w/o coordinate compression:
> 
> 1. sort by value
> 2. for each element, `dp[i] = query(0, i - 1)` where i is the position of the element
> 3. update `dp[i]` into the DS

true but that's equivalent to coord compression

dongliu0426 (approved):

content/4_Gold/LIS.mdx
```diff
@@ -192,6 +192,72 @@ def find_lis(arr: List[int]) -> int:
 
 </LanguageSection>
 
+### Fast Solution (RMQ Data Structures)
+
+(By Aryansh Shrivastava) 
+
+**Time Complexity**: $\mathcal O(N \log N)$, but perhaps a constant factor slower than binary search.
```
dolphingarlic, line 199: I don't think you need to mention the constant factor, since a BIT should run at about the same speed as a normal binary search (if not faster)

content/4_Gold/LIS.mdx
```diff
@@ -192,6 +192,72 @@ def find_lis(arr: List[int]) -> int:
...
 
+### Fast Solution (RMQ Data Structures)
+
+(By Aryansh Shrivastava) 
+
+**Time Complexity**: $\mathcal O(N \log N)$, but perhaps a constant factor slower than binary search. 
+
+Note that the following solution assumes knowledge of a basic PURQ data structure for RMQ. Suitable implementations include a segment tree or a modified Fenwick tree, but both of these are generally considered Platinum level topics. However, this solution has the advantage of being more intuitive if you're deriving LIS on the spot. 
+
+Let $\texttt{dp}[k]$ be the length of the LIS of $a$ that ends with the element $k \in a.$  Our base case is obviously $\texttt{dp}[a[0]] = 1$ (we have an LIS containing only $a[0],$ which has length $1$). Since LIS exhibits optimal substructure, we transition as follows, processing $\forall t \in a[1 \dots n-1]$ from left to right (which we must do in order to maintain the property that our subsequence strictly increases in this direction):
```
dolphingarlic, line 203:
```suggestion
Let $\texttt{dp}[k]$ be the length of the LIS of $a$ that ends with the element $k \in a$. Our base case is obviously $\texttt{dp}[a[0]] = 1$ (we have an LIS containing only $a[0]$, which has length $1$). Since LIS exhibits optimal substructure, we transition as follows, processing $\forall t \in a[1 \dots n-1]$ from left to right (which we must do in order to maintain the property that our subsequence strictly increases in this direction):
```

dolphingarlic (changes requested):

PR #1111, opened 2021-04-04
Puts DP on Bitmasks and Ranges in Gold 

Aryansh-S:
Moves from plat to gold and updates ordering.ts as discussed in the issue earlier


bqi343 (commented):
might want to rename DP on Bitmasks / DP on Ranges to Bitmask DP and Range DP respectively to be consistent with "Knapsack DP"?

Aryansh-S:
> might want to rename DP on Bitmasks / DP on Ranges to Bitmask DP and Range DP respectively to be consistent with "Knapsack DP"?

Done, go ahead and check if this works and merge if everything checks out

bqi343 (approved):
will wait for some more reviews

dongliu0426 (approved):

dolphingarlic (approved):

dolphingarlic:
Should I merge?

thecodingwizard:
Can we add redirects from the old URL? Front matter accepts a redirects field that's an array of old URL paths

Aryansh-S:
> Can we add redirects from the old URL? Front matter accepts a redirects field that's an array of old URL paths
> 
> Also does frequency need to be updated?

Probably a good idea to mention the people who handle that

content/4_Gold/DP_Bitmasks.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 id: dp-bitmasks
-title: 'Dynamic Programming on Bitmasks'
+title: 'Bitmask DP'
 author: Michael Cao, Siyong Huang, Andrew Wang, Neo Wang
 prerequisites:
   - intro-bitwise
```
thecodingwizard, line 6: Should add a redirect somewhere in the frontmatter:

```
redirects:
  - /plat/dp-bitmasks
```

thecodingwizard (changes requested):

thecodingwizard:
@Aryansh-S I think we can just set frequency to 0 for now unless somebody else has a different opinion?

Farrius (approved):

andrewwangva:
> @Aryansh-S I think we can just set frequency to 1 for now (for both modules) unless somebody else has a different opinion?

1 seems good

Aryansh-S:
I've changed the frequencies for both to 1. @thecodingwizard after your review, I think you should merge

PR #1114, opened 2021-04-05
Added Python and Java support for Global Warming

SansPapyrus683:
> We can view the temperatures as an array $v$. We want to decrement one
contiguous interval by a value $d \le x$ such that the length of the longest
increasing subsequence is longest possible.

I don't know how to reword this so a question won't be "Wait, doesn't it say that we can increase _or_ decrease in the problem statement?"


dongliu0426 (approved):

dongliu0426 (approved):
didn't i already review this 🤔 ?

SansPapyrus683:
I just wanted to know what you thought about the wording changes.

PR #1139, opened 2021-04-07
Problem suggestion automatic pr

jeffreymeng:


src/models/problem.ts
```diff
@@ -253,9 +258,104 @@ export function getProblemURL(
...
+  name: string,
+  url: string
+): ProblemSolutionMetadata | null {
+  if (isUsaco(source)) {
+    return {
+      kind: 'USACO',
+      usacoId: getTrailingCodeFromProblemURL(url),
+    };
+  } else if (source == 'IOI') {
+    for (let i = 1994; i <= 2030; ++i) {
```
dolphingarlic, line 332: Should this be something `i <= 2089` instead?

PR #1155, opened 2021-04-08
Storybook

thecodingwizard:


.github/workflows/test.yml
```diff
@@ -31,6 +33,13 @@ jobs:
       - name: Run eslint
         run: yarn eslint --quiet
 
+      - name: Run chromatic
```
jeffreymeng, line 36: wouldn't it be better to split this up into multiple actions so you can see part failed more easily

PR #1158, opened 2021-04-09
feat: validate problem source (and fix illegal sources)

jeffreymeng:


thecodingwizard:
closing as there's too many sources -- easier to allow other sources instead.

jeffreymeng:
What sources? Wouldn't those need tool tips too?

PR #1275, opened 2021-04-10
Renamed convex hull section to just geometry

dolphingarlic:


bqi343 (changes requested):
description should be changed (previous description referred to nearly all platinum geometry problems being convex hull)

content/ordering.ts
```diff
@@ -176,8 +176,8 @@ const MODULE_ORDERING: { [key in SectionID]: Chapter[] } = {
       items: ['bin-jump', 'merging', 'hld', 'centroid'],
     },
     {
-      name: 'Convex Hull',
-      description: 'Most Platinum geometry problems.',
+      name: 'Geometry',
+      description: 'More advanced concepts in computational geometry.',
```
bqi343, line 180: > More advanced

more advanced than what? Geometry hasn't appeared earlier in the guide :)

bqi343 (changes requested):

content/ordering.ts
dolphingarlic: There's rectangle geometry in bronze

bqi343: fair enough

bqi343 (approved):

PR #1351, opened 2021-04-18
- angry cows from intro dp

bqi343:
see https://forum.usaco.guide/t/introduction-to-dp-angry-cows/1168/3 for reasons


thecodingwizard:
agreed!

PR #1359, opened 2021-04-19
Problem Suggestion: Add "ben - testing" to Advanced - Counting Minimums with Segment Tree

jeffkmeng:
User `CdRpIRR4jeYtAnzeP89Q7fAWeJp1` suggested adding the problem [ben - testing](https://usaco.guide/adv/springboards?lang=cpp) to the `lazySegCnt` table of the module [Advanced - Counting Minimums with Segment Tree](https://usaco.guide/adv/CntMin#problemlist-lazySegCnt).

**Automatically Generated JSON:**
```json
{
  "uniqueId": "tlx-BenTesting",
  "name": "ben - testing",
  "url": "https://usaco.guide/adv/springboards?lang=cpp",
  "source": "TLX",
  "difficulty": "Normal",
  "isStarred": false,
  "tags": [
    ""
  ],
  "solutionMetadata": {
    "kind": "autogen-label-from-site",
    "site": "TLX"
  }
}
```
**Additional Notes**: None

*This PR was automatically generated from a user-submitted problem suggestion on the USACO guide.*

PR #1361, opened 2021-04-19
prob-suggest-updates

bqi343:
- split tags same way regardless of whether in editor or not
 - only one alert when source is not given

looks like

https://github.com/cpinitiative/usaco-guide/pull/1359#issuecomment-822114877

still not resolved though


jeffreymeng (approved):

bqi343:
I also get the following when I try to submit a problem with USACO as the source but with an improperly formatted URL:

Screen Shot 2021-04-18 at 10 19 08 PM

Don't know what to change to make the error message more informative.

thecodingwizard:
Seems like a firebase function bug. @jeffreymeng ? Is it worth moving more of the logic into the frontend rather than the function?

PR #1392, opened 2021-04-23
Update All_Roots.mdx

Qw3rTy827:
I have updated the DP on Trees - Solving For All Roots section in Gold by adding an explanation of the technique used for the example problem and commented code.

### DIFF
--- a/content/4_Gold/All_Roots.mdx
+++ b/content/4_Gold/All_Roots.mdx
@@ -32,13 +32,13 @@ Define $f[x]$ as the maximum distance from node $x$ to any node in the subtree r
 
 Define $g[x]$ as the maximum distance from node $x$ to any node outside of the subtree rooted at $x$.
 
-Then the answer for node $x$ = $max(f[x],g[x])$
+Then the answer for node $x$ = $\max(f[x],g[x])$
 
-$f[x]$ can be calculated using a DFS since $f[x]$ = $max(f[c])+1$, where $c$ is a child of $x$.
+$f[x]$ can be calculated using a DFS since $f[x]$ = $\max(f[c])+1$, where $c$ is a child of $x$.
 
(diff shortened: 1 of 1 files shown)


content/4_Gold/All_Roots.mdx
```diff
@@ -24,54 +24,91 @@ alternate solution to the problem.
...
+
+Define $g[x]$ as the maximum distance from node $x$ to any node outside of the subtree rooted at $x$.
+
+Then the answer for node $x$ = $max(f[x],g[x])$
+
+$f[x]$ can be calculated using a DFS since $f[x]$ = $max(f[c])+1$, where $c$ is a child of $x$.
+
+$g[x]$ can also be calculated using a DFS as $g[c]$ = $max(g[x]+1,f[d]+1)$, where $c$ and $d$ are both children of $x$ with $c \neq d$.
+
+To calculate $g$ in linear time, we can define another array $h$ such that $h[x]$ is the second largest distance from node $x$ to any node in the subtree rooted at $x$. So if $f[x]$ is transitioned from the branch with $c$, $g[c]$ = $max(g[x]+1,h[x]+1)$. Otherwise $g[c]$ = $max(g[x]+1,f[x]+1)$.
```
dongliu0426, line 41:
```suggestion
To calculate $g$ in linear time, we can define another array $h$ such that $h[x]$ is the second largest distance from node $x$ to any node in the subtree rooted at $x$. So if $f[x]$ is transitioned from the branch with $c$, $g[c]$ = $\max(g[x]+1,h[x]+1)$. Otherwise $g[c]$ = $max(g[x]+1,f[x]+1)$.
```

dongliu0426 (changes requested):
i think the code before is fine

dongliu0426 (changes requested):
looks good, just change all $max(..., ...)$ to $\max(..., ...)$

dongliu0426 (approved):

PR #1398, opened 2021-04-24
Editor v2

thecodingwizard:


dongliu0426:
when fetching `.problems.json` gives an 404 error, it should just return an empty json file instead of `404: Not Found`, which results in 
```
An error occurred while generating the problem solution info. This typically indicates an error in the problems JSON file:
SyntaxError: Unexpected token : in JSON at position 3

This error has also been logged to the console.
```

PR #1404, opened 2021-04-26
Cleaner Solution Code for Xenia and Tree

dutinmeow:
The previous solution was pretty blocky, so I sorted everything out into functions and added comments.

### DIFF
--- a/content/5_Plat/Centroid.mdx
+++ b/content/5_Plat/Centroid.mdx
@@ -260,8 +260,6 @@ bool r[100005];
 // r[] stores if a node has been deleted
  
 int dfs_w(int u, int p = 0) {
-	if (r[u])
-		return 0;
 	W[u] = 1;
 	for (int v : adj[u])
 		if (v != p && !r[v])
--- a/content/5_Plat/Centroid.mdx
+++ b/content/5_Plat/Centroid.mdx
@@ -241,9 +241,6 @@ private static class Centroid {
 #include <iostream>
 #include <algorithm>
 #include <vector>
-#pragma GCC optimize ("Ofast")
-#pragma GCC target ("avx2")
-#pragma GCC optimize("unroll-loops")
 using namespace std;
  
 const int inf = 2000000011;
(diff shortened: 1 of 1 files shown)


content/5_Plat/Centroid.mdx
```diff
@@ -235,99 +235,142 @@ private static class Centroid {
...
-std::vector<info> g[MN];
-
-int dfs(int n, int p=0)
-{
-	s[n]=1;
-	for(auto x:a[n])
-		if(x!=p&&!r[x])
-			s[n]+=dfs(x, n);
-	return s[n];
+#pragma GCC optimize ("Ofast")
```
bqi343, line 244: > #pragma GCC optimize ("Ofast")

don't include these, they're not relevant to the solution

content/5_Plat/Centroid.mdx
```diff
@@ -235,99 +235,142 @@ private static class Centroid {
...
-			t=n, b=v[n].size(), d=1;
-			dfs2(x, n);
-			v[n].push_back(INF);
-		}
-	r[n]=1;
-	for(auto x:a[n])
-		if(!r[x])
-			centroid(x);
+ 
+// this stuff is for tree distance, see binary jumping mondule
```
bqi343, line 292: there's no need to use binary jumping

bqi343 (changes requested):

content/5_Plat/Centroid.mdx
```diff
@@ -235,99 +235,142 @@ private static class Centroid {
...
-{
-	for(auto x:a[n])
-		if(!r[x]&&x!=p&&s[x]*2>ms)
-			return find(x, ms, n);
-	return n;
+ 
+int centroid(int u, int n, int p = 0) {
+    for (int v : adj[u])
+        if (v != p && !r[v])
+            if (W[v] * 2 > n)
```
bqi343, line 278: also tabs

bqi343 (changes requested):

content/5_Plat/Centroid.mdx
dutinmeow: How would you do it without binary jumping?

bqi343: solution on the left doesn't use binary jumping

dutinmeow: Both tutorials linked on the article use the binary jumping/lca method, so I think its more intuitive

content/5_Plat/Centroid.mdx
```diff
@@ -235,99 +235,139 @@ private static class Centroid {
...
+// ans[] is closest red node in subtree of centroid tree
+// adj[] stores original tree given in input
+// col[] stores if node has already been colored red (optional)
+ 
+bool r[100005];
+// r[] stores if a node has been deleted
+ 
+int dfs_w(int u, int p = 0) {
+	if (r[u])
+		return 0;
```
frodakcin, line 264:
```suggestion
	if (r[u])
		return 0;
```

This is redundant, right? Probably can delete.

frodakcin (commented):
Looks fine. Would it be better to cut out extra log factor though?

content/5_Plat/Centroid.mdx
dutinmeow: > Looks fine. Would it be better to cut out extra log factor though?

Both Galen Colin's tutorial and the other one mention using some method to find distance. I could use sparse table to do it in constant time, but I feel like binary jumping is just more intuitive from an educational standpoint, seeing as it's part of the platinum topics. In the end, it really doesn't matter as long as the solution pass the tests.

thecodingwizard: maybe it suffices to just mention that it's possible to cut out the extra log factor? @frodakcin @bqi343

bqi343 (changes requested):

content/5_Plat/Centroid.mdx
bqi343: Just do it in the same way as the original solution does? It's quite simple and there's no extra log factor.

content/5_Plat/Centroid.mdx
bqi343: oops started this comment a few days ago but forgot to submit it

dutinmeow: I'll work on the implementation, but I'm quite busy the next few days due to schoolwork.

bqi343:
closing as desired changes have not been made

PR #1422, opened 2021-04-28
Update USACO jan20-word-bronze.mdx

123456789x:
Add Python implementation.


maggie-j-liu:
This problem already has a python implementation, so I don't think another one is necessary.

When adding solutions, make sure to indent your code with tabs instead of spaces.

PR #1432, opened 2021-04-30
Change floodfill diagram

### LINKED ISSUE #741: Flood Fill - Change Diagrams
> I think it may be clearer if we have the red square backtrack instead of jumping from the bottom square to the top. It may not be intuitive why the red square knows to go there.

> The only problem would be that this would add like 4 more squares, making it really long. Maybe we can have 2 or 3 square on each row?

maggie-j-liu:
closes #741

### DIFF
--- a/content/3_Silver/Flood_Fill.mdx
+++ b/content/3_Silver/Flood_Fill.mdx
@@ -59,6 +59,8 @@ uncolored for nodes not yet visited.
 
 <center>
 
+<div className="grid grid-cols-3 gap-y-12 justify-evenly place-items-center max-w-lg">
+
 <table className="text-center">
 	<tbody>
 		<tr>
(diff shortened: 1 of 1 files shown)


thecodingwizard:
- should the red cell return to (0, 0)? (so the last grid would have its (0, 0) cell colored red?)
- mobile friendly? maybe `grid-cols-2 sm:grid-cols-3 space-y-6 sm:space-y-12`?

maggie-j-liu:
> should the red cell return to (0, 0)? (so the last grid would have its (0, 0) cell colored red?)

should that be the second to last grid, and then the last grid is the same (to show the outcome of the floodfill), or just replace the current last grid?

thecodingwizard:
either works. (9 cells looks nice but showing the outcome is also good :P)

PR #1450, opened 2021-05-03
Add CF Solve the Maze Problem to Floodfill

23danielsuh:


solutions/CF Solve The Maze.mdx
```diff
@@ -0,0 +1,138 @@
...
+```
+#####
+##BG#
+###..
+####.
+```
+
+Because the good person is adjacent to the bad person, the bad person can simply just move right, and now has full access to wherever the good person can move to. It won't matter if the good person can make it, as the bad person can make it as well. Furthermore, replacing a good person with a wall is not allowed, so preventing the movement of the bad person is not allowed. 
+
+With this observation, the solution is relatively simple. Check for this adjacency, and if it is, print "No." Then, surround the bad people with walls. Floodfill from the end point ($$N - 1,\space M - 1$$), and make sure that all the good people have been visited. You do not need to check if the bad people did not make it, as surrounding the bad people with walls is sufficient.
```
frodakcin, line 23:
```suggestion
With this observation, the solution is relatively simple. Check for this adjacency, and if it exists, print "No." Otherwise, proceed by surrounding the bad people with walls. Floodfill from the end point $(N - 1,\space M - 1)$, and make sure that all the good people have been visited. You do not need to check if the bad people did not make it, as surrounding the bad people with walls is sufficient. 
```

optional wording changes

frodakcin (approved):
Looks good.

jessechoe10 (approved):

PR #1461, opened 2021-05-05
CSES Counting Paths Solution

dutinmeow:
I'm not sure if this problem is in any modules.

### DIFF
--- a/solutions/CSES Counting Paths.mdx
+++ b/solutions/CSES Counting Paths.mdx
@@ -1,0 +1,108 @@
+---
+id: cses-1136
+source: CSES
+title: Counting Paths
+author: Dustin Miao
+---
+
+**Time Complexity**: $\mathcal O(m\log n + n)$
+
+For each path $(a, b)$, we can split it into two paths: $(a, lca(a, b)), (b, lca(a, b))$. We 
+can find the LCA with any method, but the easiest is with binary lifting. For each node, 
+keep track of the number paths that begin at that node and the number that end at that node. 
+
+If $k$ is the number of paths coming from a node's children, $s$ is the number of paths 
+starting at the current node, and $e$ is the number of paths ending at the current node, 
+then the answer for that node is $k + s - e / 2$. The $e / 2$ is because by splitting up
+the paths, we double count the lca for each path. The number of paths going to the node's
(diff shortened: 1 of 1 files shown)


maggie-j-liu:
I don't think it is... if it's not on the guide, why are you adding a solution? If you think it should be added, submit a problem suggestion!

dutinmeow:
> I don't think it is... if it's not on the guide, why are you adding a solution? If you think it should be added, submit a problem suggestion!

maggie-j-liu:
the problem suggestion got merged, but then removed in #1571 so i think this can be closed

thecodingwizard:
fwiw Guide supports internal solutions for problems that appear in no modules (by adding to `extraProblems.json`). Still shows up in search engines

maggie-j-liu:
oh, i didn't know that. is this something we want to actively support, and should we add it to the adding solutions module?

thecodingwizard:
sure. I think extra internal solutions generally help:
- they boost site SEO
- they attract people who need help on that problem
- we might use that problem in the future

bqi343:
@thecodingwizard so are we keeping this or not? personally, I think it should either be converted to a solution for a max flow since that problem is in the guide

thecodingwizard:
I think we should keep this solution and add it to `extraProblems.json` for the benefits outlined above. I'm not 100% sure what you mean by "convert to a solution for max flow"...? it doesn't hurt to have solutions for more problems even if the problems are redundant right?

bqi343:
I think changing one or two lines of this solution and adding file i/o solves max flow.

bqi343:
@dolphingarlic what do you think?

dolphingarlic:
@dutinmeow I agree that we should convert this solution to work for max flow (less work and I'd prefer USACO over CSES when a problem appears in both)

dolphingarlic (changes requested):
.

dutinmeow:
I have code written for both so it shouldn't take that long

dutinmeow:
I'll close this PR and make a new one because the revisions are quite big

dutinmeow:
The new PR is [here](https://github.com/cpinitiative/usaco-guide/pull/1903)

PR #1470, opened 2021-05-06
Add internal solution to clocktree feb20  

cskitty:
Describes the solution from bi-parte graph. Added a video solution. (don't know what is the correct way to add video link though)


maggie-j-liu:
To fix the build, you need to link the internal solution to the problem as described in [step 4](https://usaco.guide/general/adding-solution?lang=cpp#steps). I pretty sure the video is linked correctly.

PR #1477, opened 2021-05-07
Added Cpp Solution for the problem solutions\USACO feb19-the-great-revegetation.mdx

atharva-adhav:
Added the Cpp solution with comments


maggie-j-liu:
This problem already has an official solution in c++, so I don't think another one is needed (since your solution uses the same method)

atharva-adhav:
Sorry I didn't check the official solution. Will close the request. Thanks!

PR #1495, opened 2021-05-09
C++ solution for 2019 December Bronze "Where Am I?"

SalarC123:
There was only a python solution, so I'm adding a c++ one. The time complexity of my solution is O(2N^2) whereas the python solution is O(N^3)

### DIFF
--- a/solutions/USACO dec19-whereami.mdx
+++ b/solutions/USACO dec19-whereami.mdx
@@ -7,7 +7,7 @@ author: Ananth Kashyap
 
 [Official Analysis](http://www.usaco.org/current/data/sol_whereami_bronze_dec19.html)
 
-## Implementation
+## Python Implementation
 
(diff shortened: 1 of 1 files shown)


maggie-j-liu:
I think this is the same method as the official solution, which is also in C++, so I don't think we need another solution. For future reference, make sure to indent your code using tabs instead of spaces.

solutions/USACO dec19-whereami.mdx
```diff
@@ -43,3 +43,59 @@ print(ans, file = open('whereami.out', 'w'))
 </PySection>
 
 </LanguageSection>
+
+
+## C++ Implementation
+
+**Time Complexity:** $\mathcal{O}(2N^2)$
```
nwatx, line 50:
```suggestion
**Time Complexity:** $\mathcal{O}(N^2)$
```

nwatx (changes requested):

PR #1537, opened 2021-05-17
test add 5 video solutions

cskitty:
added five video solutions for bronze/silver questions, please check if the formatting is okay!

### DIFF
--- a/solutions/usaco-1038.mdx
+++ b/solutions/usaco-1038.mdx
@@ -9,7 +9,7 @@ author: Qi Wang (Java), Tanish Tyagi (C++), Melody Yu (Video)
 
 ## Video Solution
 
-Note: Video solution might not be the same as written solution.
+Note: The video solution might not be the same as other solutions. Code in C++.
 <Youtube id="mt8QhnFOkr0" />
 
 
--- a/solutions/usaco-1038.mdx
+++ b/solutions/usaco-1038.mdx
@@ -12,13 +12,14 @@ author: Qi Wang (Java), Tanish Tyagi (C++), Melody Yu (Video)
 Note: Video solution might not be the same as written solution.
 <Youtube id="mt8QhnFOkr0" />
 
+
+## Implementation
+
 **Time Complexity:** $\mathcal{O}((N+M)\log (maxDist))$
 
 <LanguageSection>
 
 <CPPSection>
-
-## C++ Implementation
 
 ```cpp
 #include <bits/stdc++.h>
--- a/solutions/usaco-1038.mdx
+++ b/solutions/usaco-1038.mdx
@@ -2,15 +2,20 @@
 id: usaco-1038
 source: USACO Silver 2020 US Open
 title: Social Distancing
-author: Qi Wang (Java), Tanish Tyagi (C++)
+author: Qi Wang (Java), Tanish Tyagi (C++), Melody Yu (Video)
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_socdist_silver_open20.html)
 
+## Video Solution
+
+Note: Video solution might not be the same as written solution.
+<Youtube id="mt8QhnFOkr0" />
+
 **Time Complexity:** $\mathcal{O}((N+M)\log (maxDist))$
 
 <LanguageSection>
-	
+
 <CPPSection>
 
 ## C++ Implementation
--- a/solutions/usaco-1084.mdx
+++ b/solutions/usaco-1084.mdx
@@ -2,10 +2,15 @@
 id: usaco-1084
 source: USACO Bronze 2021 January
 title: Even More Odd Photos
-author: Sathvik Chundru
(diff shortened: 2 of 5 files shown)


solutions/usaco-1084.mdx
```diff
@@ -2,11 +2,16 @@
...
-author: Sathvik Chundru
+author: Sathvik Chundru, Melody Yu (Video)
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_prob2_bronze_jan21.html)
 
+## Video Solution
+
+Note: Video solution might not be the same as written solution.
+<Youtube id="VbPjZamNA40&t=1s" />
```
maggie-j-liu, line 13:
```suggestion
<Youtube id="VbPjZamNA40" />
```
it doesn't work right now, but this should fix it

solutions/usaco-1038.mdx
```diff
@@ -2,15 +2,20 @@
...
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_socdist_silver_open20.html)
 
+## Video Solution
+
+Note: Video solution might not be the same as written solution.
+<Youtube id="mt8QhnFOkr0" />
+
 **Time Complexity:** $\mathcal{O}((N+M)\log (maxDist))$
```
maggie-j-liu, line 15: maybe move this under `## C++ Implementation`?

solutions/usaco-1038.mdx
```diff
@@ -2,15 +2,20 @@
...
-author: Qi Wang (Java), Tanish Tyagi (C++)
+author: Qi Wang (Java), Tanish Tyagi (C++), Melody Yu (Video)
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_socdist_silver_open20.html)
 
+## Video Solution
+
+Note: Video solution might not be the same as written solution.
+<Youtube id="mt8QhnFOkr0" />
```
maggie-j-liu, line 13: maybe videos should also be put in `LanguageSection`s? what does everyone else think?

maggie-j-liu (changes requested):

solutions/usaco-1038.mdx
cskitty: The code is in C++ but the algorithm description (first half or so) applies to any language; I'd be okay with moving it if needed, though.

maggie-j-liu: oops, i meant the time complexity, cause right now it looks like the time complexity for the video instead of the code

cskitty: Oh I think that was put there by someone else before I committed https://github.com/cpinitiative/usaco-guide/blob/master/solutions/usaco-1038.mdx

solutions/usaco-1038.mdx
cskitty: The code is in C++ but the algorithm description (first half or so) applies to any language; I'd be okay with moving it if needed, though.

solutions/usaco-1038.mdx
maggie-j-liu: yeah, i know you didn't put that, but because the video was added it looks kind of confusing now. can you move it below the implementation header? 

or even better; move the header and tc outside of the language section (replace C++ Implementation with just Implementation)
something like this https://usaco.guide/general/adding-solution#steps

solutions/usaco-1038.mdx
thecodingwizard: I'm somewhat against this since I think it's unlikely people will make three videos for each problem; if there is multi language support it's probably just all in one video. So maybe it's sufficient to list out what language(s) the video uses?

solutions/usaco-1038.mdx
cskitty: Got it. How is it now?

maggie-j-liu: Yep, it looks good now!

solutions/usaco-1038.mdx
dongliu0426: if the language isn't available, it defaults to the language that is available.

maggie-j-liu: yeah, which is why it would be fine to put the video in LanguageSection (since all users would still be able to see it)... but it also makes sense that there won't be that many videos for each problem and that each video can have multiple languages so it's not really necessary

thecodingwizard:
Would the note be clearer as:

> Note: Video solution might not be the same as written solution.

or 

> Note: The video solution might not be the same as other solutions.

I'm fine with either, though with the former I think it's possible that some people may mistakenly think that the "implementation" section that immediately follows the video solution is the implementation from the video? not sure how big of an issue this is.

Also, is it worth adding "Code in C++" or something equivalent?

---

Videos are excellent, thanks! :)

thecodingwizard (approved):

PR #1547, opened 2021-05-18
Improved some bronze modules

SansPapyrus683:
Added some stuff to the time complexity + rect geo module.

### DIFF
--- a/content/2_Bronze/Rect_Geo.mdx
+++ b/content/2_Bronze/Rect_Geo.mdx
@@ -244,10 +244,9 @@ public class Billboard {
 
 <PySection>
 
-Code runs marginally faster in Python when placed in a function
-(see [this](https://stackoverflow.com/questions/11241523/why-does-python-code-run-faster-in-a-function) SO thread),
-so we can use this to get all 10 test cases - when taken out of
-the function, the code passes only 9.
+Code runs marginally faster in Python [when placed in a function](https://stackoverflow.com/questions/11241523/why-does-python-code-run-faster-in-a-function),
+so we can use this to get all 10 test cases. When taken out of
+the function, the code passes only 9 test cases.
 
 ```py
 import sys
--- a/content/2_Bronze/Rect_Geo.mdx
+++ b/content/2_Bronze/Rect_Geo.mdx
@@ -78,11 +78,11 @@ int main() {
 	int a, b, c, d;
 	cin >> a >> b >> c >> d;
 
-	int union = (b - a) + (d - c);  // the sum of the two intervals
+	int total = (b - a) + (d - c);  // the sum of the two intervals
 	int intersection = max(min(b, d) - max(a, c), 0);  // subtract the intersection
-	int total = union - intersection;
+	int union = total - intersection;
 
-	cout << total << "\n";
+	cout << union << "\n";
 }
 ```
 
@@ -105,13 +105,13 @@ public class Paint {
 		int d = Integer.parseInt(second.nextToken());
 
 		// the sum of the two intervals
-		int union = (b - a) + (d - c);
+		int total = (b - a) + (d - c);
 		// subtract the intersection
 		int intersection = Math.max(Math.min(b, d) - Math.max(a, c), 0);
-		int total = union - intersection;
+		int union = total - intersection;
 
 		PrintWriter writer = new PrintWriter("paint.out");
-		writer.println(total);
+		writer.println(union);
 		writer.close();
 	}
 }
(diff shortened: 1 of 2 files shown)


mizlan:
Next time you should split it into two PRs.

SansPapyrus683:
I would do that by splitting my fork into 2 branches, right?

mizlan:
Ideally you would have two branches as well, since maintainers may only want one of your two changes.

SansPapyrus683:
Isn't that... basically what I just said?

content/2_Bronze/Time_Comp.mdx
```diff
@@ -137,14 +137,14 @@ while (i < n) {
...
-	# constant time code here
+for i in range(1, n + 1):
+	pass # constant time code here
 ```
 
 ```py
 i = 0
 while (i < n):
-	# constant time code here
+	pass # constant time code here
```
mizlan, line 147: I'm not sure if I like the use of `pass` here. Especially for the while loop which already runs fine without the `pass`, since it's another thing to search up. Have you considered putting some "dumb" placeholder code as an alternative?

mizlan:
Oh yea, my bad. Brain wasn't working.

content/2_Bronze/Rect_Geo.mdx
```diff
@@ -164,72 +202,82 @@ int main() {
...
 ```
 
 </JavaSection>
 
 <PySection>
 
-Unfortunately, the naive Python solution only passes 9/10 test cases.
+Code runs marginally faster in Python when placed in a function, so we can use
+this to get all 10 test cases (When taken out of the function, the code passes
+only 9).
```
mizlan, line 249: This is good information to know, but it may be more confusing than helpful to people new to the subject. Also, saying that the naive solution doesn't work is perfectly fine since there's a better solution anyway.

content/2_Bronze/Rect_Geo.mdx
```diff
@@ -78,18 +78,47 @@ int main() {
 	int a, b, c, d;
 	cin >> a >> b >> c >> d;
 
-	int ans = 0;
-	ans += b-a; // add length of first interval
-	ans += d-c; // add length of second interval
-	// if intervals intersect, subtract the length of the intersection
-	ans -= max(min(b, d)-max(a, c), 0);
+	int union = (b - a) + (d - c);  // the sum of the two intervals
```
mizlan, line 81: The terminology used here is incorrect. This isn't a union; the union is the final answer. I'm not even sure what to call this intermediate value, I think "total" is more suitable, as in "total paint coats."

mizlan (changes requested):

content/2_Bronze/Rect_Geo.mdx
SansPapyrus683: Well, the C++ and Java solution both work, so it's good for the Python solution to be consistent.

content/2_Bronze/Time_Comp.mdx
SansPapyrus683: `pass`'s one purpose is to be a placeholder, isn't it?

thecodingwizard: Isn't pass unnecessary since there's i+=1?

content/2_Bronze/Rect_Geo.mdx
mizlan: Yes but it's not instructive in any way and has the potential to cause confusion, the way I see it.

SansPapyrus683: Not really. All I'm saying is that Python code has the potential to run faster when placed in a function. I can add an SO thread to help explain it.

thecodingwizard: maybe link something for the "code runs marginally faster in Python when placed in a function" part?

dutinmeow: [Try this?](https://www.tutorialspoint.com/Why-does-Python-code-run-faster-in-a-function#:~:text=It%20is%20generally%20found%20that,we%20download%20from%20Python.org.)

SansPapyrus683: I was actually thinking of this thread here.

dutinmeow: Both say pretty much the same thing, I think either one works

content/2_Bronze/Rect_Geo.mdx
```diff
@@ -164,72 +202,83 @@ int main() {
...
 
 </JavaSection>
 
 <PySection>
 
-Unfortunately, the naive Python solution only passes 9/10 test cases.
+Code runs marginally faster in Python when placed in a function
+(see [this](https://stackoverflow.com/questions/11241523/why-does-python-code-run-faster-in-a-function) SO thread),
+so we can use this to get all 10 test cases - when taken out of
+the function, the code passes only 9.
```
thecodingwizard, line 250:
```suggestion
Code runs marginally faster in Python [when placed in a function](https://stackoverflow.com/questions/11241523/why-does-python-code-run-faster-in-a-function),
so we can use this to get all 10 test cases. When taken out of
the function, the code passes only 9 test cases.
```

PR #1556, opened 2021-05-20
C++ Solution for War

maggie-j-liu:
I'm not sure if the time complexity is correct... also, the explanation is kind of long, not sure how to shorten it?


dongliu0426 (changes requested):
you might want to merge small to large: when you're merging two people (`x`, `y`), you always want to merge the one with the smaller size to the one with the larger size so that when you call the `find(a)` function, the depth is at most logN because you're always moving to a parent that has a size of at least 2 times the current `x` (without this i think the time complexity would be O(NQ), even though it passes(?) whereas with this the time complexity would be O(Q\log N))

or you can just cache `parent[x]` (the time complexity here would be around O(Q\log N) too):
```cpp
int find(int a)
{
	if (a == parent[a])
	{
		return a;
	}
	return parent[a] = find(parent[a]);
}
```

dongliu0426 (approved):

PR #1573, opened 2021-05-22
Fix dashboard problem statistics

maggie-j-liu:
`moduleId` was deleted so the statistics section was showing 0 problems. I noticed that the dashboard problem total doesn't match the problem total in the section, is this because some problems show up in multiple sections/modules?


bqi343:
oops

PR #1587, opened 2021-05-23
allow sorting monthly tables by percent

bqi343:


src/components/markdown/ProblemsList/DivisionList/DivisionList.tsx
```diff
@@ -284,6 +285,20 @@ export function DivisionList(props): JSX.Element {
...
+
+  const sortOrders = ['By Contest'];
+  if (allHavePercent) sortOrders.push('By Percent');
+  const [sortOrder, setSortOrder] = React.useState('Sort: ' + sortOrders[0]);
+  const getDisplayedProblems = () => {
+    if (!allHavePercent || !sortOrder.endsWith('Percent')) return problems;
+    return [...problems].sort(
+      (a, b) => b.percentageSolved - a.percentageSolved
+    );
+  };
```
thecodingwizard, line 300: You can probably improve the performance by wrapping it with [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo). Not sure how necessary it is though

bqi343:
@thecodingwizard Like this?

thecodingwizard:
When there's no percent available I think the sort button should be hidden (otherwise it's weird since it looks clickable but it's not clickable)

PR #1603, opened 2021-05-26
update copy button for python

bqi343:
fixes https://usaco.guide/problems/usaco-807-teleportation/solution


thecodingwizard (approved):

maggie-j-liu (approved):
The gap between the copy button and the language gets bigger on small screens (because of padding), but that's not a big problem...

PR #1608, opened 2021-05-28
Update cses-1076.mdx

ar492:
added another solution

### DIFF
--- a/solutions/cses-1076.mdx
+++ b/solutions/cses-1076.mdx
@@ -2,16 +2,16 @@
 id: cses-1076
 source: CSES
 title: Sliding Median
-author: Isaac Noel
+author: Isaac Noel, Arpan Banerjee
 ---
 
 **Time Complexity:** $\mathcal{O}(N\log N)$
 
-## Abstract
+## Abstract 1
 
 Finding the median of a sliding window by maintaining two multisets.
 
-## Solution
+## Solution 1
 
 To accomplish finding the sorted median of a sliding window, we can store the
 window within two sets: one containing the lower values of the window and the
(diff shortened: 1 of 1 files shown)


solutions/cses-1076.mdx
```diff
@@ -87,3 +87,28 @@ int main() {
...
+        ordered_set oset;
+        for(int i=0; i<n; i++){
+                int u; cin>>u; a[i]=u;
+                oset.insert(u);
+                if(i>=k) oset.erase(--oset.lower_bound(a[i-k]));
+                if(i>=k-1) cout<<*oset.find_by_order((int)(k/2.0-0.1))<<endl;
+        }
+}
+```
+Alternatively, if you do not use C++ (or if you do), you can use a fenwick tree to simulate the aforementioned order statistic tree/indexed set. Clearly, the values of the array would need to be compressed. There will also be an extra $\log$ factor resulting from binary searching on the fenwick tree, yielding a total time complexity of $\mathcal{O}(n\log^2{n})$.
```
mizlan, line 114: Don't just say you _can_ simulate it, either provide how you would go about doing that and/or a link.

solutions/cses-1076.mdx
```diff
@@ -87,3 +87,28 @@ int main() {
 	cout << endl;
 }
 ```
+
+## Solution 2
+(Arpan Banerjee)
+
+We can directly use an [order statistic tree ](https://codeforces.com/blog/entry/11080) (C++) to get the median while sliding the window across the array.
+```cpp
```
mizlan, line 95: Put empty lines before the first \`\`\` and after the ending \`\`\`

solutions/cses-1076.mdx
```diff
@@ -2,16 +2,16 @@
...
 source: CSES
 title: Sliding Median
-author: Isaac Noel
+author: Isaac Noel, Arpan Banerjee
 ---
 
 **Time Complexity:** $\mathcal{O}(N\log N)$
 
-## Abstract
+## Abstract 1
```
mizlan, line 10: If you're gonna rename this to Abstract 1, you should have an Abstract 2 somewhere.

solutions/cses-1076.mdx
```diff
@@ -87,3 +87,28 @@ int main() {
 	cout << endl;
 }
 ```
+
+## Solution 2
+(Arpan Banerjee)
+
+We can directly use an [order statistic tree ](https://codeforces.com/blog/entry/11080) (C++) to get the median while sliding the window across the array.
+```cpp
+#include<bits/stdc++.h>
```
mizlan, line 96: This code needs to be formatted (spaces around operators, tabs, common sense stuff).

PR #1612, opened 2021-05-28
Bucket Brigade java and python solutions

maggie-j-liu:


nathangong (approved):

HarryWangATX (changes requested):
I think we are using Kattio for java io for consistency

maggie-j-liu:
Is Kattio really necessary? This is in general/expected knowledge, i think it's better to keep it simple?

nathangong:
Wouldn't Kattio make the code more simple and readable? Also, I think it says [here](https://usaco.guide/general/adding-solution?lang=java#java) that Kattio is preferred for consistency/readability.

PR #1618, opened 2021-05-29
Added persistent array implementation

dolphingarlic:
Also, should we maybe move this before introducing the persistent segtree?

### DIFF
--- a/content/6_Advanced/Persistent.mdx
+++ b/content/6_Advanced/Persistent.mdx
@@ -8,19 +8,13 @@ prerequisites:
 frequency: 1
 ---
 
-## Persistent Segment Tree
+## Persistent Array
 
-Persistent segment trees are very similar to sparse segment trees in the sense
-that we store child pointers. The key differences are that we have multiple
-roots and every time we "update" a node, we actually create a new node in its
-place (hence persistence).
+Persistent arrays are one of the simplest persistent data structures. A
+persistent array should be able to access and update its elements at given
+times.
 
-<FocusProblem problem="persistentSample" />
-
-### Persistent Array
-
-A simpler version of the persistent segment tree is the persistent array, where
-one only does **point** updates and queries.
+### Fat Nodes
 
 <LanguageSection>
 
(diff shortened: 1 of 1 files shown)


bqi343:
I was hoping for a path copying approach for persistent array before persistent segtree

dolphingarlic:
Would that have the same implementation as a persistent segtree then? (Minus the range query)

bqi343:
essentially yes

bqi343 (changes requested):
^ above

dolphingarlic:
I kept the fat nodes in because I think it's still worthwhile to know the alternative approaches, even if they are worse

content/6_Advanced/Persistent.mdx
```diff
@@ -8,18 +8,141 @@ prerequisites:
...
+}
+
+int get_item(int index, int time) {
+	// Gets the array item at a given index and time
+	return query(roots[time], index);
+}
+
+void update_item(int index, int value, int prev_time, int curr_time) {
+	// Updates the array item at a given index and time
+	roots[curr_time] = new Node(roots[prev_time]); // First, copy the root
```
bqi343, line 116: ^ isn't this line not necessary?

content/6_Advanced/Persistent.mdx
```diff
@@ -8,18 +8,141 @@ prerequisites:
...
+<CPPSection>
+
+```cpp
+struct Node {
+	int val;
+	Node *l, *r;
+
+	Node(ll x) : val(x), l(nullptr), r(nullptr) {}
+	Node(Node *ll, Node *rr) : val(0), l(ll), r(rr) {}
+	Node(Node *cp) : val(cp->val), l(cp->l), r(cp->r) {}
```
bqi343, line 83: same as below

content/6_Advanced/Persistent.mdx
```diff
@@ -8,18 +8,141 @@ prerequisites:
...
+Path copying is **fully persistent**. 
+
+## Persistent Segment Tree
+
+Since persistent arrays with path copying are so similar to sparse segment
+trees, it's relatively straightforward to convert one into a persistent
+segment tree - just add range queries!
+
+However, note that it's not possible to support range updates with range queries
+in persistent segment trees, since lazy propagation doesn't play well with path
```
bqi343, line 142: actually, lazy is possible

bqi343 (changes requested):

bqi343 (approved):

PR #1638, opened 2021-05-31
Lifeguards(784) with Java Implementation

brad182:
This problem only had a C++ implementation, so I coded up one in Java with Kattio. The two implementations solve the problem in the same way. Added notes to make the implementation clearer, all lines should be under the 80 character limit.

### DIFF
--- a/solutions/usaco-784.mdx
+++ b/solutions/usaco-784.mdx
@@ -2,16 +2,15 @@
 id: usaco-784
 source: USACO Bronze 2018 January
 title: Lifeguards
-author: Danh Ta Chi Thanh
+author: Danh Ta Chi Thanh, Brad Ma
 ---
 
 [Official Analysis (Java)](http://www.usaco.org/current/data/sol_lifeguards_bronze_jan18.html)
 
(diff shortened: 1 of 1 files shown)


maggie-j-liu:
the [official solution](http://www.usaco.org/current/data/sol_lifeguards_bronze_jan18.html) is in java, and uses the same method as your solution, so i don't think another solution is necessary...

brad182:
> the [official solution](http://www.usaco.org/current/data/sol_lifeguards_bronze_jan18.html) is in java, and uses the same method as your solution, so i don't think another solution is necessary...

 i see a few with the same language on the official solution and internal solution 

so we don't want the official solution language to be on the internal solution?

maggie-j-liu:
> It is **not** necessary to add an alternative implementation in the same language as the official analysis (unless the alternative implementation takes a different approach or is written better / differently from the official implementation).

*https://usaco.guide/general/adding-solution#steps*

If your solution is different from the official solution, or implements it better, then you should add it. If it's the same and in the same language, then it's not necessary.

PR #1659, opened 2021-06-04
Update ProblemHits title to be clickable

### LINKED ISSUE #1646: UI: problem search results not clickable
Small inconvenience that probably has gone unnoticed by common users of the webapp: when I perform a search on https://usaco.guide/problems/, the search results come in boxes, of which neither the box itself nor the title of the box are clickable. This was particularly confusing for me since I didn't realize there was anything clickable in the box for a while, since the links have gone so far as to be greyed out.

I'm thinking that the they should both link to the problem, but feel free to bring up other ideas.

nwatx:
Resolves #1646

### DIFF
--- a/src/components/ProblemsPage/ProblemHits.tsx
+++ b/src/components/ProblemsPage/ProblemHits.tsx
@@ -1,6 +1,5 @@
 import { Link } from 'gatsby';
 import * as React from 'react';
-import { Highlight } from 'react-instantsearch-dom';
 import { moduleIDToSectionMap } from '../../../content/ordering';
 import { AlgoliaProblemInfo, getProblemURL } from '../../models/problem';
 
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
image
So right now clicking the white text leads you straight to the problem statement- I think it'd be cool if it redirects to the USACO Guide URL instead.

maggie-j-liu:
> Some problems appear on the guide multiple times

which url do you suggest it directs to? just the first?

maggie-j-liu:
can the entire box be made clickable or are stacked links too complicated?

nwatx:
> can the entire box be made clickable or are stacked links too complicated?

it can, though then that would remove the purpose of the below URLs i think. @thecodingwizard opinion?

thecodingwizard:
err I think the entire box being clickable might be a bit weird.

Also, there is a planned rewrite of the Problems page (but who knows when that will happen!)

nwatx:
> image
> So right now clicking the white text leads you straight to the problem statement- I think it'd be cool if it redirects to the USACO Guide URL instead.

do you mean the solution page?

SansPapyrus683:
No, I mean a URL like [this](https://usaco.guide/gold/knapsack#problem-cses-1665).

maggie-j-liu:
> No, I mean a URL like [this](https://usaco.guide/gold/knapsack#problem-cses-1665).

The problem with that is that one problem might appear in multiple modules, and thus have multiple urls.

bqi343:
> So right now clicking the white text leads you straight to the problem statement- I think it'd be cool if it redirects to the USACO Guide URL instead.

Can't you just click on "String Searching" if you want the guide URL?

nwatx:
@thecodingwizard I suppose a temporary fix wouldn't hurt. But feel free to close this issue otherwise.

HarryWangATX (approved):

thecodingwizard:
I guess it doesn't hurt...

PR #1696, opened 2021-06-10
Inline problem name on problems page

maggie-j-liu:
The problem name being block made the star appear on the next line.

Is there anything else that needs to be changed about the problems page? The clickable problem name makes the view problem statement redundant, but since it's not obvious that the problem name is clickable, people who are used to clicking the link below might not know what happened.
The link styling is also inconsistent; only the title underlines when hovered and they're all different colors.


thecodingwizard:
the link update broke the problem name search highlight. Not sure how important it is to fix though if we're going to rewrite the problems page ~~at some point~~

maggie-j-liu:
i think i can fix the highlighting

maggie-j-liu:
Is there anything else I should fix/change?

sathu-beep (approved):

PR #1711, opened 2021-06-11
bump deps

thecodingwizard:


thecodingwizard:
oops I'll just combine this with #1714

PR #1715, opened 2021-06-12
Update usaco-574.mdx

ar492:
Added a detailed description of an existing implementation.

### DIFF
--- a/solutions/usaco-574.mdx
+++ b/solutions/usaco-574.mdx
@@ -49,7 +49,7 @@ int main() {
 
 ## Alternative Implementation - BFS/DFS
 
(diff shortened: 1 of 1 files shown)


HarryWangATX (changes requested):
Can u change the wording to be more of an exhaustive search-focused explanation, basically generalize the bfs?

HarryWangATX (approved):

dongliu0426 (changes requested):
update authors?

dongliu0426 (approved):
imo this bfs/dfs impl is basically dp but wtv

PR #1730, opened 2021-06-13
Problem Source: "CEOI 2012" -> "CEOI"

bqi343:
#1731

### DIFF
--- a/content/3_Silver/Binary_Search.problems.json
+++ b/content/3_Silver/Binary_Search.problems.json
@@ -170,9 +170,9 @@
     },
     {
       "uniqueId": "ceoi-12-JobScheduling",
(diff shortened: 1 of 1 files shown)


content/3_Silver/Binary_Search.problems.json
```diff
@@ -170,9 +170,9 @@
     },
     {
       "uniqueId": "ceoi-12-JobScheduling",
-      "name": "Job Scheduling",
+      "name": "2012 - Job Scheduling",
```
andrewwangva, line 173: wait we name all ceoi problems like this

andrewwangva: ^ignore that lol

PR #1736, opened 2021-06-13
add support for hiding problem suggestion button

thecodingwizard:


HarryWangATX (approved):
🙏

HarryWangATX:
actually i think itd be good if u disable the button and add a hover toolkit that says module is complete or smth like that

andrewwangva:
> actually i think itd be good if u disable the button and add a hover toolkit that says module is complete or smth like that

wait it's only temporary so there's no need to add more.

HarryWangATX:
> > actually i think itd be good if u disable the button and add a hover toolkit that says module is complete or smth like that
> 
> wait it's only temporary so there's no need to add more.

true

PR #1752, opened 2021-06-16
remove some problems as suggested by darren, fix some compilation errors

bqi343:


thecodingwizard:
should we add the removed problems into the conclusion module for silver?

HarryWangATX (approved):

andrewwangva:
> should we add the removed problems into the conclusion module for silver? seems like a waste to just delete them...

If the problem isn't good at all or are from a lesser site then no. Otherwise, yes.

PR #1767, opened 2021-06-19
Create USACO open18-talent-show.mdx

ar492:
added detailed description of the common fakesolve


solutions/usaco-839.mdx
```diff
@@ -0,0 +1,48 @@
...
+using namespace std;
+
+const int inf=2e9;
+int n,w; double ans=0;
+
+signed main(){
+       ifstream cin("talent.in"); ofstream cout("talent.out");
+       cin>>n>>w;
+       pair<int, int> cows[n];
+       for(int i=0; i<n; i++) cin>>cows[i].w>>cows[i].t;
```
brad182, line 26: change to tabs not spaces, use the editor on the guide website for help with this

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,48 @@
+---
+id: usaco-839
+source: USACO Gold 2018 Open
+title: Talent Show
+author: Arpan Banerjee
+---
+
+## Incorrect Solution — Common Misconception/Fakesolve
```
brad182, line 8: just write knapsack solution or something the other words are unnecessary, also you should transfer over the official code and call it binary search solution or something

brad182 (changes requested):

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,56 @@
+---
+id: usaco-839
+source: USACO Gold 2018 Open
+title: Talent Show
+author: Arpan Banerjee
+---
+
+[Official Editorial](http://www.usaco.org/current/data/sol_talent_gold_open18.html)
+
+## Incorrect Solution — Common Misconception
```
dongliu0426, line 10: looks okay, but perhaps wrap this in a warning block?

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,56 @@
...
+Instead of using binary search as the official solution does, a knapsack can be done. An array of talents can be kept, where `talent[i]` is the minimum weight required to create a talent sum of i. To make sure each cow is used at most once and the order of the cows doesn't matter, the indices of `talent` are iterated though from greatest to least, and the `cows` are iterated though in the outer loop; the implementation should make this clear. This maximizes the ratio $t/w$ as desired. Then, the maximum ratio of $t_i / w_i$ can be easily found by going through `talent` and ignoring any $w_i$ that is less than $W$. 
+
+My $\mathcal{O}(250 \cdot 10^3 \cdot n)$ code is below:
```
dongliu0426, line 13: err something like O(MAXT \cdot N) would be better (mention that MAXT is max talent as well)

dongliu0426 (changes requested):

solutions/usaco-839.mdx
dongliu0426: https://usaco.guide/general/working-mdx/#warning-block

```
<Warning title="Common Misconception">
put what you have here
</Warning>
```

dongliu0426 (approved):

brad182 (approved):

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+id: usaco-839
+source: USACO Gold 2018 Open
+title: Talent Show
+author: Arpan Banerjee
+---
+
+[Official Editorial](http://www.usaco.org/current/data/sol_talent_gold_open18.html)
+
+<Warning title="Common Misconception">
+Instead of using binary search as the official solution does, a knapsack can be done. An array of talents can be kept, where `talent[i]` is the minimum weight required to create a talent sum of i. To make sure each cow is used at most once and the order of the cows doesn't matter, the indices of `talent` are iterated though from greatest to least, and the `cows` are iterated though in the outer loop; the implementation should make this clear. This maximizes the ratio $t/w$ as desired. Then, the maximum ratio of $t_i / w_i$ can be easily found by going through `talent` and ignoring any $w_i$ that is less than $W$.
```
jessechoe10, line 11:
```suggestion
Rather than using binary search, we can also use knapsack DP. We can initialize an array called 'talent', where `talent[i]` is the minimum weight required to create a talent sum of i. We can iterate 'talent' from greatest to least, and iterate the "cows" through the outer loop. This maximizes the ratio $t/w$ as desired. The maximum ratio of $t_i / w_i$ can be easily found by iterating through `talent` and ignoring any $w_i$ that is less than $W$. 
```

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+                            talent[i+cow.t]=min(talent[i+cow.t], talent[i]+cow.w);
+       for(double t=1; t<talent.size(); t++)
+              if(talent[t] >= w && t / talent[t] > ans) ans = t / talent[t];
+       cout<<(int)(1000*ans)<<endl;
+}
+```
+
+</CPPSection>
+</LanguageSection>
+
```
jessechoe10, line 45:
```suggestion
<Warning>
```

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+       vector<int> talent(1e6+10, inf); talent[0]=0;
+       for(auto cow:cows)
+              for(int i=250*1e3; i>=0; i--) // 250 * 10^3 is total talent
+                     if(talent[i]!=inf)
+                            talent[i+cow.t]=min(talent[i+cow.t], talent[i]+cow.w);
+       for(double t=1; t<talent.size(); t++)
+              if(talent[t] >= w && t / talent[t] > ans) ans = t / talent[t];
+       cout<<(int)(1000*ans)<<endl;
+}
+```
```
jessechoe10, line 41: Use tabs instead of spaces. Also, you should add whitespace and curly braces in your loops and conditionals.

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+#define w first
+#define t second
+using namespace std;
+
+const int inf=2e9;
+int n,w; double ans=0;
+
+signed main(){
+       ifstream cin("talent.in"); ofstream cout("talent.out");
+       cin>>n>>w;
```
jessechoe10, line 29:
```suggestion
       cin >> n >> w;
```

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+       pair<int, int> cows[n];
```
jessechoe10, line 30: [Read here about why you should do the following](https://usaco.guide/general/adding-solution?lang=cpp)

You should change it to:

```suggestion
       vector<pair<int, int>> cows(n)
```

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+       for(int i=0; i<n; i++) cin>>cows[i].w>>cows[i].t;
+       vector<int> talent(1e6+10, inf); talent[0]=0;
+       for(auto cow:cows)
+              for(int i=250*1e3; i>=0; i--) // 250 * 10^3 is total talent
```
jessechoe10, line 34: Perhaps store `250 * 1e3` in a variable...

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+       for(double t=1; t<talent.size(); t++)
+              if(talent[t] >= w && t / talent[t] > ans) ans = t / talent[t];
+       cout<<(int)(1000*ans)<<endl;
+}
+```
+
+</CPPSection>
+</LanguageSection>
+
+However, even though this passes all of the test cases on USACO, there is a fatal flaw to this. We are finding the minimum weight required for each talent sum. What if the minimum weight for a talent sum is less than the weight threshold? Consider if the threshold is indeed attainable with a particular talent sum but the DP concluded it is not. The DP never considers a slightly higher weight, one that can pass the threshold, as a candidate for the answer. This means the DP fails on cases where the aforementioned scenario occurs and is optimal; one example is as follows:
```
jessechoe10, line 46:
```suggestion
However, even though this passes all of the test cases on USACO, there is an edge case with this knapsack solution. What if the minimum weight for a talent sum is less than the provided weight threshold? Consider the scenario when the threshold is indeed attainable with a particular talent sum, but the DP concluded it is not. The DP never considers a slightly higher weight, one that can pass the threshold, as a candidate for the answer. This means the DP fails on cases where the aforementioned scenario occurs and is optimal; one example is as follows:
```

jessechoe10 (changes requested):

solutions/usaco-839.mdx
dongliu0426: >  Also, you should add whitespace and curly braces in your loops and conditionals.

this is personal preference

content/4_Gold/Knapsack_DP.problems.json
```diff
@@ -123,7 +123,7 @@
       "isStarred": false,
       "tags": ["DP", "Knapsack", "Binary Search"],
       "solutionMetadata": {
-        "kind": "USACO",
+        "kind": "internal",
         "usacoId": "839"
```
maggie-j-liu, line 127:
```suggestion
        "kind": "internal"
```

solutions/usaco-839.mdx
ar492: This is effectively just rewording what I wrote. It does not add anything; in fact, it detracts from the original meaning slightly, in my opinion.

maggie-j-liu, line 11:
```suggestion
Instead of using binary search as the official solution does, a knapsack can be done. An array of talents can be kept, where $\texttt{talent}[i]$ is the minimum weight required to create a talent sum of $i$. To make sure each cow is used at most once and the order of the cows doesn't matter, the indices of $\texttt{talent}$ are iterated though from greatest to least, and the $\texttt{cows}$ are iterated though in the outer loop; the implementation should make this clear. This maximizes the ratio $\frac{t}{w}$ as desired. Then, the maximum ratio of $\frac{t_i}{w_i}$ can be easily found by going through $\texttt{talent}$ and ignoring any $w_i$ that is less than $W$. 
```

maggie-j-liu (changes requested):

solutions/usaco-839.mdx
ar492: done

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+#define t second
+using namespace std;
+
+const int inf=2e9;
+const int total_talent=250000;
+int n,w; double ans=0;
+
+signed main(){
+       ifstream cin("talent.in"); ofstream cout("talent.out");
+       cin >> n >> w;
```
maggie-j-liu, line 30: the spacing is inconsistent; either add spaces around all operators or don't put any spaces (most don't have spaces but this line and line 39 are different)

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+id: usaco-839
+source: USACO Gold 2018 Open
+title: Talent Show
+author: Arpan Banerjee
+---
+
+[Official Editorial](http://www.usaco.org/current/data/sol_talent_gold_open18.html)
+
+<Warning title="Common Misconception">
+Instead of using binary search as the official solution does, a knapsack can be done. An array of talents can be kept, where $\texttt{talent}[i]$ is the minimum weight required to create a talent sum of $i$. To make sure each cow is used at most once and the order of the cows doesn't matter, the indices of $\texttt{talent}$ are iterated though from greatest to least, and the $\texttt{cows}$ are iterated though in the outer loop; the implementation should make this clear. This maximizes the ratio $\frac{t}{w}$ as desired. Then, the maximum ratio of $\frac{t_i}{w_i}$ can be easily found by going through $\texttt{talent}$ and ignoring any $w_i$ that is less than $W$.
```
maggie-j-liu, line 11: i know that it says below that this solution is wrong, but i think it might be better to add a note at the top too

maggie-j-liu (changes requested):

solutions/usaco-839.mdx
```diff
@@ -0,0 +1,57 @@
...
+       cin>>n>>w;
+       vector<pair<int, int>> cows(n)
+       for(int i=0; i<n; i++) cin>>cows[i].w>>cows[i].t;
+       vector<int> talent(1e6+10, inf); talent[0]=0;
+       for(auto cow:cows)
+              for(int i=total_talent; i>=0; i--)
+                     if(talent[i]!=inf)
+                            talent[i+cow.t]=min(talent[i+cow.t], talent[i]+cow.w);
+       for(double t=1; t<talent.size(); t++)
+              if(talent[t] >= w && t / talent[t] > ans) ans = t / talent[t];
```
maggie-j-liu, line 39:
```suggestion
              if(talent[t]>=w&&t/talent[t]>ans) ans=t/talent[t];
```
huh i think this looks worse but to be consistent

jessechoe10:
We are closing this PR since you're rejecting all of our reviews. Also, this code isn't necessarily the write solution for all inputs, and the problem never stated the constraints that would provide a WA.

maggie-j-liu:
I think the explanation is fine, but the code should be removed...

maggie-j-liu (approved):

sathu-beep (changes requested):
why would we add incorrect code 🤣

maggie-j-liu:
Oops, https://github.com/cpinitiative/usaco-guide/pull/1767#discussion_r654748968 was not addressed. @ar492 can you open a new pr to fix?

PR #1814, opened 2021-06-25
Remove CCO and DMOJ (Module Revamp)

brad182:
Removed from the problems in the module and moved to extra problems


content/extraProblems.json
```diff
@@ -472,6 +472,31 @@
         "kind": "internal"
       }
     },
+    {
```
maggie-j-liu, line 475: idt we have to keep this one (cube world) since it doesn't have an internal solution

maggie-j-liu (approved):

PR #1823, opened 2021-06-25
test ceoi fix

nwatx:


bqi343:
should this be in the silver conclusion module?

nwatx:
> should this be in the silver conclusion module?

I think it would be a good addition, so I moved it.

bqi343 (approved):

PR #1857, opened 2021-07-01
Portals C++ Solution

maggie-j-liu:


SansPapyrus683 (approved):
nothing too egregious

solutions/usaco-1138.mdx
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+int find(int a)
+{
+	if (a == parent[a])
+	{
+		return a;
+	}
+	return parent[a] = find(parent[a]);
```
SansPapyrus683, line 46:
```suggestion
	return a == parent[a] ? a : (parent[a] = find(parent[a]));
```

maggie-j-liu: ehh this looks concise but i'm pretty sure it decreases readability. what do other people think?

bqi343: I think it's fine as is. If you really want to be concise, you should use one array for parents and sizes like I do here:

https://github.com/bqi343/USACO/blob/4aa96cd195a770c3a7f8977441020036d84b4f24/Implementations/content/graphs%20(12)/DSU/DSU%20(7.6).h

SansPapyrus683: removes 4 lines, and i don't think any people in gold don't know about the ternary expression

maggie-j-liu: i still think the original is more readable...

PR #1865, opened 2021-07-02
Add Floodfill Warning

brad182:
This issue has happened to me so many times where making the matrix just a little bit bigger gets you the rest of the test cases. The module does not currently mention this.

### DIFF
--- a/content/3_Silver/Flood_Fill.mdx
+++ b/content/3_Silver/Flood_Fill.mdx
@@ -531,6 +531,12 @@ for r in range(row_num):
 </PySection>
 
 </LanguageSection>
+	
+<Warning>
+
+In some problems if you make your grid exactly N x M your code will output the wrong answer even though the rest of your algorithm is correct. This is because some areas that can be flooded can be blocked if there is an unvisitable element right at the edge of your matrix. The solution is to create your grid at least one row and one column larger and offsetting your points based on how you increased the size of the matrix.    	
+	
+</Warning>
 
 ## Example - Counting Rooms
 


caoash:
I've never had this happen to me. Aren't the borders for floodfill problems usually blocked, so making the grid larger would make cells that shouldn't be flooded be flooded?

Can you give an example of when this would happen?

maggie-j-liu:
would this https://usaco.guide/problems/usaco-596-build-gates/solution be a good example?

brad182:
> I've never had this happen to me. Aren't the borders for floodfill problems usually blocked, so making the grid larger would make cells that shouldn't be flooded be flooded?
> 
> Can you give an example of when this would happen?

this happened to me quite a few times before, where changing the matrix size just a little bit bigger but not changing the algorithm gets you the rest of the test cases

yes, doing this in many cases will get the wrong answer which is why it is only done on some.

one example is on perimeter calculation:

######
##AA##
######

this is with the outer ring

###AA#
######
######

this is without the outer ring, if you try to calculate the perimeter using floodfill it will be 2 under because the places above the A's are not visited

a problem that I did that adding the outer ring fixed the test cases is this: http://www.usaco.org/index.php?page=viewproblem2&cpid=124

you do a floodfill inside a bfs, if you make the matrix perfectly N*N it will get the last test case wrong because the origin is actually reachable but the visiting is blocked off by bales on the edge of the matrix

caoash:
That seems very problem-specific and probably not worth a warning? I think it's just a specific case of constructing the graph differently for some problems (which is a very general concept).

brad182:
i'd say its quite common, maybe 30-40% of floodfill problems

caoash:
Maybe I just got unlucky when I was practicing but I've never done this before (it could have to do with my implementation?)

For example, in the perimeter problem, you can probably just count the number of A's and then use that to compute the perimeter. For the tractor problem, can't you just run dijkstras? I don't actually see how floodfill is that useful.

brad182:
> Maybe I just got unlucky when I was practicing but I've never done this before (it could have to do with my implementation?)
> 
> For example, in the perimeter problem, you can probably just count the number of A's and then use that to compute the perimeter. For the tractor problem, can't you just run dijkstras? I don't actually see how floodfill is that useful.
> 
> (Edit: I can see how it's useful for the tractor problem now, but I think that's just because you assume the grid is the size of the min/max x/y coordinates of the haybales. If the grid is unbounded I can see how this is useful, but I do think it's too specific and in most problems the edge of the grid isn't passable.)

it happens quite often to me though, hmm ill request review from others and see what they think

maggie-j-liu:
idk, i don't think i ever forgot to make the grid larger, but there are problems that need the grid to be larger so you can floodfill outside too (i think this one https://usaco.guide/problems/usaco-596-build-gates/solution ?)

SansPapyrus683:
i've never had this problem before lmao

nathangong:
I've definitely run into cases where extending the grid makes the problem a lot easier, although there were also ways of solving it without doing so

nwatx:
never had this problem but I've had cases where extending made it easier

bqi343:
Agree with @caoash

PR #1906, opened 2021-07-10
Added actual editorial

timothygao8710:


solutions/usaco-838.mdx
```diff
@@ -2,11 +2,15 @@
 id: usaco-838
 source: USACO Gold 2018 Open
 title: Milking Order
-author: Qi Wang
+author: Timothy Gao and Qi Wang
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_milkorder_gold_open18.html)
 
+Let's draw a directed edge from each node $u$ to another node $v$, if we node that $u$ must come before $v$. Note that this can be done when we process the $M$ observations. Now, notice that an ordering is valid only if the resulting graph is a DAG.  Additionally, we can use a PriorityQueue(instead of a queue) with topological sort to generate the lexographically minimum topological ordering. It remains to find how many observations we can follow starting from the first observation, which we can binary search for.
```
SansPapyrus683, line 10: > Let's draw a directed edge from each node $u$ to another node $v$, if we node that $u$ must come before $v$.

what
> we can use a PriorityQueue(instead of a queue)

replace w/ `we can use a PriorityQueue (instead of a queue)`

SansPapyrus683: also write priority queue instead of PriorityQueue not everyone does java

SansPapyrus683 (changes requested):

solutions/usaco-838.mdx
```diff
@@ -2,11 +2,15 @@
 id: usaco-838
 source: USACO Gold 2018 Open
 title: Milking Order
-author: Qi Wang
+author: Timothy Gao and Qi Wang
```
maggie-j-liu, line 5:
```suggestion
author: Timothy Gao, Qi Wang
```

solutions/usaco-838.mdx
maggie-j-liu, line 10:
```suggestion
Let's draw a directed edge from each node $u$ to another node $v$, if we node that $u$ must come before $v$. Note that this can be done when we process the $M$ observations. Now, notice that an ordering is valid only if the resulting graph is a DAG. Additionally, we can use a PriorityQueue(instead of a queue) with topological sort to generate the lexicographically minimum topological ordering. It remains to find how many observations we can follow starting from the first observation, which we can binary search for.
```

maggie-j-liu (changes requested):

solutions/usaco-838.mdx
```diff
@@ -2,11 +2,15 @@
 id: usaco-838
 source: USACO Gold 2018 Open
 title: Milking Order
-author: Qi Wang
+author: Timothy Gao, Qi Wang
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_milkorder_gold_open18.html)
 
+Denote each cow as a node in a graph. Let's draw a directed edge from each node $u$ to another node $v$, if we node that $u$ must come before $v$. Note that this can be done when we process the $M$ observations. Now, notice that an ordering is valid only if the resulting graph is a DAG. Additionally, we can use a PriorityQueue(instead of a queue) with topological sort to generate the lexicographically minimum topological ordering. It remains to find how many observations we can follow starting from the first observation, which we can binary search for.
```
maggie-j-liu, line 10: maybe use priority queue instead of PriorityQueue since that's only java

dongliu0426:
```suggestion
Denote each cow as a node in a graph. Let's draw a directed edge from each node $u$ to another node $v$, if we node that $u$ must come before $v$. Note that this can be done when we process the $M$ observations. Now, notice that an ordering is valid only if the resulting graph is a DAG. Additionally, we can use a priority queue (instead of a queue) with topological sort to generate the lexicographically minimum topological ordering. It remains to find how many observations we can follow starting from the first observation, which we can binary search for.
```

dongliu0426 (approved):
lgtm

dongliu0426:
@maggie-j-liu @SansPapyrus683 can you review?

SansPapyrus683 (approved):
i was never the best at reviewing actual content

maggie-j-liu (approved):

PR #1919, opened 2021-07-11
Add java solution to clock tree

nathangong:


solutions/usaco-1016.mdx
```diff
@@ -0,0 +1,180 @@
...
+			int b = io.nextInt() - 1;
+			adj.get(a).add(b);
+			adj.get(b).add(a);
+		}
+
+		dfs(0, 0, -1);
+
+		// same mod means you can start anywhere
+		if ((sum0 % 12) == (sum1 % 12)) {
+			io.println(n);;
```
brad182, line 139: two semicolons
```suggestion
			io.println(n);
```

solutions/usaco-1016.mdx
```diff
@@ -0,0 +1,180 @@
...
+			io.println(nodes0);
+		}
+		// no way to get all clocks pointing to 12
+		else {
+			io.println(0);
+		}
+		io.close();
+	}
+
+	static void dfs(int i, int color, int par) {
```
brad182, line 156:
```suggestion
	static void dfs (int i, int color, int par) {
```

solutions/usaco-1016.mdx
```diff
@@ -0,0 +1,180 @@
...
+```
+
+</CPPSection>
+
+<JavaSection>
+
+```java
+public class ClockTree {
+	static int[] clocks;
+	static List<ArrayList<Integer>> adj;
```
brad182, line 115: probably could be an array of arraylists, but doesn't matter that much

solutions/usaco-1016.mdx
```diff
@@ -0,0 +1,180 @@
...
+	static int sum0 = 0, sum1 = 0, nodes0 = 0, nodes1 = 0;
+
+	public static void main(String[] args) throws IOException {
```
brad182, line 118:
```suggestion
	public static void main (String[] args) throws IOException {
```

brad182 (changes requested):
minor changes

solutions/usaco-1016.mdx
nathangong: Yeah I did this cause we're suposed to do polymorphism, but it also seems a bit weird so idk

solutions/usaco-1016.mdx
nathangong: I think it's convention to not leave a space in between the method name and parameters?
image

brad182: oh

brad182:
wait you changed it to arraylists of arraylists i was saying it was fine, just commented that I usually do array of arraylists instead

nathangong:
Yeah I think ArrayList of ArrayLists makes more sense and is less confusing, and unfortunately Java doesn't allow making a List of Lists

brad182 (approved):

HarryWangATX (changes requested):
time complexity

HarryWangATX:
> I tried to do List<List> but it didn't work, so the only other option would be List<ArrayList> but I think that may be a bit weird/confusing

wdym by doesn't work? that syntax is valid.

nathangong:
Java throws this error when I try to do
```java List<List<Integer>> adj = new ArrayList<ArrayList<Integer>>();```
image

nathangong:
Wait I think ```List<List<Integer>> = new ArrayList<>();``` might work

HarryWangATX (approved):
code looks good, havent looked at explanation yet, core needs to review it too.

nathangong:
The explanation was approved a while back, the reason why it looks like I added everything was because I changed the file name to match the new naming standards

maggie-j-liu (approved):

Issue #863, opened 2021-02-26
POI 2004 Spies


darren-yao:
module link:https://usaco.guide/solutions/poi-04-szp?lang=cpp
Explain the optimizations to get from the MLE code to the AC code
specifically, turning the 
`vi adj[mx]`
 into 
`int adj[mx]`

bqi343:
Can you do it? :((

caoash:
I can do it.

SansPapyrus683:
why did 3 prs related to this get closed what

bqi343:
Believe the first two weren't meant to be open, the last one didn't have a complete explanation so it was eventually closed due to staleness.

SansPapyrus683:
i looked at the solution
why is this problem in the tree dp module
if the solution doesn't involve tree dp

bqi343:
It's related to "Tree Matching" from the module

bqi343:
Alternative way to get past the ML: https://codeforces.com/blog/entry/67883

angelikakasia:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll review the solution and the memory optimization explanation and start working on a draft. Thank you!

PR #1934, opened 2021-07-14
update spies w/ explanation for memory

caoash:
from [863](https://github.com/cpinitiative/usaco-guide/issues/863)

### DIFF
--- a/solutions/poi-04-spies.mdx
+++ b/solutions/poi-04-spies.mdx
@@ -8,7 +8,7 @@ author: Benjamin Qi
 Same thing as "Tree Matching" but on a functional graph. We can use the same
 greedy strategy of repeatedly matching a leaf to an adjacent vertex.
 
-MLE:
+Here's an implementation that uses `vector` and `boolean` arrays, which recieves MLE:
 
 ```cpp
 #include <bits/stdc++.h>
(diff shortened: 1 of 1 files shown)


solutions/poi-04-spies.mdx
```diff
@@ -262,7 +262,10 @@ int main() {
 }
 ```
 
-OK:
+To avoid this, we can make a few memory optimizations.
+
+1. Since this is a functional graph, each node is directed towards one other node. Instead of `vector`, we can use an `int` array to store each node's successor as the adjacency list.
```
bqi343, line 267: ^ it's not clear what this means

bqi343 (changes requested):

solutions/poi-04-spies.mdx
darren-yao: suggest something like:

Since this is a functional graph, each node has exactly one edge leaving it. Therefore, instead of representing the graph as an adjacency list, we can represent it as an array where arr[i] stores f(i).

bqi343: that doesn't explain what `ST` and `EN` are for though

caoash: oh didn't notice that. are ST and EN meant for some cycle detection stuff?

bqi343:
explanation is inaccurate

PR #1940, opened 2021-07-15
Update cf-616D.mdx

Mohammad-Nour-Massri:
Adding a well-commented C++ solution

### DIFF
--- a/solutions/cf-616D.mdx
+++ b/solutions/cf-616D.mdx
@@ -2,20 +2,67 @@
 id: cf-616D
 source: CF
 title: Longest k-Good Segment
-author: Sofia Yang
+author: Sofia Yang, Mohammad Nour Massri
 ---
 
 [Official Editorial](https://codeforces.com/blog/entry/22712)
-
-<LanguageSection>
-
-<JavaSection>
-
-## Solution
+## Implementation
 
 **Time Complexity:** $\mathcal{O}(N)$
+<LanguageSection>
+<CPPSection>
+```cpp 
 
-## Java Implementation
+#include <bits/stdc++.h>
+using namespace std;
+int n,k;
+int a[500009];
+int frequency[1000009];//frequency of each number in the window
+int maxSegment,leftMaxSegment,rightMaxSegment;//max size of the segment and its ends
+
+int main(){
+    scanf("%d%d",&n,&k);//large input data use scanf/printf instead of cin/cout
+    for(int i=0;i<n;i++)
+        scanf("%d",a+i);
+
+    int r=-1;//the right end of the current window
+    int cntDifVal = 0;//the number of different values in the window
+
+    //if frequency[val] == 0 then adding val to the window will result in
+    //adding a new different element which increases the number of different
+    //values by 1
+
+    for(int l=0;l<n;l++){//l is the left end of the window
+
+        //we can expand the window [l,r] to [l,r+1] if these 2 conditions hold:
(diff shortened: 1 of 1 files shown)


solutions/cf-616D.mdx
```diff
@@ -2,20 +2,67 @@
...
 <LanguageSection>
+<CPPSection>
+```cpp 
 
-<JavaSection>
+#include <bits/stdc++.h>
+using namespace std;
+int n,k;
+int a[500009];
+int frequency[1000009];//frequency of each number in the window
```
Qw3rTy827, line 20: I think it would be better to use a constant variable instead of hard-coding the array size. More aesthetic and less prone to potential errors.

solutions/cf-616D.mdx
```diff
@@ -2,20 +2,67 @@
...
+    for(int i=0;i<n;i++)
+        scanf("%d",a+i);
 
-**Time Complexity:** $\mathcal{O}(N)$
+    int r=-1;//the right end of the current window
+    int cntDifVal = 0;//the number of different values in the window
+
+    //if frequency[val] == 0 then adding val to the window will result in
+    //adding a new different element which increases the number of different
+    //values by 1
```
Qw3rTy827, line 33: Try to use ```/* */``` to add large blocks of code instead of ```//``` on every individual line.

solutions/cf-616D.mdx
```diff
@@ -2,20 +2,67 @@
...
+
+    for(int l=0;l<n;l++){//l is the left end of the window
+
+        //we can expand the window [l,r] to [l,r+1] if these 2 conditions hold:
+        //1. r < n-1 making sure we don't go beyond the array's boundaries
+        //2. adding the value a[r+1] to the window does not increase the
+        //number of different values in the window to be larger than k
```
Qw3rTy827, line 40: Just a suggestion but I think it would be better to move the comments about the thought process in front of the code and form it into a proper editorial as it is easy to read the actual code that way. You should keep comments explaining individual lines and the functions of variables of course.

solutions/cf-616D.mdx
```diff
@@ -2,20 +2,67 @@
...
+            maxSegment = r-l+1;
+            leftMaxSegment = l;
+            rightMaxSegment = r;
+        }
+
+        //here we discard the first element a[l] making the current segment [l+1,r]
+        frequency[a[l]]--;//update the frequency
+        cntDifVal -= frequency[a[l]] == 0;//update the number of different values
+    }
+    // we add one because the answer should be 1-indexed
```
Qw3rTy827, line 59: Having either a space or nothing after ```//``` are both ok, but try to keep the comment style consistent.

solutions/cf-616D.mdx
```diff
@@ -2,20 +2,67 @@
...
 source: CF
 title: Longest k-Good Segment
-author: Sofia Yang
+author: Sofia Yang, Mohammad Nour Massri
 ---
 
 [Official Editorial](https://codeforces.com/blog/entry/22712)
+## Implementation
 
+**Time Complexity:** $\mathcal{O}(N)$
```
Qw3rTy827, line 11: Put the time complexity before ```## Implementation```

Qw3rTy827 (changes requested):

darren-yao (approved):

dongliu0426 (changes requested):
use tabs for indentation

dongliu0426 (approved):

PR #1946, opened 2021-07-16
Improve urls

### LINKED ISSUE #1011: Improve URLs
The URLs for modules are not very descriptive. This may or may not be bad for SEO, but it definitely makes it harder to figure out what a link is when somebody sends you one, especially in places where there are no link previews, such as email or zoom chat.

It might be better to just slugify the module name (the old URLs should still redirect though).

Examples of questionable URLs:
https://usaco.guide/general/factors-choosing?lang=cpp
https://usaco.guide/general/generic?lang=cpp
https://usaco.guide/silver/ff?lang=cpp
https://usaco.guide/gold/sp?lang=cpp
https://usaco.guide/plat/2DRQ?lang=cpp

bqi343:
hi everyone

- closes #1011 (though maybe some module names still need to be changed, particularly in Gold+)
- add some error handling and warnings in `gatsby-node.ts`


solutions/cf-919D.mdx
```diff
@@ -47,7 +47,7 @@ using pi = pair<int,int>;
 void setIO(string name = "") {
 	cin.tie(0)->sync_with_stdio(0); // see /general/fast-io
 	if (sz(name)) {
-		(void)!freopen((name+".in").c_str(), "r", stdin); // see /general/io
+		(void)!freopen((name+".in").c_str(), "r", stdin); // see /general/input-output
```
maggie-j-liu, line 50: maybe refactor all of these to use the codesnip instead? so they don't all need to be changed when the codesnip is updated

bqi343:
have not verified that these actually compile :/

SansPapyrus683 (approved):
maybe we should rename the modules and stuff as well idk but

PR #1949, opened 2021-07-16
Fixed images for Blocked Billboard II

ryanchou-dev:
It appeared like this before:
image

### DIFF
--- a/solutions/usaco-783.mdx
+++ b/solutions/usaco-783.mdx
@@ -128,6 +128,7 @@ public class Billboard {
 </JavaSection>
 
 <PySection>
+	
 ```py
 fin, fout = open("billboard.in"), open("billboard.out", "w")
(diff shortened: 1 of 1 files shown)


maggie-j-liu:
they still don't show up https://usaco-guide-duflij434-cpinitiative.vercel.app/problems/usaco-783-blocked-billboard-ii/solution ?
i think the path of the images before was correct, so it's probably a website problem
@thecodingwizard

ryanchou-dev:
> they still don't show up https://usaco-guide-duflij434-cpinitiative.vercel.app/problems/usaco-783-blocked-billboard-ii/solution ?
> i think the path of the images before was correct, so it's probably a website problem
> @thecodingwizard

oh, i guess it's just when i preview the file it shows :/

PR #1950, opened 2021-07-17
Java Solution for Hoof Paper Scissors (Gold Version of the Problem)

brad182:
I cannot make the lines any shorter

### DIFF
--- a/solutions/usaco-694.mdx
+++ b/solutions/usaco-694.mdx
@@ -2,7 +2,7 @@
 id: usaco-694
 source: USACO Gold 2017 January
 title: Hoof, Paper, Scissors
-author: Neo Wang (C++), Melody Yu (Video)
+author: Melody Yu (Video), Neo Wang (C++), Brad Ma (Java)
 ---
 
 [Official Analysis](http://www.usaco.org/current/data/sol_hps_gold_jan17.html)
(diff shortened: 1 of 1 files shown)


solutions/usaco-694.mdx
```diff
@@ -2,7 +2,7 @@
 id: usaco-694
 source: USACO Gold 2017 January
 title: Hoof, Paper, Scissors
-author: Neo Wang (C++), Melody Yu (Video)
+author: Melody Yu (Video), Neo Wang (C++), Brad Ma (Java)
```
maggie-j-liu, line 5: i think we can remove the (Video), (C++) and (Java)
```suggestion
author: Melody Yu, Neo Wang, Brad Ma
```

solutions/usaco-694.mdx
```diff
@@ -104,5 +104,84 @@ int main() {
...
+				// try switching to the three possibilities
+				int add1 = 0;
+				if (moves[currentGame] == 0) {
+					add1 = 1;
+				}
+				// take the maximum of the game before + add (win or lose)
+				// and switching from the other two options 
+				dp[currentGame][numberOfSwitches][0] = Math.max(dp[currentGame - 1][numberOfSwitches][0] 
+				+ add1, Math.max(dp[currentGame - 1][numberOfSwitches - 1][1] + add1, 
+				dp[currentGame - 1][numberOfSwitches - 1][2] + add1));
```
maggie-j-liu, line 156: this line splitting looks kind of weird, but these lines are kind of long so i'm not sure what the best way to format it is? anyone else have ideas?

maggie-j-liu (changes requested):

nathangong:
You can considerably reduce the line lengths by making your variable names shorter

solutions/usaco-694.mdx
```diff
@@ -104,5 +104,84 @@ int main() {
...
+	static int[] moves;
+
+	public static void main (String[] args) throws IOException {
+		BufferedReader br = new BufferedReader(new FileReader("hps.in"));
+		PrintWriter pw = new PrintWriter(new BufferedWriter(new FileWriter("hps.out")));
+
+		StringTokenizer inputLine = new StringTokenizer(br.readLine());
+		int numGames = Integer.parseInt(inputLine.nextToken());
+		int maximumSwitches = Integer.parseInt(inputLine.nextToken());
+		// dp[game][switches][type] = maximum number of wins
```
nathangong, line 123: Move this comment to where dp is declared

solutions/usaco-694.mdx
brad182: yeah it basically takes the maximum of the three matrix values in the dp so the line is that long. i could make the variable names shorter but then they won't be as descriptive and it will still be over the limit

brad182:
> You can considerably reduce the line lengths by making your variable names shorter

yes but they won't be as descriptive and won't follow the strict variable naming rules

solutions/usaco-694.mdx
brad182: maybe i could get an exemption for this one because it is a special case taking the max of three values in a dp?

maggie-j-liu: long lines are fine, but the way there are split onto multiple lines looks kind of weird?? not sure the best way to split them though...

brad182: if this was c++ you could probably just macro each dp position

nathangong:
Like for example you can use n and k instead of numGames and maximumSwitches (because that's given in the problem). Also, instead of using currentGame you can use i (which is pretty standard for dp). numberOfSwitches can just be numSwitches

HarryWangATX:
> Like for example you can use n and k instead of numGames and maximumSwitches (because that's given in the problem). Also, instead of using currentGame you can use i (which is pretty standard for dp). numberOfSwitches can just be numSwitches

optional change btw it's fine as long as it's readable

dongliu0426:
> > You can considerably reduce the line lengths by making your variable names shorter
> 
> 
> 
> yes but they won't be as descriptive and won't follow the strict variable naming rules and it will still be over

I think it's ok to do that as long as you clarify what the variables mean in a comment.

brad182:
ok now it is under the limit with one letter variable names

solutions/usaco-694.mdx
```diff
@@ -104,5 +104,90 @@ int main() {
...
+				+ add1, Math.max(dp[c - 1][n - 1][1] + add1,
+				dp[c - 1][n - 1][2] + add1));
+
+				int add2 = 0;
+				if (moves[c] == 1) {
+					add2 = 1;
+				}
+				dp[c][n][1] = Math.max(dp[c - 1][n][1]
+				+ add2, Math.max(dp[c - 1][n - 1][0] + add2,
+				dp[c - 1][n - 1][2] + add2));
```
maggie-j-liu, line 170:
```suggestion
				dp[c][n][1] = Math.max(
								dp[c - 1][n][1] + add2, 
								Math.max(
									dp[c - 1][n - 1][0] + add2,
									dp[c - 1][n - 1][2] + add2
								)
							);
```
i'm not sure; does this look better?

brad182: i don't think so, that looks kind of weird

dongliu0426 (changes requested):

solutions/usaco-694.mdx
```diff
@@ -104,5 +104,90 @@ int main() {
...
+		c = current game number
+		n = number of switches
+		 */
+
+		for (int c = 1; c <= numGames; c++) {  // for each game
+			// for each switch amount
+			for (int n = 1; n < maximumSwitches + 2; n++) {
+				// try switching to the three possibilities
+				int add1 = 0;
+				if (moves[n] == 0) {
```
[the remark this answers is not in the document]
brad182: yep, my code has that idk why it got changed here

dongliu0426 (approved):
yay

PR #1951, opened 2021-07-17
Problem-Search-UI

AAkashyap1:


nathangong:
I feel like the blue banner might be a bit too big, because I can barely see the problems without having to scroll
image

IMO you could probably just remove the banner and move the search bar and heading next to the filters.

maggie-j-liu (approved):

SansPapyrus683:
what makes you think i know anything about web development

AAkashyap1:
> what makes you think i know anything about web development

i mean like you can still look at the actual content

AAkashyap1:
> > what makes you think i know anything about web development
> 
> i mean like you can still look at the actual content

not necessarily the code

SansPapyrus683 (approved):
i'll just nod along with maggie

maggie-j-liu:
😆 please at least go to the website, click some buttons, and make sure nothing's broken

nathangong:
Looks good, my main comment is would it be possible to remove the inner scrollbar so there isn't a double scrollbar? Again, I'm not super experienced in web dev so take my comments with a grain of salt.

AAkashyap1:
> Looks good, my main comment is would it be possible to remove the inner scrollbar so there isn't a double scrollbar? Again, I'm not super experienced in web dev so take my comments with a grain of salt.

yeah im working to fix this, but i dont know how lol :(

nathangong:
I think I figured out why it's happening. Try removing the h-screen class from this div:

image

nathangong:
After that it looks fine
image

src/pages/problems.tsx
```diff
@@ -27,30 +37,70 @@ export default function ProblemsPage(props: PageProps) {
...
-              transformItems={items =>
-                items.sort((x, y) => x.label.localeCompare(y.label))
-              }
-            />
           </div>
-          <div className="flex justify-center mt-4">
+          <div className="flex mt-4 mb-1 mx-9 justify-center" >
             <PoweredBy />
+          </div> 
+          <div className="h-screen pt-3 px-9 pb-4 grid grid-cols-10">
```
nathangong, line 55:
```suggestion
          <div className="pt-3 px-9 pb-4 grid grid-cols-10">
```

AAkashyap1:
> After that it looks fine
> image

the only problem with this is that the tags become way too long and longer than results

AAkashyap1:
> > After that it looks fine
> > image
> 
> the only problem with this is that the tags become way too long and longer than results

which is fine if everyone agrees

PR #1966, opened 2021-07-21
Add a CPP solution to the problem "Great Revegetation" in the Silver DFS module

Qw3rTy827:


solutions/usaco-920.mdx
```diff
@@ -2,13 +2,85 @@
...
+        cout << 1;
+        for(int i = 0; i < ans; i++){cout << 0;}
+        cout << "\n";
+    }
+    return 0;
+}
+```
+
+</CPPSection>
+
```
jessechoe10, line 83: How is this any different from the official editorial?

jessechoe10: I looked at the official solution, and it's basically the same logic / implementation.

jessechoe10:
I'm going to close this PR, but we can reopen if someone states otherwise.

solutions/usaco-920.mdx
Qw3rTy827: I added this solely because there is only an java solution and not a CPP solution when I opened the internal solution. You can close this if you think that a CPP solution is unnecessary.

PR #1972, opened 2021-07-21
added video solution for p838

cskitty:


Qw3rTy827 (approved):
Clear explanation in the video ! Though the use of binary search could be explained a bit better. Rather than saying that a lower X has a higher probability of working, I think you should say that X in this problem possesses an unique property where if an integer K works, then all integers <= K will work (monotonicity). This syncs better with the base property that allows binary search to produce the correct results.

ryanchou-dev (approved):
👍 clean code and clear diagrams in the video :)

PR #1989, opened 2021-07-25
add workshop banner

thecodingwizard:


dongliu0426:
yay more ads

SansPapyrus683:
~~maybe we should actually start monetizing the site~~

PR #1997, opened 2021-07-27
Update SearchBox.tsx

### LINKED ISSUE #1994: Automatically set input field in problem page
when i go on most search engines, i can usually just type in the url and get right to typing my query
that isn't the case in the problem page- i actually have to click on the field and then start typing
a bit annoying tbh

bqi343:
Closes #1994


SansPapyrus683 (approved):
perfect

PR #2002, opened 2021-07-28
Update Status.tsx

AAkashyap1:


src/components/ProblemsPage/Status.tsx
```diff
@@ -4,10 +4,12 @@ import UserDataContext from '../../context/UserDataContext/UserDataContext';
 import { useDarkMode } from '../../context/DarkModeContext';
 
 const options = [
+  { label: 'Not Attempted', value: 'Not Attempted' },
```
maggie-j-liu, line 7: if the labels and values are all the same maybe import from here https://github.com/cpinitiative/usaco-guide/blob/master/src/models/problem.ts#L806?

nathangong:
Could you also fix the problem where the background abruptly ends if there aren't that any problems that show up?

image

src/components/ProblemsPage/Status.tsx
AAkashyap1: Good idea but i think react-select has some weird requirements where all items need to have a label (for option display) and a value

AAkashyap1: idk how to combat that

AAkashyap1:
> Could you also fix the problem where the background abruptly ends if there aren't that many problems that show up?
> 
> image

i think i fixed this - can you check it when the deployment finishes?

nathangong:
Yep, it looks good

PR #2017, opened 2021-07-31
Add resource to bitmask dp module

ar492:


darren-yao (approved):

andrewwangva:
@darren-yao why merge? Idt the resource introduces anything new and isn't very thorough. Also note small changes like these don't warrant an author addition.

ar492:
@andrewwangva the resource is the only one that gives a diagram to more clearly explain bitmask DP (maybe I should put this in the description of the resource). Regarding your second point, I am also not sure about the guidelines for being contributor vs author, but I added myself as contributor since I saw nwatx's blog as a resource and he is listed as contributor (not sure if he wrote any part of the module though).

nwatx:
I wrote that specifically for the guide - cross compatible mdx and review process.

ar492:
I'll make a new pr without myself as contributor then.

PR #2020, opened 2021-08-01
Add resource to resources section of bitmask DP + resources page

ar492:


caoash:
Looks like a good resource.

PR #2022, opened 2021-08-01
Revert "Add resource to resources section of bitmask DP + resources page"

caoash:
Reverts cpinitiative/usaco-guide#2020


caoash:
@ar492 did you remove yourself as contributor as said in #2017? i don't see which commit did this, looks like you're still author on bitmask dp/resources for cp, we tend to not add people as author for being resource authors

PR #2050, opened 2021-08-05
Update Intro_Sets.mdx

sosudo:


SansPapyrus683:
nice first pr
sadly i don't really think this is needed(?)

maggie-j-liu:
i agree, it basically gets rid of the `distinct_nums` variable, but i think the original code is short enough and more readable

Issue #2060, opened 2021-08-06
clicking link on table of contents of module removes lang query param


thecodingwizard:
see title

bqi343:
Actually, the above branch seems fine to merge?

maggie-j-liu:
fixed in #2815

PR #2061, opened 2021-08-07
Add lang query param for table of content links

nathangong:
This should address #2060, let me know if there are any issues


maggie-j-liu:
seems fine, but i noticed another problem (idk might be related) where if you're on a certain module and you click on the module name again, the lang param also disappears
so if i'm on the two pointers module and i click two pointers again, the lang disappears
image

nathangong:
hm interesting

nathangong:
Should I just add the language parameter for the side nav links?

nathangong:
Rip this PR also causes some unintended behaviors (copying-pasting the TOC link will include the language param), so not sure if it's worth it or not

nathangong:
Closing this for now -- I think there probably exists a better solution that can solve both problems at once

PR #2069, opened 2021-08-08
Update cses-1640.mdx

MinorWannaBeDev:
Create map Vals with size n instead of the default, otherwise, it gives TLE due to resizing operations.

### DIFF
--- a/solutions/cses-1640.mdx
+++ b/solutions/cses-1640.mdx
@@ -101,6 +101,7 @@ void setIO(string name = "") { // name is nonempty for USACO file I/O
 
 int main() {
 	int n, x; cin >> n >> x;
+	//initialize length of dynamic containers to prevent TLE down the road
 	vi a(n);
 	unordered_map<int, int> vals(n);
 	for (int i = 0; i < n; i++) {
--- a/solutions/cses-1640.mdx
+++ b/solutions/cses-1640.mdx
@@ -102,7 +102,7 @@ void setIO(string name = "") { // name is nonempty for USACO file I/O
 int main() {
 	int n, x; cin >> n >> x;
 	vi a(n);
-	map<int, int> vals(n);
+	unordered_map<int, int> vals(n);
 	for (int i = 0; i < n; i++) {
 		cin >> a[i];
 	}
--- a/solutions/cses-1640.mdx
+++ b/solutions/cses-1640.mdx
@@ -101,7 +101,7 @@ void setIO(string name = "") { // name is nonempty for USACO file I/O
 
 int main() {
 	int n, x; cin >> n >> x;
-	//initialize length of dynamic containers to prevent TLE down the road
+	// initialize length of dynamic containers to prevent TLE down the road
 	vi a(n);
 	unordered_map<int, int> vals(n);
 	for (int i = 0; i < n; i++) {
--- a/solutions/cses-1640.mdx
+++ b/solutions/cses-1640.mdx
@@ -102,7 +102,7 @@ void setIO(string name = "") { // name is nonempty for USACO file I/O
 int main() {
 	int n, x; cin >> n >> x;
 	vi a(n);
-	map<int, int> vals;
+	map<int, int> vals(n);
 	for (int i = 0; i < n; i++) {
 		cin >> a[i];
 	}


ryanchou-dev (approved):
👍

SansPapyrus683:
maybe add a comment explaining this

dongliu0426 (changes requested):
Could you change this to an `unordered_map`?

MinorWannaBeDev:
Even the unordered map version gets TLE, as I have already stated, it gets TLE due to the fact that it needs to resize the map for very large inputs, which means copying all elements, changing map implementation wouldn't fix the issue.

solutions/cses-1640.mdx
```diff
@@ -101,6 +101,7 @@ void setIO(string name = "") { // name is nonempty for USACO file I/O
 
 int main() {
 	int n, x; cin >> n >> x;
+	//initialize length of dynamic containers to prevent TLE down the road
```
SansPapyrus683, line 104:
```suggestion
	// initialize length of dynamic containers to prevent TLE down the road
```

HarryWangATX (approved):

nwatx:
wait, this didn't TLE before? https://cses.fi/problemset/result/2658248/

@HarryWangATX @MinorWannaBeDev

MinorWannaBeDev:
@nwatx it consistently does TLE for me, maybe something changed on their side?

Anyways, specifying container size at initialization if possible is in my opinion much better practice.

SansPapyrus683:
is it? could i have a SO link or style guide pointing to that being a best practice?

nwatx:
> @nwatx it consistently does TLE for me, maybe something changed on their side?
> 
> Anyways, specifying container size at initialization if possible is in my opinion much better practice.

@MinorWannaBeDev can you link your submission on CSES?

MinorWannaBeDev:
https://stackoverflow.com/questions/25108854/initializing-the-size-of-a-c-vector/25108894

https://www.geeksforgeeks.org/using-stdvectorreserve-whenever-possible/

I think these links explain it quite well.
And these apply even more for CP where O(n) memory reallocation can drastically influence the timing.
https://cses.fi/problemset/result/2657464/
https://cses.fi/problemset/result/2658737/
https://cses.fi/problemset/result/2658740/
these are some submissions with the same code from the internal solution, always TLE on the 11th test.
@nwatx

nwatx:
@MinorWannaBeDev for some reason the cses.fi results don't load for me. Could you share them through another source? Curious.

SansPapyrus683:
Those links are for vectors, not for maps.

MinorWannaBeDev:
@SansPapyrus683 
I was dead wrong about it, sorry for confounding maps and vectors, I didn't realize that maps weren't contiguous in memory.

But how does it explain that this single change makes it go from TLE to AC?
Is there something going over my head?
image
image
A nearly 10 times decrease in time!
@SansPapyrus683

SansPapyrus683:
no clue lmao

MinorWannaBeDev:
No clue either.
Some things are better left unanswered.

nwatx:
> @SansPapyrus683 
> I was dead wrong about it, sorry for confounding maps and vectors, I didn't realize that maps weren't contiguous in memory.
> 
> But how does it explain that this single change makes it go from TLE to AC?
> Is there something going over my head?
> image
> image
> A nearly 10 times decrease in time!
> @SansPapyrus683 

Er I meant the code. Also, presizing unordered maps also saves time so they dont have to resize themselves.

MinorWannaBeDev:
### Good code:
```cpp
#include <bits/stdc++.h>
using namespace std;
using ll = long long;
using vi = vector<int>;
#define pb push_back
#define rsz resize
#define all(x) begin(x), end(x)
#define sz(x) (int)(x).size()
using pi = pair<int,int>;
#define f first
#define s second
#define mp make_pair
void setIO(string name = "") { // name is nonempty for USACO file I/O
	ios_base::sync_with_stdio(0); cin.tie(0); // see Fast Input & Output
	if(sz(name)){
		freopen((name+".in").c_str(), "r", stdin); // see Input & Output
		freopen((name+".out").c_str(), "w", stdout);
	}
}
 
int main() {
	int n, x; cin >> n >> x;
	vi a(n);
	unordered_map<int, int> vals(n);
	for (int i = 0; i < n; i++) {
		cin >> a[i];
	}
	for (int i = 0; i < n; i++) {
		if(vals.count(x - a[i])){
			cout << i + 1 << " " << vals[x - a[i]] << "\n";
			return 0;
		}
		vals[a[i]] = i + 1;
	}
	cout << "IMPOSSIBLE" << '\n';
}
```
### Bad code:
```cpp
#include <bits/stdc++.h>
using namespace std;
using ll = long long;
using vi = vector<int>;
#define pb push_back
#define rsz resize
#define all(x) begin(x), end(x)
#define sz(x) (int)(x).size()
using pi = pair<int,int>;
#define f first
#define s second
#define mp make_pair
void setIO(string name = "") { // name is nonempty for USACO file I/O
	ios_base::sync_with_stdio(0); cin.tie(0); // see Fast Input & Output
	if(sz(name)){
		freopen((name+".in").c_str(), "r", stdin); // see Input & Output
		freopen((name+".out").c_str(), "w", stdout);
	}
}
 
int main() {
	int n, x; cin >> n >> x;
	vi a(n);
	unordered_map<int, int> vals;
	for (int i = 0; i < n; i++) {
		cin >> a[i];
	}
	for (int i = 0; i < n; i++) {
		if(vals.count(x - a[i])){
			cout << i + 1 << " " << vals[x - a[i]] << "\n";
			return 0;
		}
		vals[a[i]] = i + 1;
	}
	cout << "IMPOSSIBLE" << '\n';
}
```

That's the only difference, I have no idea why it's acting this way.

nwatx:
@MinorWannaBeDev you don't include the use of setIO(); hence your io routine is inherently slow.

nwatx:
also, that test presumably uses I think an unordered map killer which blows it up see: https://codeforces.com/blog/entry/62393

MinorWannaBeDev:
> @MinorWannaBeDev you don't include the use of setIO(); hence your io routine is inherently slow.

That's not my code, it's the internal solution.
However, even after optimizing the IO it still gives TLE on the 11th test.
```cpp
#include <bits/stdc++.h>
using namespace std;
using ll = long long;
using vi = vector<int>;
#define pb push_back
#define rsz resize
#define all(x) begin(x), end(x)
#define sz(x) (int)(x).size()
using pi = pair<int,int>;
#define f first
#define s second
#define mp make_pair
void setIO(string name = "") { // name is nonempty for USACO file I/O
	ios_base::sync_with_stdio(0); cin.tie(0); // see Fast Input & Output
	if(sz(name)){
		freopen((name+".in").c_str(), "r", stdin); // see Input & Output
		freopen((name+".out").c_str(), "w", stdout);
	}
}
 
int main() {
    setIO("");
	int n, x; cin >> n >> x;
	vi a(n);
	unordered_map<int, int> vals;
	for (int i = 0; i < n; i++) {
		cin >> a[i];
	}
	for (int i = 0; i < n; i++) {
		if(vals.count(x - a[i])){
			cout << i + 1 << " " << vals[x - a[i]] << "\n";
			return 0;
		}
		vals[a[i]] = i + 1;
	}
	cout << "IMPOSSIBLE" << '\n';
}
```
I don't think that IO is the problem.
@nwatx

nwatx:
@MinorWannaBeDev see comment above; unordered map killer

MinorWannaBeDev:
> also, that test presumably uses I think an unordered map killer which blows it up to see: https://codeforces.com/blog/entry/62393

This can't be true, look at the "good code" and the "bad code".
Both of them use unordered_map, the only difference is (n).

nwatx:
> > also, that test presumably uses I think an unordered map killer which blows it up to see: https://codeforces.com/blog/entry/62393
> 
> This can't be true, look at the "good code" and the "bad code".
> Both of them use unordered_map, the only difference is (n).
> @nwatx

the following code passes (using custom hash)

```cpp
#include <bits/stdc++.h>
using namespace std;
using ll = long long;
using vi = vector<int>;
#define pb push_back
#define rsz resize
#define all(x) begin(x), end(x)
#define sz(x) (int)(x).size()
using pi = pair<int,int>;
#define f first
#define s second
#define mp make_pair
void setIO(string name = "") { // name is nonempty for USACO file I/O
	ios_base::sync_with_stdio(0); cin.tie(0); // see Fast Input & Output
	if(sz(name)){
		freopen((name+".in").c_str(), "r", stdin); // see Input & Output
		freopen((name+".out").c_str(), "w", stdout);
	}
}

/**
 * Description: Hash map with the same API as unordered\_map, but \tilde 3x faster.
	 * Initial capacity must be a power of 2 if provided.
 * Source: KACTL
 * Usage: ht<int,int> h({},{},{},{},{1<<16});
 */

const long double PI = acos((long double)-1);

#include <ext/pb_ds/assoc_container.hpp>
using namespace __gnu_pbds;
struct chash { /// use most bits rather than just the lowest ones
	const uint64_t C = ll(2e18*PI)+71; // large odd number
	const int RANDOM = 69;
	ll operator()(ll x) const { /// https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
		return __builtin_bswap64((x^RANDOM)*C); }
};
template<class K,class V> using um = unordered_map<K,V,chash>;
template<class K,class V> using ht = gp_hash_table<K,V,chash>;
template<class K,class V> V get(ht<K,V>& u, K x) {
	auto it = u.find(x); return it == end(u) ? 0 : it->s; }
 
int main() {
	int n, x; cin >> n >> x;
	vi a(n);
	um<int, int> vals;
	for (int i = 0; i < n; i++) {
		cin >> a[i];
	}
	for (int i = 0; i < n; i++) {
		if(vals.count(x - a[i])){
			cout << i + 1 << " " << vals[x - a[i]] << "\n";
			return 0;
		}
		vals[a[i]] = i + 1;
	}
	cout << "IMPOSSIBLE" << '\n';
}
```

MinorWannaBeDev:
> > > also, that test presumably uses I think an unordered map killer which blows it up to see: https://codeforces.com/blog/entry/62393
> > 
> > 
> > This can't be true, look at the "good code" and the "bad code".
> > Both of them use unordered_map, the only difference is (n).
> > @nwatx
> 
> the following code passes (using custom hash)
> 
> ```c++
> #include <bits/stdc++.h>
> using namespace std;
> using ll = long long;
> using vi = vector<int>;
> #define pb push_back
> #define rsz resize
> #define all(x) begin(x), end(x)
> #define sz(x) (int)(x).size()
> using pi = pair<int,int>;
> #define f first
> #define s second
> #define mp make_pair
> void setIO(string name = "") { // name is nonempty for USACO file I/O
> 	ios_base::sync_with_stdio(0); cin.tie(0); // see Fast Input & Output
> 	if(sz(name)){
> 		freopen((name+".in").c_str(), "r", stdin); // see Input & Output
> 		freopen((name+".out").c_str(), "w", stdout);
> 	}
> }
> 
> /**
>  * Description: Hash map with the same API as unordered\_map, but \tilde 3x faster.
> 	 * Initial capacity must be a power of 2 if provided.
>  * Source: KACTL
>  * Usage: ht<int,int> h({},{},{},{},{1<<16});
>  */
> 
> const long double PI = acos((long double)-1);
> 
> #include <ext/pb_ds/assoc_container.hpp>
> using namespace __gnu_pbds;
> struct chash { /// use most bits rather than just the lowest ones
> 	const uint64_t C = ll(2e18*PI)+71; // large odd number
> 	const int RANDOM = 69;
> 	ll operator()(ll x) const { /// https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
> 		return __builtin_bswap64((x^RANDOM)*C); }
> };
> template<class K,class V> using um = unordered_map<K,V,chash>;
> template<class K,class V> using ht = gp_hash_table<K,V,chash>;
> template<class K,class V> V get(ht<K,V>& u, K x) {
> 	auto it = u.find(x); return it == end(u) ? 0 : it->s; }
>  
> int main() {
> 	int n, x; cin >> n >> x;
> 	vi a(n);
> 	um<int, int> vals;
> 	for (int i = 0; i < n; i++) {
> 		cin >> a[i];
> 	}
> 	for (int i = 0; i < n; i++) {
> 		if(vals.count(x - a[i])){
> 			cout << i + 1 << " " << vals[x - a[i]] << "\n";
> 			return 0;
> 		}
> 		vals[a[i]] = i + 1;
> 	}
> 	cout << "IMPOSSIBLE" << '\n';
> }
> ```
@nwatx 
Your code finishes at about the same speed on test case 11 (0.15s vs 0.11s), but I still can't get why specifying (n) makes the code AC instead of TLE, and why would changing hashes makes the code faster.
As you can tell by the looks of it I am not a very experienced programmer.

nwatx:
@MinorWannaBeDev this is because the bucket size does not have to be resized upon adding items, which makes it noticeably faster - perhaps it changed something internally as well to avoid blowing up w/ unordered_map killers

HarryWangATX:
> @MinorWannaBeDev this is because the bucket size does not have to be resized upon adding items, which makes it noticeably faster - perhaps it changed something internally as well to avoid blowing up w/ unordered_map killers

I'm pretty sure it is just changing the bucket size. @MinorWannaBeDev you should make a new pr to re-explain the dynamic size in comments as it is not resizing length but rather bucket size.

PR #2076, opened 2021-08-10
Corrected comment and the internal solution to prevent TLE

MinorWannaBeDev:


SansPapyrus683:
wait uh can you put a space after the two `//`'s tho
like `// this is a comment`

nwatx:
doesn't prevent TLE: the solution already passes.

PR #2100, opened 2021-08-14
Add resource to divisibility module

ar492:


content/4_Gold/Divisibility.mdx
```diff
@@ -28,7 +28,13 @@ to start.
 	>
 		good book
 	</Resource>
-
+	<Resource
```
dolphingarlic, line 31: Move this to the other resource block

dolphingarlic (changes requested):

content/4_Gold/Divisibility.mdx
Aadit-Ambadkar: Also I feel like "contains nice proofs and problems" would be better than "nice proofs and problems"

SansPapyrus683: tbh i feel like the original is good

ryanchou-dev: i think "contains nice proofs and problems" would be better, but it doesn't really matter.

ryanchou-dev:
please address dolphingarlic's comment :)

ar492:
> please address dolphingarlic's comment :)

done

PR #2106, opened 2021-08-16
Link All Focus Problems with Modules (Bronze)

ryanchou-dev:
i'll do the other divisions later :)


SansPapyrus683 (approved):
i don't know how problems work so

solutions/cses-1621.mdx
```diff
@@ -1,102 +0,0 @@
----
```
maggie-j-liu, line 1: hmm, i think this was intentional to keep only the solution with sets in the module

ryanchou-dev: but isn't the module Introduction to Sets & Maps? 🤔 should i ask @thecodingwizard?

maggie-j-liu: yeah, i'm not too sure about this

what i meant was that the solution using sets would be in the module, since the module is intro to sets and maps, and the solution using sorting would be an internal solution since it doesn't relate to the module

dongliu0426: I agree w/ Maggie

ryanchou-dev: 👍

ryanchou-dev: i thought the other implementation was maps, sorry about that :)

maggie-j-liu (approved):

PR #2110, opened 2021-08-18
Created Solution for "CSES - 1740"

dutinmeow:


dutinmeow:
Is Vercel failing normal?

maggie-j-liu:
no, it's not normal 👀 
you can click on "Preview" 
image
to check the error (it tells you how to fix it)

dongliu0426:
this is already [a focus problem](https://usaco.guide/plat/range-sweep?lang=cpp#solution---intersection-points)

content/5_Plat/Range_Sweep.problems.json
```diff
@@ -10,7 +10,7 @@
       "isStarred": false,
       "tags": ["PURS"],
       "solutionMetadata": {
-        "kind": "none"
+        "kind": "internal"
```
dongliu0426, line 13: change this to 
```
      "solutionMetadata": {
        "kind": "in-module",
        "moduleId": "range-sweep"
      }
```

dongliu0426 (commented):
also this problem is in plat/sweep-line as well

dutinmeow:
> this is already [a focus problem](https://usaco.guide/plat/range-sweep?lang=cpp#solution---intersection-points)

Maybe just put a link from the RQ with SL to the solution? I use order-statistics tree instead of BIT

ar492:
> Maybe just put a link from the RQ with SL to the solution? I use order-statistics tree instead of BIT

what is rq and sl

dongliu0426:
> > this is already [a focus problem](https://usaco.guide/plat/range-sweep?lang=cpp#solution---intersection-points)
> 
> Maybe just put a link from the RQ with SL to the solution? I use order-statistics tree instead of BIT

sure

dongliu0426:
> > Maybe just put a link from the RQ with SL to the solution? I use order-statistics tree instead of BIT
> 
> what is rq and sl

range queries, sweep line

dutinmeow:
It still says that I'm missing the solution link from one of the modules

dongliu0426:
> also this problem is in plat/sweep-line as well

@dutinmeow

dutinmeow:
It should be fixed now, I thought that range sweep and sweep line were the same module for some reason

dutinmeow:
Vercel is dumb

dutinmeow:
How do I fix the issue now -_-

dutinmeow:
Already has solution

PR #1988, opened 2021-07-25
Add Distinct Colors Editorial

timothygao8710:
language: CPP

PR #2111, opened 2021-08-18
Add Distinct Colors Editorial

nwatx:
@SansPapyrus683


nwatx:
- [ ] detemplate

dongliu0426:
@nwatx isn't this already the focus problem of plat/merging? can you just provide a link to plat/merging instead? I think the intent of putting this problem in gold is because of the solution with PURS (#1988).

nwatx:
dong's comment ^

PR #2112, opened 2021-08-18
LIS page typo fix/clarity enhancements for proofs

topshenyi-web:


topshenyi-web:
Wait actually to make things clearer why not just make the definition of L_i[j] to be the smallest element from the first i elements with an increasing sequence of length j ending on it? (instead of j+1)

SansPapyrus683 (approved):

content/4_Gold/LIS.mdx
```diff
@@ -113,26 +113,26 @@ Thus, $L_i[j]$ has to be greater than $L_i[j - 1]$. From that, we can write this
 equation: $L_i[j - 1] \leq a_{j-1} < a_j = L_i[j]$, which is exactly what we
 wanted.
 
-**Lemma 2:** The length of the LIS ending on $A[i + 1]$ is equal to the least
-index $j$ (with 0-indexing) such that $L_i[j] \geq A[i + 1]$.
+**Lemma 2:** The length of the LIS ending on $A[i + 1]$ is equal to $j+1$, where $j$ is the least
```
dongliu0426, line 116: length of LIS is index?

topshenyi-web: I thought the length was index+1?
"Let $L_i$ be an array (0-indexed) where $L_i[j]$ is the smallest element from
the first $i$ elements of $A$ with **an increasing sequence of length $j + 1$
ending on it** (or $\infty$ if there is no such element)."

topshenyi-web: then again, that's kind of confusing to begin with... maybe I should just change the definition instead

PR #2118, opened 2021-08-19
walls internal solution

SansPapyrus683:
someone tell me if the complexity is right all i know is that it works
also i hate java so no java sol
(i could do it, it just would take so long given that there's no pair class and it needs a hashmap)

### DIFF
--- a/solutions/ioi-00-walls.mdx
+++ b/solutions/ioi-00-walls.mdx
@@ -147,7 +147,7 @@ neighbors = [[] for _ in range(region_num)]
 for r in range(region_num):
 	for w in regions[r]:
 		if w in prev_walls:
-			# dupes might hppen but that doesn't really matter
+			# dupes might happen but that doesn't really matter
 			neighbors[r].append(prev_walls[w])
 			neighbors[prev_walls[w]].append(r)
 		else:
--- a/solutions/ioi-00-walls.mdx
+++ b/solutions/ioi-00-walls.mdx
@@ -10,7 +10,7 @@ author: Kevin Sheng
 Instead of treating the towns as nodes, we're going to treat the regions themselves as nodes instead, with
 the walls still being the edges.
 We BFS from each node, logging the minimum distance from every region to every other region.
-Finally, we try all meeting regions and print out the one that minimizes the walls the members have to cross.
+Finally, we try all meeting regions and print out the one that minimizes the number of walls the members have to cross.
 
 ## Implementation
 
--- a/solutions/ioi-00-walls.mdx
+++ b/solutions/ioi-00-walls.mdx
@@ -10,7 +10,7 @@ author: Kevin Sheng
 Instead of treating the towns as nodes, we're going to treat the regions themselves as nodes instead, with
 the walls still being the edges.
 We BFS from each node, logging the minimum distance from every region to every other region.
-Finally, we try all meeting regions and print out the one that minimizes the walls the memebers have to cross.
+Finally, we try all meeting regions and print out the one that minimizes the walls the members have to cross.
 
 ## Implementation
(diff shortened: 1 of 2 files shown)


solutions/ioi-00-walls.mdx
```diff
@@ -0,0 +1,192 @@
...
+title: Walls
+author: Kevin Sheng
+---
+
+## Explanation
+
+Instead of treating the towns as nodes, we're going to treat the regions themselves as nodes instead, with
+the walls still being the edges.
+We BFS from each node, logging the minimum distance from every region to every other region.
+Finally, we try all meeting regions and print out the one that minimizes the walls the memebers have to cross.
```
ar492, line 13:
```suggestion
Finally, we try all meeting regions and print out the one that minimizes the walls the members have to cross.
```

ar492 (changes requested):

solutions/ioi-00-walls.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+</CPPSection>
+<PySection>
+
+```py
+region_num = int(input())
+town_num = int(input())
+member_num = int(input())
+members = [int(i) - 1 for i in input().split()]
+assert member_num == len(members)
```
ryanchou-dev, line 130:
```suggestion
```
forgot to remove?

solutions/ioi-00-walls.mdx
```diff
@@ -0,0 +1,192 @@
...
+	for t in range(border_num):
+		regions[r].append((border_towns[t], border_towns[(t + 1) % border_num]))
+
+# the walls previously processed (along with their associated regions)
+prev_walls = {}
+neighbors = [[] for _ in range(region_num)]
+for r in range(region_num):
+	for w in regions[r]:
+		if w in prev_walls:
+			# dupes might hppen but that doesn't really matter
```
ryanchou-dev, line 150:
```suggestion
			# dupes might happen but that doesn't really matter
```

ryanchou-dev (approved):
geniosity :D

solutions/ioi-00-walls.mdx
SansPapyrus683: i mean no harm in it being there

solutions/ioi-00-walls.mdx
```diff
@@ -0,0 +1,192 @@
...
+title: Walls
+author: Kevin Sheng
+---
+
+## Explanation
+
+Instead of treating the towns as nodes, we're going to treat the regions themselves as nodes instead, with
+the walls still being the edges.
+We BFS from each node, logging the minimum distance from every region to every other region.
+Finally, we try all meeting regions and print out the one that minimizes the walls the members have to cross.
```
nathangong, line 13:
```suggestion
Finally, we try all meeting regions and print out the one that minimizes the number of walls the members have to cross.
```

solutions/ioi-00-walls.mdx
nwatx: i think this is ok

nwatx (changes requested):
it seems like some of the comments could potentially be put in the editorial

SansPapyrus683:
what's "some"

nwatx:
> if we've seen this wall before, add that associated region

> dupes might not happen but they don't really matter

probably explaining this would be helpful (if it's not clear already)

SansPapyrus683:
1. how would i explain this in the editorial
2. it's just an implementation thing, doesn't really matter

nwatx (approved):

TruaShamu (approved):

HarryWangATX (approved):

PR #2123, opened 2021-08-21
Fix multiple activity counts for the same problem

maggie-j-liu:
If you solve a problem and unsolve/solve a few times, the activity in the chart will increase. This pr fixes it so that activity is only counted once per problem or module per day.

This doesn't fix some other problems such as solve + unsolve still counting as a solve, but hopefully makes the calculations a bit more accurate.


ryanchou-dev (approved):
looks good!

ar492 (approved):
maggie porz!

PR #2131, opened 2021-08-24
Update usaco 667

GZGavinZhao:
For solution 2 that uses an `int` array, all elements of the `seen` array should be initialized to `0`. Tested with the official USACO site analysis mode.

### DIFF
--- a/solutions/usaco-667.mdx
+++ b/solutions/usaco-667.mdx
@@ -55,7 +55,7 @@ $[0,26^4)$ and use an array instead of a map.
 
 using namespace std;
 
-int seen[26 * 26 * 26 * 26];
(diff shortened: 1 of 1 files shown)


dongliu0426:
Global `int`s or `int` arrays are initialized to `0` by default.

maggie-j-liu:
also, in c++ initializer lists initialize the array with values provided, but if there are less values than the size, it will set the remaining elements to zero (see [here](https://www.learncpp.com/cpp-tutorial/arrays-part-ii/) for more info). so you only need to do `int seen[26 * 26 * 26 * 26]{};` to set all values to zero

but, as Dong mentioned, this is unnecessary since global arrays are already zero initialized

PR #2142, opened 2021-08-29
Solution for CSES-2101

dutinmeow:
I'm not sure if my code's style is good enough.


solutions/cses-2101
```diff
@@ -0,0 +1,166 @@
+---
+id: cses-2101
+source: CSES
+title: New Roads Queries
+author: Dustin Miao
+---
+
+Let's construct a graph where edges have weights equal to the time they are added. The problem boils down to efficiently finding the path with the minimum maximum edge for each query. 
+
+Let's consider a single query between nodes $u, v$. One way of approaching this is to sort the edge weights from smallest to largest, and then add edges one by one until $u$ and $v$ are connected. If we use a [disjoint set union](https://usaco.guide/gold/dsu), the time complexity is $\mathcal{O}(m \cdot \alpha(n))$, which is too slow.
```
dongliu0426, line 10:
```suggestion
Let's consider a single query between nodes $u, v$. One way of approaching this is to sort the edge weights from smallest to largest, and then add edges one by one until $u$ and $v$ are connected. If we use a [disjoint set union](/gold/dsu), the time complexity for each query is $\mathcal{O}(m \cdot \alpha(n))$, which is too slow.
```

solutions/cses-2101
```diff
@@ -0,0 +1,166 @@
...
+
+However, this hints at a faster solution. Note that our slow approach bears many similarities with [kruskal's MST algorithm](https://usaco.guide/gold/mst/#kruskals), which implies that the path between $u$ and $v$ on any minimum spanning tree will minimize the maximum edge weight. For a more formal proof, see [here](https://www.geeksforgeeks.org/minimum-bottleneck-spanning-treembst/).
```
dongliu0426, line 12:
```suggestion
However, this hints at a faster solution. Note that our slow approach bears many similarities with [Kruskal's MST algorithm](/mst/#kruskals), which implies that the path between $u$ and $v$ on any minimum spanning tree will minimize the maximum edge weight. For a more formal proof, see [here](https://www.geeksforgeeks.org/minimum-bottleneck-spanning-treembst/). 
```

solutions/cses-2101
```diff
@@ -0,0 +1,166 @@
...
+author: Dustin Miao
+---
+
+Let's construct a graph where edges have weights equal to the time they are added. The problem boils down to efficiently finding the path with the minimum maximum edge for each query. 
+
+Let's consider a single query between nodes $u, v$. One way of approaching this is to sort the edge weights from smallest to largest, and then add edges one by one until $u$ and $v$ are connected. If we use a [disjoint set union](https://usaco.guide/gold/dsu), the time complexity is $\mathcal{O}(m \cdot \alpha(n))$, which is too slow.
+
+However, this hints at a faster solution. Note that our slow approach bears many similarities with [kruskal's MST algorithm](https://usaco.guide/gold/mst/#kruskals), which implies that the path between $u$ and $v$ on any minimum spanning tree will minimize the maximum edge weight. For a more formal proof, see [here](https://www.geeksforgeeks.org/minimum-bottleneck-spanning-treembst/). 
+
+All that is left to do is to efficiently query for the minimum node on a tree path. There are many methods to do this, but the simplest is with [binary jumping](https://usaco.guide/plat/binary-jump).
```
dongliu0426, line 14: There's actually an easier method (modifying the DSU & small to large merging), but wtvr

dongliu0426 (changes requested):
rename file to `cses-2101.mdx`

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+ 
+/** Disjoint Set Union with Small-to-Large Merging 
+ *  Verified: https://judge.yosupo.jp/problem/unionfind */
```
SansPapyrus683, line 29:
```suggestion
// Simple DSU class
```

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+using namespace std;
+ 
+/** Disjoint Set Union with Small-to-Large Merging 
+ *  Verified: https://judge.yosupo.jp/problem/unionfind */ 
+template<size_t N> struct UnionFind {
+	int par[N], sze[N], max_size;
+ 
+	UnionFind(int n = N) { init(n); }
+ 
+	/** Complexity: O(n) */
```
SansPapyrus683, line 35:
```suggestion
```

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+ 
+/** Disjoint Set Union with Small-to-Large Merging 
+ *  Verified: https://judge.yosupo.jp/problem/unionfind */ 
+template<size_t N> struct UnionFind {
+	int par[N], sze[N], max_size;
+ 
+	UnionFind(int n = N) { init(n); }
```
SansPapyrus683, line 33:
```suggestion
	UnionFind(int n=N) { init(n); }
```
think this is the style for default arguments

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+ 
+	/** Complexity: O(n) */ 
+	void init(int n = N) {
+		iota(par, par + n, 0);
+		fill(sze, sze + n, 1);
+		max_size = 1;
+	}
+ 
+	/** Complexity: O(alpha(n)) */
```
SansPapyrus683, line 42: you can just remove all these complexity lines, don't think they're needed

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+ 
+	/** Complexity: O(alpha(n)) */ 
+	int size(int a) { 
+		return sze[find(a)]; 
+	}
+};
+ 
+int N, M, Q;
+vector<pair<int, int>> G[200005];
+UnionFind<200005> dsu;
```
SansPapyrus683, line 76: so magic numbers are bad blah blah blah
put 200k in a number called `MAX_N` or something and then in declaring these arrays
do something like `int arr[MAX_N + 1]`

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+ 
+namespace LCA {
+	int dep[200005], par[200005][18], val[200005][18];
```
SansPapyrus683, line 79: yeah same here lol
also the heck is `18` for

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,166 @@
...
+		ans = max(ans, val[a][0]);
+		ans = max(ans, val[b][0]);
+ 
+		return ans;
+	}
+}
+ 
+ 
+int main() {
+	ios_base::sync_with_stdio(0); cin.tie(0);
```
SansPapyrus683, line 141: if it runs w/o this, delete it

SansPapyrus683 (changes requested):
this is what i've nitpicked so far

solutions/cses-2101.mdx
dutinmeow: 18 = log_2 N

SansPapyrus683: you should definitely add that as a comment

solutions/cses-2101.mdx
```diff
@@ -0,0 +1,163 @@
...
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+	
+const int MAXN = 2e5;
+const int LOGN = 18;
```
SansPapyrus683, line 29: still add a comment for this

dutinmeow: why?????

SansPapyrus683: log what? base what?
just say like `// log of MAXN base 2`

SansPapyrus683 (approved):
i have arrays but there's nothing i can do

PR #2154, opened 2021-08-31
Internal solution + java code for CSES Range Update Queries

nathangong:


caoash:
I think this explanation would benefit from a custom module similar to the one in the prefix sums module (since it's not immediately obvious why incrementing diff[a] and decrementing diff[b + 1] will update a range). However, I'm going to merge for now because the explanation is good and we can always add it later.

PR #2169, opened 2021-09-04
Update Data_Types.mdx

tvottra:
The range for 32-bit int should be -2^31 to 2^31-1 because the 0th bit is used as the sign bit.


Aadit-Ambadkar (approved):

SansPapyrus683 (approved):
yeah alright

dongliu0426 (approved):
thanks

PR #2178, opened 2021-09-07
Update cf-919D.mdx

TruaShamu:
Updated sol.

1. Changed 1 or 2 letters in the explanation to make it clearer
2. Replaced implementation because the old one is wrong, and gives compile error. I'll probably make another pr later with the cpp code as well.


SansPapyrus683 (approved):
ok rest looks alr

dongliu0426 (approved):

PR #2203, opened 2021-09-18
Fix Tex

Anri-Lombard:
Literally the smallest change, but it bugged me, so thought I could point it out.


maggie-j-liu (approved):
thank you!

PR #2227, opened 2021-09-26
Update MST.mdx

Aadit-Ambadkar:


SansPapyrus683 (approved):
i'll just assume that the html is correct
and just

brad182 (approved):

PR #2250, opened 2021-10-09
Add formulas to Rectangle Geometry

Navigator365:
Per Ryan's suggestion: 
> suggestion:
we should add formulas to bronze rectangle geometry, stuff like finding the overlapping area of a rectangle from coordinates, checking to see if two rectangles intersect?
i think looking this stuff up was pretty hard (at least when i was in bronze, now i just store these formulas in a doc) :( 

Added the two formulas, but not code. Lmk if code is needed too.


ryanchou-dev:
> Lmk if code is needed too.

hmm it might help? if you want to, i can provide c++ and python code :D

ryanchou-dev:
> tr_a represents the entire top-right coordinate of rectangle a, which needs to be clarified bc some parts deal with 2 rectangles. Invididual x and ys are tr_a{_y}. Though if this is confusing, feel free to suggest another way.

maybe `a_tr_x` would be better

PR #2255, opened 2021-10-10
added c++ support for family tree

SansPapyrus683:
i tested yeah


SansPapyrus683:
isn't lca like a frickin gold topic these days

PR #2259, opened 2021-10-11
Fix modal bug

MrinallU:


MrinallU:
And I don't think decreasing the z index changes anything.

thecodingwizard:
@MrinallU it makes the scrollbar look a bit off, and might be messy on mobile too. I don't think `inset-0` is a good solution (if we wanted to change the position of the modal, we should use padding, not inset).

If increasing the z-index doesn't fix it, it's possible there's some stacking context issue: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

MrinallU:
@thecodingwizard How's this?

thecodingwizard:
@MrinallU I meant on the dashboard page

MrinallU:
@thecodingwizard How do I get the announcement to show up on my side?

PR #2305, opened 2021-11-03
Update usaco-736 - Add C++ Solution in O(M * N).mdx

theMasix:
Add C++ Solution in O(M * N)

done._

- [x] I have tested my code

### DIFF
--- a/solutions/usaco-736.mdx
+++ b/solutions/usaco-736.mdx
@@ -188,3 +188,75 @@ public class BovineGenomics {
 </LanguageSection>
 
 
+## Solution 3 - Unordered Map
+
+<LanguageSection>
+
+<CPPSection>
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(M * N)$
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int n, m, ans = 0;
+vector<string> gens;
+
+// Use `unordered_map` to achieve O(1) time complexity in queries
+unordered_map<char, int> spotties;
+
+int main()
+{
+  ios::sync_with_stdio(false);
+  cin.tie(0), cout.tie(0);
+
+  cin >> n >> m;
+  gens = vector<string>(2 * n);
+
+  for (int i = 0; i < 2 * n; i++)
+  {
+    cin >> gens[i];
+  }
+
+  for (int i = 0; i < m; i++)
+  {
+    // Let's clear characters history
+    spotties.clear();
+
+    for (int j = 0; j < 2 * n; j++)
+    {
+      char c = gens[j][i];
+
+      if (j < n)
+      {
+        // Spotty cows
+        // Check character `c` as seen
+        spotties[c]++;
+      }
+      else
+      {
+        // Plain cows
+        // Check if the gen has been seen on spotty cows
+        if (spotties.find(c) != spotties.end())
+          break;
+      }
+
+      // If the for-loop hasn't been broken yet, then we found another gen
+      if (j == 2 * n - 1)
+        ans++;
+    }
+  }
+
+  cout << ans;
+  return 0;
+}
+```
+
+</CPPSection>
+
+</LanguageSection>
+
+


SansPapyrus683:
> I have followed the code style guidelines mentioned
here

image

solutions/usaco-736.mdx
```diff
@@ -188,3 +188,75 @@ public class BovineGenomics {
...
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int n, m, ans = 0;
+vector<string> gens;
+
+// Use `unordered_map` to achieve O(1) time complexity in queries
+unordered_map<char, int> spotties;
```
caoash, line 209: use array, since you're storing chars. unordered map is sometimes slow (in this case, it won't matter, but in general it's a good idea to use array whenever possible)

solutions/usaco-736.mdx
```diff
@@ -188,3 +188,75 @@ public class BovineGenomics {
...
+      else
+      {
+        // Plain cows
+        // Check if the gen has been seen on spotty cows
+        if (spotties.find(c) != spotties.end())
+          break;
+      }
+
+      // If the for-loop hasn't been broken yet, then we found another gen
+      if (j == 2 * n - 1)
```
caoash, line 248: can you stay consistent and either use brackets or put it on the same line

caoash (changes requested):

solutions/usaco-736.mdx
theMasix: Sure

solutions/usaco-736.mdx
theMasix: How is it possible to be slow?
https://github.com/gibsjose/cpp-cheat-sheet/blob/master/Data%20Structures%20and%20Algorithms.md#15-map-stdmap-and-stdunordered_map

caoash: constant factor

SansPapyrus683: i honestly disagree with michael here- i'd recommend using a map instead of an array

caoash: here its fine. but there's times where you TLE with map and passes with array

SansPapyrus683: those only happen at and above gold tbh

solutions/usaco-736.mdx
```diff
@@ -188,3 +188,75 @@ public class BovineGenomics {
...
+
+int n, m, ans = 0;
+vector<string> gens;
+
+// Use `unordered_map` to achieve O(1) time complexity in queries
+unordered_map<char, int> spotties;
+
+int main()
+{
+  ios::sync_with_stdio(false);
```
dutinmeow, line 213: If the code works without these then remove them
Also the style guide explicitly mentions that `cout.tie(0)` is not allowed
image

solutions/usaco-736.mdx
```diff
@@ -188,3 +188,75 @@ public class BovineGenomics {
...
+      }
+
+      // If the for-loop hasn't been broken yet, then we found another gen
+      if (j == 2 * n - 1)
+        ans++;
+    }
+  }
+
+  cout << ans;
+  return 0;
```
dutinmeow, line 254: remove, unnecessary

solutions/usaco-736.mdx
dutinmeow, line 213: tabs

dutinmeow (changes requested):

solutions/usaco-736.mdx
caoash: in general its good to teach that you shouldnt use maps if not needed so you dont die to it in the future

i've died enough times to say this :P

SansPapyrus683: ok yeah true
but why not vectors then?

caoash: thats ok too, idrc

solutions/usaco-736.mdx
```diff
@@ -188,3 +188,75 @@ public class BovineGenomics {
...
+**Time Complexity:** $\mathcal{O}(M * N)$
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int n, m, ans = 0;
+vector<string> gens;
+
+// Use `unordered_map` to achieve O(1) time complexity in queries
```
Aadit-Ambadkar, line 208:
```suggestion
// Use `unordered_map` to achieve average O(1) time complexity in queries
```

Aadit-Ambadkar (changes requested):

solutions/usaco-736.mdx
dutinmeow: Isn't it always `O(1)`

PR #2306, opened 2021-11-04
Matrix Exponentiation Focus problem

nwatx:
done._

- [x] I have tested my code

### DIFF
--- a/content/5_Plat/Matrix_Expo.mdx
+++ b/content/5_Plat/Matrix_Expo.mdx
@@ -100,17 +100,17 @@ vector<ll> mul(vector<ll> a, vector<ll> b) {
 	// [0 1]
 	// [2 3]
 	return vector<ll>{
-		(a[0]*b[0]+a[1]*b[2]) % MOD,
-		(a[0]*b[1]+a[1]*b[3]) % MOD,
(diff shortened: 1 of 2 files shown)


content/5_Plat/Matrix_Expo.mdx
```diff
@@ -37,6 +37,90 @@ frequency: 1
...
+```cpp
+#include <bits/stdc++.h>
+
+using namespace std;
+
+using ll = long long;
+
+const ll MOD = (ll)1e9 + 7;
+
+vector<ll> mul(vector<ll> a, vector<ll> b) {
```
bqi343, line 99: why not `array<array<ll,2>,2>`?

bqi343 (changes requested):

dutinmeow (approved):
Change the `vector<ll>` into `Matrix` by doing something like
`using Matrix = array<array<int ,2>, 2>;`

content/5_Plat/Matrix_Expo.mdx
nwatx: it was easier to type but I'll go change it real quick

bqi343 (approved):

PR #2356, opened 2021-11-23
Update String_Hashing.problems.json

ar492:
Added a problem

done._

- [x] I have tested my code

### DIFF
--- a/content/4_Gold/String_Hashing.problems.json
+++ b/content/4_Gold/String_Hashing.problems.json
@@ -30,6 +30,19 @@
       }
     },
     {
+      "uniqueId": "cses-1753",
+      "name": "String Matching",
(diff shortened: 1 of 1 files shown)


dongliu0426:
I don't see a point in adding this problem. It's exactly the same as the focus problem (https://dmoj.ca/problem/ccc20s3).

ar492:
> I don't see a point in adding this problem. It's exactly the same as the focus problem (https://dmoj.ca/problem/ccc20s3).

They are not the same; the focus problem deals with permutations

SansPapyrus683:
i mean the essence is basically the same- using hashing for comparing strings
focus problem is a more big brain use of it as well imo

ar492:
On second thought, I agree. The focus problem also has better test data.

PR #2361, opened 2021-11-24
Update PGS.ts

### LINKED ISSUE #2360: Wrong page of book linked on Rectangle Geometry
Hello! 

When I view the Rectangle Geometry page, and click on the link to the book that the module is based off, it takes me to a page (page 33) after the section about Rectangle Geometry (page 31). I couldn't find where these Resources links are stored, so I couldn't send a PR.

nwatx:
done._

- [x] I have tested my code

resolves #2360


SansPapyrus683 (approved):
can only assume this is correct

src/components/markdown/PGS.ts
```diff
@@ -26,9 +26,9 @@ export default {
     '6.2': 31,
     '6.3': 31,
     '7': 33,
-    '7.1': 33,
-    '7.2': 33,
-    '7.3': 34,
+    '7.1': 31,
```
bqi343, line 29: no way this is correct? the numbers are no longer in increasing order ...

bqi343 (changes requested):

src/components/markdown/PGS.ts
nwatx: it's no longer in increasing order, but it's [correct](https://darrenyao.com/usacobook/cpp.pdf#page=31), I believe the others have to be updated as well

nwatx: i'll look into it more i think a shift in pages occurred

nwatx: er ok it seems like all the pages are shifted by a non-constant amount, do you want me to renumber the entire file

bqi343: Yes. I believe that I copied the table of contents and used a Python script to extract the section and page numbers, but I'm not sure where it is anymore.

bqi343:
Could you format this the same way as it was originally? also, looks like some things are missing for cpp.

bqi343 (approved):

PR #2433, opened 2021-12-24
Dec21 probs

bqi343:
Based off #902

done._

- [x] I have tested my code


content/3_Silver/Greedy_Sorting.problems.json
```diff
@@ -213,6 +213,19 @@
         "kind": "autogen-label-from-site",
         "site": "CF"
       }
+    },
+    {
+      "uniqueId": "usaco-1158",
+      "name": "Closest Cow Wins",
+      "url": "http://www.usaco.org/index.php?page=viewproblem2&cpid=1158",
+      "source": "Silver",
+      "difficulty": "Hard",
```
jessechoe10, line 222:
```suggestion
      "difficulty": "Very Hard",
```

bqi343:
Some notes:
 - Marked "Closest Cow Wins" as "Hard" instead of "Very Hard" in the module since it seemed weird to jump from "Normal" to "Very Hard." Plus, it's not really that much harder than the other problems in the module.
 - Not planning on adding "Connecting Two Barns" or "Convoluted Intervals" to the corresponding modules unless someone can make a convincing case that they should be.

content/extraProblems.json
```diff
@@ -472,6 +472,110 @@
...
+      "difficulty": "Hard",
+      "isStarred": false,
+      "tags": ["Greedy"],
+      "solutionMetadata": {
+        "kind": "USACO",
+        "usacoId": "1156"
+      }
+    },
+    {
+      "uniqueId": "usaco-1159",
```
jessechoe10, line 489: If you don't want to add this to DFS module, maybe add it to the conclusion problems module

content/extraProblems.json
```diff
@@ -472,6 +472,110 @@
...
+      "difficulty": "Normal",
+      "isStarred": false,
+      "tags": ["Connected Components"],
+      "solutionMetadata": {
+        "kind": "USACO",
+        "usacoId": "1159"
+      }
+    },
+    {
+      "uniqueId": "usaco-1160",
```
jessechoe10, line 502: Maybe add this to module. It wouldn't hurt to add a "hard" prefix sums problem to the module (consider there is only one).

jessechoe10 (changes requested):

jessechoe10:
Also, why is Closest Cow wins the only problem with a module. Do these problems not teach something new lol.

content/extraProblems.json
bqi343: Can't recall having ever added anything to the conclusion problems, you can do in separate PR if you want.

content/extraProblems.json
bqi343: From DM: I don't think convoluted intervals teaches you anything new about prefix sums. The point was just that you should use M being small somehow

PR #2458, opened 2021-12-31
n * log(n) time complexity for python

C0derTang:
for python, since the starting list is already sorted, you just have to look at the elements to the left and right of the list. The base loop takes O(n), but the sorting takes O(n * log(n), so the worst case scenario is O(n * log(n))

This could also be implemented for the other languages, but I'm only familiar with python

done._

- [ ] I have tested my code
- [ ] I have followed the code style guidelines mentioned
- [ ] I have properly formatted the files according to the steps mentioned

### DIFF
--- a/solutions/usaco-808.mdx
+++ b/solutions/usaco-808.mdx
@@ -34,40 +34,35 @@ balls = [0] * n
 ans = 0
 
 def nearestneighbor(ind):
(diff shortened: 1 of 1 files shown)


bqi343 (changes requested):
> I have followed the code style guidelines mentioned here

Please do

PR #2492, opened 2022-01-10
Java solution for connecting two barns

nathangong:

- [x] I have tested my code.
- [x] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions), which inclues the following: 

### DIFF
--- a/solutions/usaco-1159.mdx
+++ b/solutions/usaco-1159.mdx
@@ -86,11 +86,11 @@ public class ConnectingTwoBarns {
 				int dist = Math.abs(barn1.get(barn1Index) - i);
 
 				/*
-				* if the distance between i and the current field in barn1's
-				* component is larger than the distance between i and the next
-				* field in barn1's component, increment barn1Index until this
-				* no longer holds true
-				*/
+				 * if the distance between i and the current field in barn1's
+				 * component is larger than the distance between i and the next
+				 * field in barn1's component, increment barn1Index until this
+				 * no longer holds true
+				 */
 				while (
 					// make sure we don't go out of bounds
 					barn1Index < barn1.size() - 1
--- a/solutions/usaco-1159.mdx
+++ b/solutions/usaco-1159.mdx
@@ -98,11 +98,11 @@ public class ConnectingTwoBarns {
 				}
 
 				/*
-				* we've now found the field in barn 1's component that's closest to i
-				*
-				* we can use this to update the minimum distance between field i's
-				* component and barn 1's component
-				*/
+				 * we've now found the field in barn 1's component that's closest to i
+				 *
+				 * we can use this to update the minimum distance between field i's
+				 * component and barn 1's component
+				 */
 				dist1[visited[i]] = Math.min(dist, dist1[visited[i]]);
 			}
 
--- a/solutions/usaco-1159.mdx
+++ b/solutions/usaco-1159.mdx
@@ -28,7 +28,9 @@ public class ConnectingTwoBarns {
 			int m = io.nextInt();
 
 			List<List<Integer>> adj = new ArrayList<>();
-			for (int i = 0; i < n; i++) adj.add(new ArrayList<>());
+			for (int i = 0; i < n; i++) {
+		   		adj.add(new ArrayList<>());
+			}
 			for (int i = 0; i < m; i++) {
 				int a = io.nextInt() - 1;
 				int b = io.nextInt() - 1;
--- a/solutions/usaco-1159.mdx
+++ b/solutions/usaco-1159.mdx
@@ -1,0 +1,140 @@
+---
+id: usaco-1159
+source: USACO Silver 2021 December
+title: Connecting Two Barns
+author: Nathan Gong
+---
+
+[Official Analysis (C++)](http://www.usaco.org/current/data/sol_prob2_gold_open21.html)
+
+## Implementation - DFS + Two Pointers
+
+**Time Complexity:** $\mathcal{O}(T\cdot (N + M))$
+
+<LanguageSection>
+
+<JavaSection>
+
+```java
+import java.util.*;
+import java.io.*;
+
+public class ConnectingTwoBarns {
+	public static void main(String[] args) {
+		Kattio io = new Kattio();
+		int t = io.nextInt();
+		for (int test = 0; test < t; test++) {
+			int n = io.nextInt();
+			int m = io.nextInt();
+
+			List<List<Integer>> adj = new ArrayList<>();
+			for (int i = 0; i < n; i++) adj.add(new ArrayList<>());
+			for (int i = 0; i < m; i++) {
+				int a = io.nextInt() - 1;
+				int b = io.nextInt() - 1;
+				adj.get(a).add(b);
+				adj.get(b).add(a);
+			}
+
+			// stores the component each node is in
+			int[] visited = new int[n];
+			// stores a list of nodes for each component
+			List<List<Integer>> components = new ArrayList<>();
+
+			int numComponents = 0;
+			Arrays.fill(visited, -1);
+			for (int i = 0; i < n; i++) {
+				// run DFS on node if it hasn't been visited
+				if (visited[i] == -1) {
+					components.add(new ArrayList<>());  // add new component
+					Stack<Integer> stack = new Stack<>();
+					stack.push(i);
+					while (!stack.isEmpty()) {
+						int curr = stack.pop();
+						if (visited[curr] != -1) continue;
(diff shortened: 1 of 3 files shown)


solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,140 @@
...
+public class ConnectingTwoBarns {
+	public static void main(String[] args) {
+		Kattio io = new Kattio();
+		int t = io.nextInt();
+		for (int test = 0; test < t; test++) {
+			int n = io.nextInt();
+			int m = io.nextInt();
+
+			List<List<Integer>> adj = new ArrayList<>();
+			for (int i = 0; i < n; i++) adj.add(new ArrayList<>());
```
SansPapyrus683, line 31: braces

SansPapyrus683: also can't `adj` just be an array of arraylists?

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,140 @@
...
+			int[] visited = new int[n];
+			// stores a list of nodes for each component
+			List<List<Integer>> components = new ArrayList<>();
+
+			int numComponents = 0;
+			Arrays.fill(visited, -1);
+			for (int i = 0; i < n; i++) {
+				// run DFS on node if it hasn't been visited
+				if (visited[i] == -1) {
+					components.add(new ArrayList<>());  // add new component
```
SansPapyrus683, line 49: seems a bit too long, idk tho

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,140 @@
...
+				while (
+					// make sure we don't go out of bounds
+					barn1Index < barn1.size() - 1
+					&& Math.abs(barn1.get(barn1Index + 1) - i) < dist
+				) {
+					barn1Index++;
+				}
+
+				/*
+				* we've now found the field in barn 1's component that's closest to i
```
SansPapyrus683, line 101: missing a space in front of the *

SansPapyrus683 (changes requested):
wow what long code

ryanchou-dev (approved):
looks good formatting wise, great job!

solutions/usaco-1159.mdx
nathangong: 76 chars

solutions/usaco-1159.mdx
nathangong: it can, but i just prefer this so the syntax doesn't change

nathangong:
Yeah it's slightly on the long side because i have a lot of comments and want to make it more readable

solutions/usaco-1159.mdx
SansPapyrus683: still, braces?

bqi343:
> Note, the deployment preview throws an error because it thinks that the problem isn't linked but it should work when merged through Ryan's PR

huh? I think you need to change the solution metadata to "internal"

nathangong:
🤦‍♂️ oh yeah whoops I didn't realize the DFS problems were also merged into this PR

nathangong:
Wait nvm it's not. The reason why I can't change it is Ryan pushed a PR that removes the problem from extraProblems and adds it to DFS problems. In this PR the problem is still in extraProblems so I can't change the solution to internal or else it would cause conflicts once it gets merged

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,142 @@
...
+						// mark node as visited
+						visited[curr] = numComponents;
+
+						for (int neighbor : adj.get(curr)) {
+							stack.push(neighbor);
+						}
+					}
+					numComponents++;
+				}
+			}
```
jessechoe10, line 68: The tabs looks super weird here idk. Also idt it matters but maybe for the sake of making ur code shorter do recursive 👀

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,142 @@
...
+			// sort nodes in each component by position
+			for (List<Integer> component : components) {
+				Collections.sort(component);
+			}
```
jessechoe10, line 72: Wait if you do this, the solution would be `O(m + n log n)`. I thought the editorial says it's `O(n + m)`, since the order of the DFS guarantees that the nodes in each component are in sorted order 😛.

solutions/usaco-1159.mdx
bqi343: this is fine

nathangong:
> Wait if you do this, the solution would be O(m + n log n). I thought the editorial says it's O(n + m), since the order of the DFS guarantees that the nodes in each component are in sorted order 😛.

So the edges are given in sorted order?

nathangong:
Oh wow it works, i'll probably just keep the sort though to prevent any confusion

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,142 @@
...
+			Arrays.fill(dist1, Integer.MAX_VALUE);
+			Arrays.fill(dist2, Integer.MAX_VALUE);
+
+			// use two pointer to populate dist1
+			int barn1Index = 0;
+			for (int i = 0; i < n; i++) {
+				int dist = Math.abs(barn1.get(barn1Index) - i);
+
+				/*
+				* if the distance between i and the current field in barn1's
```
SansPapyrus683, line 89: missing leading spaces

SansPapyrus683 (changes requested):
just one small thing

SansPapyrus683 (approved):

jessechoe10:
> Oh wow it works, i'll probably just keep the sort though to prevent any confusion

I sugges that you comment out the sort. We want the solution on guide to be the most efficient, so idk maybe comment out the sort to indicate that it'll still work.

bqi343:
> Oh wow it works, i'll probably just keep the sort though to prevent any confusion

I'm confused; if the sort is unnecessary, then doesn't leaving it in introduce confusion rather than preventing it?

nathangong:
Nvm i resubmitted it and it semes like the solution returns a wrong answer for test case 10. But, I think leaving it in makes it more clear because the two pointers requires all of the positions to be sorted; since the question doesn't explicitly say that the edges are given in sorted order, it would be confusing to anyone who doesn't already know that

bqi343:
Even if the edges were in sorted order, why would that guarantee that both `barn1` and `barn2` are in sorted order?

nathangong:
Wdym, I dont sort the edges, I sort the positions in each component from the DFS, which includes barn1 and barn2.

nathangong:
In terms of the input, I think it's given in a way where a regular DFS will give the nodes in each component in order (but this doesn't happen with tc 10)

bqi343:
> Wdym, I dont sort the edges, I sort the positions in each component from the DFS, which includes barn1 and barn2.

I was referring to your earlier comment: "So the edges are given in sorted order?"

> In terms of the input, I think it's given in a way where a regular DFS will give the nodes in each component in order (but this doesn't happen with tc 10)

This seems unlikely - I think what's happening is that the correct answer is being printed *despite* the nodes in each component being out of order.

nathangong:
Yeah that's also possible, which is why sorting would probably make more sense

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,142 @@
...
+			for (int i = 0; i < numComponents; i++) {
+				long cost = dist1[i] * dist1[i] + dist2[i] * dist2[i];
+				min = Math.min(min, cost);
+			}
+			io.println(min);
+		}
+		io.close();
+	}
+
+	// CodeSnip{Kattio}
```
bqi343, line 136: not related to this PR specifically but is there a reason having `Kattio` as a `static class` inside of `ConnectingTwoBarns` is preferred to having `Kattio` outside of `ConnectingTwoBarns`?

solutions/usaco-1159.mdx
```diff
@@ -0,0 +1,142 @@
+---
+id: usaco-1159
+source: USACO Silver 2021 December
+title: Connecting Two Barns
+author: Nathan Gong
+---
+
+[Official Analysis (C++)](http://www.usaco.org/current/data/sol_prob2_gold_open21.html)
```
nwatx, line 8: this is the wrong editorial link?

nwatx (changes requested):

solutions/usaco-1159.mdx
nathangong: whoops my bad

nwatx (approved):

PR #2516, opened 2022-01-18
Implement Template System

MrinallU:
done._

- [x] I have tested my code


RealMyth21 (approved):
looks good

RealMyth21:
just keep in mind, it failed build

MrinallU:
I'll fix that.

src/components/Groups/GroupPage/FeedItem.tsx
```diff
@@ -77,11 +92,116 @@ export default function FeedItem({
...
+            getFirestore(firebaseApp),
+            'groups',
+            key,
+            'posts',
+            docRef.id,
+            'problems'
+          )
+        );
+
+        problems.map(async problem => {
```
maggie-j-liu, line 172: problems are not exported... maybe you need to await this (also wrap in `Promise.all`?)

maggie-j-liu (changes requested):
is there a way to export multiple posts at once? can you make a way to copy an entire group?

src/components/Groups/GroupPage/FeedItem.tsx
MrinallU: Forgot to add that, I'll fix it.

MrinallU:
> is there a way to export multiple posts at once? can you make a way to copy an entire group?

So for the first question: yeah I can do that pretty easily. But as for copying an entire group, would it just be the same group wiht a different name?

maggie-j-liu:
> would it just be the same group with a different name?

basically, but it would allow the person who copied to become owner of the group and remove all other members (so we could make a "template" group for clubs, then club leaders can copy the entire group instead of having to set everything up themselves

PR #2584, opened 2022-02-16
Added c++ solution for Where's Bessie (usaco-740)

pranavgithub1:

- [x] I have tested my code.

### DIFF
--- a/solutions/usaco-740.mdx
+++ b/solutions/usaco-740.mdx
@@ -8,6 +8,154 @@ author: Qi Wang
 [Official Analysis](http://www.usaco.org/current/data/sol_where_silver_open17.html)
 
 <LanguageSection>
+
+<CPPSection>
+```cpp
+#include <bits/stdc++.h> 
+using namespace std;
+
+const int MAX_N = 20;
+int n;
+vector<vector<char>> image(MAX_N, vector<char>(MAX_N));
+vector<vector<bool>> visited(MAX_N, vector<bool>(MAX_N));
+int PCL_count = 0;
+
+// PCL described by top-left (i1,j1) and bottom-right (i2,j2) corners
+struct PCL {
+	int i1, j1, i2, j2;
+};
+
+// Floodfill to find connected regions
+int i_min, i_max, j_min, j_max;
+void floodfill(int i , int j, char color){
+	if(
+		i < i_min || j < j_min || i > i_max || j > j_max ||
+		visited[i][j] ||
+		image[i][j] != color
+	){
+		return;
+	}
+
+	visited[i][j] = true;
+
+	floodfill(i + 1, j, color);
+	floodfill(i - 1, j, color);
+	floodfill(i, j + 1, color);
+	floodfill(i, j - 1, color);
+}
+
+// Check whether a given region is a PCL
+bool is_PCL(int i1, int j1, int i2, int j2){
+
+	// keep track of num of regions for each color A-Z
+	vector<int> region_count(26, 0);
+
+	// set bounds for floodfill
+	i_min = i1;
+	i_max = i2;
+	j_min = j1;
+	j_max = j2;
+
+	// Count # of connected regions for each color present in bounds
+	for(int i = i1; i <= i2; i++){
+		for(int j = j1; j <= j2; j++){
+			if(!visited[i][j]){
+				char cur_color = image[i][j];
+				region_count[cur_color - 'A']++;
+				floodfill(i, j, cur_color);
+			}
+		}
+	}
+
+	// reset visited vector for the next call
+	fill(visited.begin(), visited.end(), vector<bool>(MAX_N, false));
+
+	// verify PCL conditions
+	int color_count = 0;
+	bool color_with_one_region = false;
+	bool color_with_more_regions = false;
+	for(int i = 0; i < 26; i++){
+		if(region_count[i]){
+			color_count++;
+		}
+		if(region_count[i] == 1){
+			color_with_one_region = true;
+		}
+		if(region_count[i] > 1){
+			color_with_more_regions = true;
+		}
+	}
+
+	if(
+		color_count == 2 &&
+		color_with_one_region &&
+		color_with_more_regions
+	){
+		return true;
+	}
+	else {
+		return false;
+	}
+
+}
+
+// Check if PCL a is inside PCL b
+bool is_inside(PCL a, PCL b){
+	return (
+		a.i1 >= b.i1 &&
+		a.i2 <= b.i2 &&
+		a.j1 >= b.j1 &&
+		a.j2 <= b.j2
+	);
+}
+
+int main() {
+	freopen("where.in", "r", stdin);
+	freopen("where.out", "w", stdout);
+	cin >> n;
+	for(int i = 0; i < n; i++){
+		for(int j = 0; j < n; j++){
+			cin >> image[i][j];
+		}
+	}
+	vector<PCL> PCL_list;
+
+	// brute force every rectangle to find PCLs
+	for(int i1 = 0; i1 < n; i1++){
+		for(int j1 = 0; j1 < n; j1++){
+			for(int i2 = 0; i2 < n; i2++){
+				for(int j2 = 0; j2 < n; j2++){
+					if(is_PCL(i1, j1, i2, j2)){
+						PCL new_PCL = {i1, j1, i2, j2};
+						PCL_list.push_back(new_PCL);
+					}
+				}
+			}
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


solutions/usaco-740.mdx
```diff
@@ -2,12 +2,160 @@
...
+using namespace std;
+
+const int MAX_N = 20;
+int n;
+vector<vector<char>> image(MAX_N, vector<char>(MAX_N));
+vector<vector<bool>> visited(MAX_N, vector<bool>(MAX_N));
+int PCL_count = 0;
+
+// PCL described by top-left (i1,j1) and bottom-right (i2,j2) corners
+struct PCL {
```
SansPapyrus683, line 24: the heck is a PCL

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,160 @@
...
+	cin >> n;
+	for(int i = 0; i < n; i++){
+		for(int j = 0; j < n; j++){
+			cin >> image[i][j];
+		}
+	}
+	vector<PCL> PCL_list;
+
+	// brute force every rectangle to find PCLs
+	for(int i1 = 0; i1 < n; i1++){
```
SansPapyrus683, line 124: spaces around your control flow statements
```suggestion
	for (int i1 = 0; i1 < n; i1++) {
```

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,160 @@
...
+				is_PCL = false;
+				break;
+			}
+		}
+		if(is_PCL){
+			PCL_count++;
+		}
+	}
+
+	cout << PCL_count;
```
SansPapyrus683, line 154: add an endl @ the end

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,160 @@
...
+	if(
+		color_count == 2 &&
+		color_with_one_region &&
+		color_with_more_regions
+	){
+		return true;
+	}
+	else {
+		return false;
+	}
```
SansPapyrus683, line 98: just do a `return [this ungodly expression here]`

SansPapyrus683 (commented):
let's start slow

solutions/usaco-740.mdx
SansPapyrus683: how can there be two top-left corners?

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,154 @@
...
+
+	// Reset visited vector for the next call
+	fill(visited.begin(), visited.end(), vector<bool>(MAX_N, false));
+
+	// Verify PCL conditions
+	int color_count = 0;
+	bool color_with_one_region = false;
+	bool color_with_more_regions = false;
+	for (int i = 0; i < 26; i++) {
+		if (region_count[i]) {
```
SansPapyrus683, line 78:
```suggestion
		if (region_count[i] != 0) {
```

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,154 @@
...
+			color_count++;
+		}
+		if (region_count[i] == 1) {
+			color_with_one_region = true;
+		}
+		if (region_count[i] > 1) {
+			color_with_more_regions = true;
+		}
```
SansPapyrus683, line 86:
```suggestion
		color_with_more_regions = region_count[i] >= 1;
```

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,154 @@
...
+int main() {
+	freopen("where.in", "r", stdin);
+	freopen("where.out", "w", stdout);
+	cin >> n;
+	for (int i = 0; i < n; i++) {
+		for (int j = 0; j < n; j++) {
+			cin >> image[i][j];
+		}
+	}
+	vector<PCL> PCL_list;
```
SansPapyrus683, line 115: don't use all caps in snake case
just use names like `pcl_list` and `is_pcl`

SansPapyrus683 (changes requested):

solutions/usaco-740.mdx
pranavgithub1: I am assuming you meant
```cpp
color_with_one_region = region_count[i] == 1;
color_with_more_regions = region_count[i] > 1;
```
Because otherwise `color_with_one_region` would not be accounted for.

pranavgithub1: Oh wait actually this wouldn't work because the `color_with_one_region` and `color_with_more_regions` flags need to be true only once for any value of `i`. With an assignment like this, they might get marked true and then get marked false later on.

SansPapyrus683: hm
yeah that makes sense
then
```suggestion
		color_with_more_regions = color_with_more_regions || region_count[i] >= 1;

File: solutions/usaco-740.mdx [hunk 8]
```diff
@@ -2,12 +2,154 @@
...
+			if (i == j) {continue;}
+			PCL other_pcl = pcl_list[j];
+			if (is_inside(current_pcl, other_pcl)) {
+				valid_pcl = false;
+				break;
+			}
+		}
+		if (valid_pcl) {
+			pcl_count++;
+		}
```
SansPapyrus683, line 145:
```suggestion
			pcl_count += valid_pcl;
```

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,154 @@
...
+
+	floodfill(i + 1, j, color);
+	floodfill(i - 1, j, color);
+	floodfill(i, j + 1, color);
+	floodfill(i, j - 1, color);
+}
+
+// Check whether a given region is a PCL
+bool is_pcl(int i1, int j1, int i2, int j2) {
+
```
SansPapyrus683, line 49:
```suggestion
```

solutions/usaco-740.mdx
```diff
@@ -2,12 +2,154 @@
...
+<CPPSection>
+```cpp
+#include <bits/stdc++.h> 
+using namespace std;
+
+const int MAX_N = 20;
+int n;
+vector<vector<char>> image(MAX_N, vector<char>(MAX_N));
+vector<vector<bool>> visited(MAX_N, vector<bool>(MAX_N));
+int pcl_count = 0;
```
SansPapyrus683, line 21: n & pcl_count don't need to be global

SansPapyrus683 (changes requested):
almost there

solutions/usaco-740.mdx
pranavgithub1: Do you think we should leave it as 
```cpp
if (region_count[i] == 1) {
	color_with_one_region = true;
}
if (region_count[i] > 1) {
	color_with_more_regions = true;
}
```
Since it is more immediately apparent what it does (for me at least) than the shortened expression
`color_with_more_regions = color_with_more_regions || region_count[i] >= 1;`

SansPapyrus683: wrong segment of code lol

pranavgithub1: yeah its not showing me a comment prompt underneath that one, idk why. Its grouping these together

SansPapyrus683: hm, yeah you can leave this as is (but don't resolve it), let's wait for a third opinion

pranavgithub1:
The last commit has all the requested changes except the one we need a third opinion on.

HarryWangATX (approved):

solutions/usaco-740.mdx
HarryWangATX: i think its okay either way

SansPapyrus683: harry is this about this suggestion or the one we're talking about

HarryWangATX: both, but ig it could be more clear if u still add the if statements?

SansPapyrus683: thought about it, the first one should be just formatted as one if statement

pranavgithub1: > thought about it, the first one should be just formatted as one if statement

Do you mean like this?
```cpp
if (region_count[i]  >= 1) {
	color_with_one_region = true;
        color_with_more_regions = true;
}
```
Because that wouldn't work since for the condition you need two colors : 
 - one color has exactly one region
 - another color has 2 or more regions
 
 so with this way, a color with 2 or more regions would set both flags to true which could falsely trigger the condition.

SansPapyrus683: OH
THE VARIABLES ARE DIFFERENT
SO SORRY LMAO

SansPapyrus683 (approved):

SansPapyrus683:
two mods, should be good enough

PR #2594, opened 2022-02-24
Groups Templating System

MrinallU:
- Implement Template System
- Format 7ca952a
- Implement Template System
- Remove function
- merge content changes
- feat: individual problem templating system

- [x] I have tested my code.


MrinallU:
@maggie-j-liu So the individual problem templating works now, I won't be able to work on the full group template becuase I am very busy this week.

maggie-j-liu (approved):
i think this looks good, @thecodingwizard do you want to review?

src/components/Groups/GroupPage/FeedItem.tsx
```diff
@@ -77,11 +92,121 @@ export default function FeedItem({
...
+      setMap(new Map(map.set(g.id, new MapData(true, g))));
+      console.log(map.get(g.id).used);
+    }
+    console.log(map.size);
+  }
+
+  async function exportSelectedPosts() {
+    const type = post.type;
+    const defaultPost: Omit<PostData, 'timestamp'> = {
+      name: post.name,
```
thecodingwizard, line 144: we can potentially load the problems from firebase in this function

thecodingwizard (changes requested):

MrinallU:
@maggie-j-liu @thecodingwizard I believe this should be fine?

thecodingwizard (changes requested):
I think we're getting there! However, to help make it easier to maintain in the future:

- The code necessary for the modal, export the post, etc is quite significant, and probably doesn't belong in `FeedItem.tsx`
- We're loading all the problems for the post in `FeedItem.tsx`. This is quite inefficient, because when we go to the group page, we have one feed item for every post (and there can be a significant number of posts). Then, for every post we're loading every problem for that post. This means that when we go to the group page we end up loading every single problem for that group (which is a lot!) We don't do anything with the problems unless the user decides to export the post (which is a rare action), so we shouldn't load the problems until we actually have to.

Here's one way to address these issues:

1. Create a new file just for the modal, `PostExportModal.tsx` or something. This exports a component for the modal that accepts a limited number of props (such as `isOpen`, `onClose`, `groupID`, `postID` etc). 
2. Move the modal JSX as well as the logic for exporting the problems into `PostExportModal.tsx`. The only new code added to `FeedItem.tsx` for this feature should be to render the PostExportModal.tsx, show a "Export Post" button in the dropdown, and a `isExportPostModalOpen` state needed to control whether the post export modal is open. Everything else goes to `PostExportModal.tsx`. These two steps helps address issue #1 above.
3. In `PostExportModal.tsx`, instead of having a `useEffect` that subscribes to problems, move the code to fetch the problems for the post into the `exportSelectedPosts` function. That way, the problems are only fetched when the user actually decides to export the posts (or at least when the user opens the export post modal), not before.

Let me know if anything doesn't make sense, and thanks for working on this issue!

MrinallU:
> I think we're getting there! However, to help make it easier to maintain in the future:
> 
> * The code necessary for the modal, export the post, etc is quite significant, and probably doesn't belong in `FeedItem.tsx`
> * We're loading all the problems for the post in `FeedItem.tsx`. This is quite inefficient, because when we go to the group page, we have one feed item for every post (and there can be a significant number of posts). Then, for every post we're loading every problem for that post. This means that when we go to the group page we end up loading every single problem for that group (which is a lot!) We don't do anything with the problems unless the user decides to export the post (which is a rare action), so we shouldn't load the problems until we actually have to.
> 
> Here's one way to address these issues:
> 
> 1. Create a new file just for the modal, `PostExportModal.tsx` or something. This exports a component for the modal that accepts a limited number of props (such as `isOpen`, `onClose`, `groupID`, `postID` etc).
> 2. Move the modal JSX as well as the logic for exporting the problems into `PostExportModal.tsx`. The only new code added to `FeedItem.tsx` for this feature should be to render the PostExportModal.tsx, show a "Export Post" button in the dropdown, and a `isExportPostModalOpen` state needed to control whether the post export modal is open. Everything else goes to `PostExportModal.tsx`. These two steps helps address issue #1 above.
> 3. In `PostExportModal.tsx`, instead of having a `useEffect` that subscribes to problems, move the code to fetch the problems for the post into the `exportSelectedPosts` function. That way, the problems are only fetched when the user actually decides to export the posts (or at least when the user opens the export post modal), not before.
> 
> Let me know if anything doesn't make sense, and thanks for working on this issue!

Got it, thank you for the feedback!

thecodingwizard (changes requested):
Were you able to address step 3 above?

> In PostExportModal.tsx, instead of having a useEffect that subscribes to problems, move the code to fetch the problems for the post into the exportSelectedPosts function. That way, the problems are only fetched when the user actually decides to export the posts (or at least when the user opens the export post modal), not before.

it seems like problems are still being loaded on every render

PR #2597, opened 2022-02-24
Add a √n log n time complexity for n ≤ 10¹²

megargayu:
I was recently looking at some previous USACO Silver problems and stumbled upon [2020 January Contest - Loan Repayment](http://www.usaco.org/index.php?page=viewproblem2&cpid=991), which uses the limit <code>1 ≤ N ≤ 10<sup>12</sup></code>. According to [the editorial](http://www.usaco.org/current/data/sol_loan_silver_jan20.html):
> As the numbers in the statement are up to <code>10<sup>12</sup></code>, not <code>10<sup>18</sup></code>, **this suggests that some sort of √N factor is involved.**

It turns out that the time complexity of the intended solution to that problem is √N log N (using binary search on a function with an O(√N) time complexity).

However, on the [Time Complexity](https://usaco.guide/bronze/time-comp?lang=cpp) page, there is no mention of a <code>10<sup>12</sup></code> time complexity. I feel this will be helpful to include in the chart (even if this might be rare), so here is the pull request! I am not sure if it should be <code>10<sup>12</sup></code> or some higher value, though - I'm just going off of the problem I mentioned above.

- [x] I have tested my code.

### DIFF
--- a/content/2_Bronze/Time_Comp.mdx
+++ b/content/2_Bronze/Time_Comp.mdx
@@ -418,6 +418,7 @@ check whether an algorithm is viable.
 | $n \le 7 \cdot 10^4$ | $\mathcal{O}(n \sqrt n)$                                         |
 | $n \le 5 \cdot 10^5$ | $\mathcal{O}(n \log n)$                                          |
 | $n \le 5 \cdot 10^6$ | $\mathcal{O}(n)$                                                 |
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
imo this complexity is too niche to be put on the graph, but idk lol

andrewwangva:
I agree with @SansPapyrus683. At this point, you shouldn't be memorizing sqrt(n)log(n) and instead be able to estimate the amount of operations/time it takes for niche complexities.

PR #2713, opened 2022-04-28
Fix semicolon and uninitialized variable

SnowballSH:
In the Knuth–Morris–Pratt algorithm C++ solution for "String Matching", the code does not work due to:
1. there is a missing semicolon which is an syntax error
2. variable `j` is uninitialized which causes segmentation error on most machines I've tested on.

This small fix makes the code snippet work.

- [x] I have tested my code.


ryanchou-dev (approved):
sure ok

PR #2751, opened 2022-05-10
add python solution for snow boots, silver (811)

jjliewie:

- [x] I have tested my code.

### DIFF
--- a/solutions/usaco-811.mdx
+++ b/solutions/usaco-811.mdx
@@ -261,7 +261,7 @@ for i in range(n):
 for i in range(b):
 	if stor[-1][i]:
 		print(i, file=open('snowboots.out', 'w'))
-		exit()
+		break
 ```
 
 </PySection>
--- a/solutions/usaco-811.mdx
+++ b/solutions/usaco-811.mdx
@@ -2,7 +2,7 @@
 id: usaco-811
 source: USACO Silver 2018 February
 title: Snow Boots
-author: Qi Wang, Ryan Chou
+author: Qi Wang, Ryan Chou, Juheon Rhee
 ---
 
 [Official Analysis (C++)](http://www.usaco.org/current/data/sol_snowboots_silver_feb18.html)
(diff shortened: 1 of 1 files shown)


solutions/usaco-811.mdx
```diff
@@ -229,4 +229,38 @@ public class snowboot {
...
+
+<PySection>
+
+```python
+with open('snowboots.in') as r:
+	n, b = map(int, r.readline().split())
+	depth = list(map(int, r.readline().split()))
+	max_depth = [[*map(int, r.readline().split())] for _ in range(b)]
+
+stor = [[0] * b for _ in range(n)]
```
SansPapyrus683, line 241: better name pls

SansPapyrus683 (approved):

solutions/usaco-811.mdx
jjliewie: stor as in storage?

jjliewie: or are you talking about depth and max_depth?

SansPapyrus683: `stor`
bc like what's it storing?

jjliewie: so should i change it to dp?

SansPapyrus683: no like the point is that name isn't really informative about what it's storing

jjliewie: oh okay i'll change it.

ryanchou-dev (approved):
⭐

solutions/usaco-811.mdx
SansPapyrus683: bump

solutions/usaco-811.mdx
```diff
@@ -229,4 +229,41 @@ public class snowboot {
...
+
+	for j in range(cur + 1, b):
+		s, d = max_depth[j]
+		if s >= depth[i]:
+			stor[i][j] = 1
+
+for i in range(b):
+	if stor[-1][i]:
+		print(i, file=open('snowboots.out', 'w'))
+		exit()
```
chezbgone, line 264: `break` instead of `exit`ing

solutions/usaco-811.mdx
```diff
@@ -229,4 +229,41 @@ public class snowboot {
...
+```python
+with open('snowboots.in') as r:
+	n, b = map(int, r.readline().split())
+	depth = list(map(int, r.readline().split()))
+	max_depth = [[*map(int, r.readline().split())] for _ in range(b)]
+
+stor = [[0] * b for _ in range(n)]
+stor[0][0] = 1
+for i in range(n):
+	cur = -1
```
chezbgone, line 244: It's more clear if we use `None` instead of `-1` here, since `None` explicitly conveys that there is nothing found yet.

It may be true that `cur` can never be negative, but this is an extra step of reasoning someone has to make in order to understand the code.

jjliewie: could I just explain this in a comment? in line 256, it's simpler and easier if cur is set it -1, rather than None.

chezbgone: I missed how `cur` was being used later when reading the code. I think the original code is fine now that I realize this.

You can add a comment if you want. I think it's fine either way.

PR #2814, opened 2022-05-30
quiz codeblock fix

maggie-j-liu:
fixes an issue where the intro-ds page would glitch upon reload (thanks @cirex-web for reporting)

basically the dom nesting was invalid because of nested `buttons`, so i just removed the copy button when the right answer hasn't been selected yet


SansPapyrus683:
how the hell has checking not finished yet

PR #2837, opened 2022-06-14
fix type annotations and convention ii

SansPapyrus683:

- [x] I have tested my code.

i was reading [this type annotation PEP](https://peps.python.org/pep-0585/) and realized i've been getting type annotations wrong
so i ctrl f'd and tried to fix all of them

### DIFF
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -92,10 +92,7 @@ public static int findLis(int[] a) {
 <PySection>
 
 ```py
-from typing import List
-
-
-def find_lis(a: List[int]) -> int:
+def find_lis(a: list[int]) -> int:
 	lis = 0
 	dp = [1] * len(a)
 	for i in range(len(a)):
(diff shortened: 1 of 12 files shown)


ryanchou-dev (approved):
⭐

content/4_Gold/LIS.mdx
```diff
@@ -92,10 +92,7 @@ public static int findLis(int[] a) {
 <PySection>
 
 ```py
-from typing import List
-
-
-def find_lis(a: List[int]) -> int:
+def find_lis(a: list[int]) -> int:
```
bqi343, line 95: this doesn't compile in python 3.8 and lower (https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc) so maybe just leave it as is? note that CF currently uses python 3.8 and USACO uses python 3.6

SansPapyrus683: alr, everything should be back to using `typing` now

SansPapyrus683:
OH NO FORGOT THIS EXISTED

envyaims (approved):
code seems fine

solutions/usaco-859.mdx
```diff
@@ -28,46 +29,172 @@ int main() {
...
+		}
+
+		// sort by arrival time
+		Arrays.sort(cows, Comparator.comparingInt(c -> c.start));
+	
+		int time = 0;
+		int curr = 0;
+		int longestWait = 0;
+
+		// sorted by priority so that the highest seniority starts eating first
```
envyaims, line 106:
```suggestion
		// sorted so that the highest seniority starts eating first
```
idk

Issue #2831, opened 2022-06-10
Groups: back button after creating a new group has weird behavior


maggie-j-liu:
After creating a new group, you'd expect the "back" button to behave like a cancel button:

![image](https://user-images.githubusercontent.com/63619830/173161121-54e6313d-2d7d-487b-9ac0-0a3a7a514766.png)

Since the group was created before that page comes up, and it's actually just the edit page, the back button doesn't cancel the creation of the group or delete the group. We could make a different modal for creating groups, or just remove the button.

Also there are 2 save buttons...

MrinallU:
This seems pretty easy so I'll do it

cirex-web:
Did you start this yet? If not, could I do it? @MrinallU

MrinallU:
> Did you start this yet? If not, could I do it? @MrinallU 

I'm almost done with it.

MrinallU:
@cirex-web Can you do #2817 instead?

cirex-web:
@MrinallU hmm I'll think about it

MrinallU:
Resolved by #2868

PR #2868, opened 2022-06-22
additional group fixes

MrinallU:

- [ ] I have tested my code.


thecodingwizard:
wait what's the point of making a dedicated create group page?

(also it seems like most of the code is equivalent to the edit group page?)

MrinallU:
> wait what's the point of making a dedicated create group page?
> 
> (also it seems like most of the code is equivalent to the edit group page?)

I just removed the back button as well as the duplicate save buttons in the latest commit, does this work?

thecodingwizard:
uhhh I personally thought the duplicate buttons were fine (it was an intentional choice so that if the page got really long you could click the save button at the very bottom without scrolling back up to the top). but I'm also fine with removing it

MrinallU:
I was told in #2831 to remove it, but if you wanted it that's fine with me.

thecodingwizard:
ah. okay considering both save buttons are almost always visible at the same time I think this change makes sense. Thanks!

MrinallU:
@thecodingwizard Should I remove the back button as well?

thecodingwizard:
Only when making a new group. I think it should stay when you're editing an existing group

PR #2947, opened 2022-07-20
Update Vectorization.mdx (Pragmas actually work for USACO)

Aryansh-S:
It turns out pragmas do work for USACO as mentioned in [a recent editorial](http://www.usaco.org/current/data/sol_prob3_silver_open22.html). Updated this module accordingly.


envyaims (approved):
sus, but still state you cant prove they work for the most part though

ryanchou-dev (approved):
...suree

jessechoe10 (approved):

Aryansh-S:
I think we should be able to merge this.

nwatx (approved):

PR #2982, opened 2022-08-07
Rewrite CSES Flight Routes Check Proof

izhang05:
I think writing the proof without the unnecessary contradiction makes it a lot clearer and easier to read.

- [ ] I have tested my code.

### DIFF
--- a/solutions/cses-1682.mdx
+++ b/solutions/cses-1682.mdx
@@ -27,14 +27,14 @@ $1$ $x$ if you're running the DFS on $G$ and $x$ $1$ otherwise.
 
 # Proof
 
-Let's do a proof by contradiction. Assume that $\texttt{can[1][x]}$ is true for
-all vertices $x$ in both $G$ and $G'$, and there exists a pair of vertices
-$u, v$ such that $\texttt{can[u][v]} = \texttt{false}$. Since
-$\texttt{can[1][u]}$ is true in $G'$, then $\texttt{can[u][1]}$ must be true in
-$G$. Additionally, $\texttt{can[1][v]}$ must be true in $G$. So, you can travel
-from $u \rightarrow 1 \rightarrow v$, which contradicts the statement that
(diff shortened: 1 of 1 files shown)


solutions/cses-1682.mdx
```diff
@@ -27,14 +27,14 @@ $1$ $x$ if you're running the DFS on $G$ and $x$ $1$ otherwise.
...
 
-Let's do a proof by contradiction. Assume that $\texttt{can[1][x]}$ is true for
-all vertices $x$ in both $G$ and $G'$, and there exists a pair of vertices
-$u, v$ such that $\texttt{can[u][v]} = \texttt{false}$. Since
-$\texttt{can[1][u]}$ is true in $G'$, then $\texttt{can[u][1]}$ must be true in
-$G$. Additionally, $\texttt{can[1][v]}$ must be true in $G$. So, you can travel
-from $u \rightarrow 1 \rightarrow v$, which contradicts the statement that
-$\texttt{can[u][v]} = \texttt{false}$. Thus, $\texttt{can[u][v]}$ is true for
-all vertices $u, v$.
+We will show that if $\texttt{can[1][x]}$ for $1 \leq x \leq n$ in both $G$ and $G'$,
```
SansPapyrus683, line 30: don't think it's good to abuse latex
can you explain this in layman terms?

izhang05: Yeah I agree, I just used that because it's the notation used in the first half of the solution.
Should I also rewrite the first part of the solution then?

chezbgone:
From Discord discussion: Is the proof even necessary? I feel like adding an extra sentence to the first line of the "Main Idea" section would be better. The sentence can be something like "This is because if we want to find a path `u -> v` (u \to v), we can compose the two paths `u -> x` and `x -> v`.

izhang05:
Yeah, in that case I'm not sure we need the can[u][v] either. It would probably be clearer to say something like "If for every node $x$, there is a path from $1$ to $x$ and a path from $x$ to $1$..."

PR #2985, opened 2022-08-08
Improve consistency and add java implementation for MortalKombatTower (CF1418C)

ChuangSheep:
fix: rename overriden var i in cpp
style: make variable naming in expl and cpp code consistent
feat: add java implementation

--- 

- [x] I have tested my code.

### DIFF
--- a/solutions/cf-1418C.mdx
+++ b/solutions/cf-1418C.mdx
@@ -2,27 +2,27 @@
 id: cf-1418C
 source: CF
 title: Mortal Kombat Tower
-author: Ryan Chou
+author: Ryan Chou, Chuyang Wang
 ---
 [Official Editorial (C++)](https://codeforces.com/blog/entry/82673)
 
 ## Alternate Solution - DP
 ### Explanation
 
-If we define $\texttt{dp}[i][j]$ as the minimum number of skip points needed to get to the $i$th boss on turn $j$ 
+If we define $\texttt{dp}[i][j]$ as the minimum number of skip points needed to get to the $j$th boss on player $i$'s turn 
 (our turn is zero), then our transitions would be:
 
-$\texttt{dp}[i + 1][0] =\min(\texttt{dp}[i + 1][0], \texttt{dp}[i][1] + \texttt{hard}[i])$
+$\texttt{dp}[0][j + 1] =\min(\texttt{dp}[0][j + 1], \texttt{dp}[1][j] + \texttt{hard}[j])$
 
-$\texttt{dp}[i + 2][0] =\min(\texttt{dp}[i + 2][0], \texttt{dp}[i][1] + \texttt{hard}[i] + \texttt{hard}[i + 1])$
+$\texttt{dp}[0][j + 2] =\min(\texttt{dp}[0][j + 2], \texttt{dp}[1][j] + \texttt{hard}[j] + \texttt{hard}[j + 1])$
 
-$\texttt{dp}[i + 1][1] =\min(\texttt{dp}[i + 1][1], \texttt{dp}[i][0])$
+$\texttt{dp}[1][j + 1] =\min(\texttt{dp}[1][j + 1], \texttt{dp}[0][j])$
 
-$\texttt{dp}[i + 2][1] =\min(\texttt{dp}[i + 2][1], \texttt{dp}[i][0])$
+$\texttt{dp}[1][j + 2] =\min(\texttt{dp}[1][j + 2], \texttt{dp}[0][j])$
 
 This is because from each state, either the previous move, or the previous two moves could have been from the other player.
 
-So, the answer would be $\min(\texttt{dp}[N][0], \texttt{dp}[N][1])$.
+So, the answer would be $\min(\texttt{dp}[0][N], \texttt{dp}[1][N])$.
 
 ### Implementation
 
@@ -43,9 +43,9 @@ int main() {
 		int n;
 		cin >> n;
 
-		vector<int> bosses(n);
-		for (int i = 0; i < n; i++) {
-			cin >> bosses[i];
+		vector<int> hard(n);
+		for (int j = 0; j < n; j++) {
+			cin >> hard[j];
 		}
 
 		/*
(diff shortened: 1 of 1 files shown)


solutions/cf-1418C.mdx
```diff
@@ -43,9 +43,9 @@ int main() {
 		int n;
 		cin >> n;
 
-		vector<int> bosses(n);
-		for (int i = 0; i < n; i++) {
-			cin >> bosses[i];
+		vector<int> hard(n);
+		for (int j = 0; j < n; j++) {
+			cin >> hard[j];
```
SansPapyrus683, line 48: i'd rename everything to `bosses` instead of `hard`

solutions/cf-1418C.mdx
```diff
@@ -2,27 +2,27 @@
...
-author: Ryan Chou
+author: Ryan Chou, Chuyang Wang
 ---
 [Official Editorial (C++)](https://codeforces.com/blog/entry/82673)
 
 ## Alternate Solution - DP
 ### Explanation
 
-If we define $\texttt{dp}[i][j]$ as the minimum number of skip points needed to get to the $i$th boss on turn $j$ 
+If we define $\texttt{dp}[i][j]$ as the minimum number of skip points needed to get to the $j$th boss on player $i$'s turn
```
SansPapyrus683, line 12: why swap the two?

solutions/cf-1418C.mdx
```diff
@@ -61,20 +61,68 @@ int main() {
...
+
+			/*
+			 * dp[i][j] = the min amount of skip points needed if it is
+			 * player i's turn (you are 0 and your friend 1) on the j-th boss.
+			 * At the beginning (before fighting any bosses), dp[1][0] = 0.
+			 */
+			int[][] dp = new int[2][n + 1];
+			for (int j = 0; j <= n; j++) {
+				dp[0][j] = 1000000000;
+				dp[1][j] = 1000000000;
```
SansPapyrus683, line 105: `Integer.MAX_VALUE`

SansPapyrus683 (changes requested):

solutions/cf-1418C.mdx
ChuangSheep: cause it was the other way around in the code. I either have to write $\texttt{dp}[j][i]$ or swap the definition of those two vars, or, change the cpp code completely to make it a dp[n][2] array (it was dp[2][n]).

solutions/cf-1418C.mdx
ChuangSheep: will lead to overflow. I can add a constant INTMAX=1000000000 if that's better.

SansPapyrus683: if you're doing overflow just check if you're adding INT MAX

solutions/cf-1418C.mdx
SansPapyrus683: fair enough

solutions/cf-1418C.mdx
ChuangSheep: sry u mean i *should* use the constant?

SansPapyrus683: no i mean
use `Integer.MAX_VALUE`
you shouldn't really be adding those invalid values anyways, so it's best to add a check

ChuangSheep: So i should do 
```
dp[0][j + 2] = Math.min(dp[0][j + 2], dp[1][j] == Integer.MAX_VALUE ? Integer.MAX_VALUE :  dp[1][j] + bosses[j] + bosses[j + 1]);
```
?

SansPapyrus683: if it's that long i'd just use an if statement

ChuangSheep: is this acceptable `dp[0][j] = Integer.MAX_VALUE - 3` ?

At this point, I *don't* think this is more clear but if it's preferred..
```
if (dp[1][j] < Integer.MAX_VALUE) {
    dp[0][j + 2] = Math.min(dp[0][j + 2], dp[1][j] + bosses[j] + bosses[j + 1]);
}
```

SansPapyrus683: why the -3?
if it's MAX, you shouldn't be adding it in the first place, no?

ChuangSheep: actually, mathematically, adding a number to $\infty$ still lead to infinity. In our case, infinity means that it is infeasible to use this state (we need infinity points), and therefore the next state based on this should also be infeasible, i.e. valued with infinity. We actually don't really want the MAX VALUE of a data type but rather the idea of infinity. Using if statement is one way to simulate this (adding to inf is inf), or, use a sightly smaller val than MAX VALUE, which makes it to *behave* like inf since one can add to it and it's still larger than a certain line, from which we call a number infinity. 

As said, I don't think adding two if statements (add thereby more nesting levels) make the code any easier to read. But if u insist, I don't really have a big problem with that either.

SansPapyrus683: then set it to infinity if one of the values is infinity?

ChuangSheep: so `static final int INF = 1000000000;`?

SansPapyrus683: no, just use `Integer.MAX_VALUE` and do checks

ChuangSheep: alright will do

envyaims (approved):
erm ok even though its kind of painful to watch you guys used $i$ for test case iteration

SansPapyrus683:
oh yeah, OP, i'd use `t` for that (for this i don't really care, up to you)

solutions/cf-1418C.mdx
```diff
@@ -84,12 +84,11 @@ import java.io.*;
 import java.util.Arrays;
 
 public class MortalKombatTower {
-	public static final int INTMAX = 1000000000;
 	public static void main(String[] args) throws IOException {
 		BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
 		int t = Integer.parseInt(br.readLine());
 
-		for (int i = 0; i < t; i++) {
+		for (int testcase = 0; testcase < t; testcase++) {
```
SansPapyrus683, line 91: Uh please do t not test_case

SansPapyrus683 (approved):

solutions/cf-1418C.mdx
ChuangSheep: well t is given in the problem statement as a variable. How should I name this variable then if I use t as ite var?

SansPapyrus683: `test_case_num` as the actual variable and `t` for the iteration

Aadit-Ambadkar: I'd prefer `while (t-->0)`, without spaces as I think the `-->` looks like an arrow, but `for (; t > 0; t--)` is probably the most clear.

solutions/cf-1418C.mdx
```diff
@@ -61,20 +61,72 @@ int main() {
...
+			for (int j = 0; j <= n; j++) {
+				dp[0][j] = Integer.MAX_VALUE;
+				dp[1][j] = Integer.MAX_VALUE;
+			}
+			dp[1][0] = 0;
+
+			for (int j = 0; j < n; j++) {
+				// the opposite player switches on the previous move.
+				if (dp[1][j] < Integer.MAX_VALUE) {
+					dp[0][j + 1] = Math.min(dp[0][j + 1], dp[1][j] + bosses[j]);
```
Aadit-Ambadkar, line 112: 81 chars, i think you should be fine though

solutions/cf-1418C.mdx
SansPapyrus683: nah fam

PR #3038, opened 2022-09-10
Refactor usaco 619 cpp solution

RmdanJr:
rewrite the pos_of_interest function in a cleaner way.

- [x] I have tested my code.

### DIFF
--- a/solutions/usaco-619.mdx
+++ b/solutions/usaco-619.mdx
@@ -259,12 +259,16 @@ constant). Similar reasoning holds for the $y$-coordinates.
 <CPPSection>
 
 ```cpp
-#include <bits/stdc++.h>
-using namespace std;
+#include <iostream>
+#include <fstream>
+#include <vector>
+#include <algorithm>
 
-struct Cow {
-	int x, y;
-};
+using std::cout;
+using std::endl;
+using std::vector;
+
+struct Cow { int x, y; };
 
 /** Returns the list of *important* coordinates */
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
don't feel like this is necessary, how is it cleaner?

RmdanJr:
@SansPapyrus683 
The function returns the list of *important* coordinates. The current implementation loop through all coordinates and stores distinct ones in `distinct` vector and their indices in `num_left` vector. Then loop through all distinct elements in `distinct` vector and add the element to returned `important` vector only if its index in the original `pos` array is between `pos.size() / 3` and `2 * pos.size() / 3`.
Why all this if we could loop through elements in `pos` from `pos.size() / 3` and `2 * pos.size() / 3` and add an element to returned `important` vector if it's distinct. This is what I did. I think this is more cleaner, has fewer steps, and is easy to understand.

SansPapyrus683:
cool, but why change all the other things like the `using`s?

RmdanJr:
I just changed them to make the code shorter. But improving `the pos_of_interest()` function was my main goal.

SansPapyrus683:
can you revert those other changes?

RmdanJr:
Sure. Reverted.

bqi343:
Actually, this solution is wrong; it outputs 3 on this input:

```
3
1 7
5 1
7 9
```

Will revert to the previous solution.

PR #3086, opened 2022-09-25
Added problem and solution for cf-1713E

Keys9873:

- [_] I have tested my code.
- [_] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [_] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions), which include the following: 

### DIFF
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -1,3 +1,14 @@
+---
+id: cf-1713E
+source: Codeforces Div 2
+title: Cross Swapping
+author: Aakash Gokhale
+---
+
+## Solution
+
+[Official Analysis](https://codeforces.com/blog/entry/105419)
+
 Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operations are performed.
 
 In addition, we want to lexografically sort the array. Thus, we can iterate and greedily process the array.
@@ -8,7 +19,7 @@ If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to 
 
 However, it is possible that not all conditions can work together with each other and swapping something means we can't swap something else later on.
 
-We can use the Amogus Trick to represent this problem. Create a disjoint set union of $2n$ nodes where node $i$ represents the state where the operation is performed on $i$ and $i + n$ represents the state where operation is not performed on $i$.
+We can use the Amogus Trick] o represent this problem. Create a disjoint set union of $2n$ nodes where node $i$ represents the state where the operation is performed on $i$ and $i + n$ represents the state where operation is not performed on $i$.
 
 As we are iterating greedily through the array, we can merge ($i$ and $j$) and ($i + n$ and $j + n$) if we don't want to swap. This works because if we swap neither or swap both the arrangement remains the same.
 If we do want to swap we can merge ($i$ and $j + n$) and ($i + n$ and $j$). This works because if we swap exactly one of these, then the arrangement will change. 
@@ -16,3 +27,105 @@ If we do want to swap we can merge ($i$ and $j + n$) and ($i + n$ and $j$). This
 After this we can construct the answer.
 
 We can iterate over the entire grid. If it is a diaganal it will never change. If $a[i][j] == a[j][i]$ then we can manually set them to normal (since we didn't merge them in the disjoint set union when comparing them the states might have not been merged). If it is not a diaganol we can check whether $i$ and $j$ are merged in the disjoint set union. If they aren't then we will swap $a[i][j]$ and $a[j][i]$. Otherwise we will keep them the same.
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(N^2)$
+
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+using vi = vector<int>;
+
+struct DSU {
+  vi e;
+  DSU(int n) {
+    e = vi(n, -1);
+  }
+  int get(int x) {
+    return e[x] < 0 ? x : e[x] = get(e[x]);
+  }
+  bool same(int a, int b) {
+    return get(a) == get(b);
+  }
+  int size(int x) {
+    return -e[get(x)];
+  }
+  bool unite(int x, int y) {
+    x = get(x); y = get(y);
+    if (x == y) return false;
+    if (e[x] > e[y]) swap(x, y);
+    e[x] += e[y]; e[y] = x;
+    return true;
+  }
+};
+
+void solve() {
+  int n;
+  cin >> n;
+  int a[n][n];
+  for (int i = 0; i < n; i++)
+    for (int j = 0; j < n; j++) cin >> a[i][j];
+  DSU dsu(2 * n);
+  for (int i = 0; i < n; i++) {
+    for (int j = i + 1; j < n; j++) { // Only check each pair once and diagnals don't matter
+      if (a[i][j] == a[j][i]) continue; // If they are the same we should not merge either way because it could affect later merges
+      else if (a[i][j] > a[j][i]) {
+        // Try to swap the elements
+        if (dsu.get(i) != dsu.get(j)) { // Check if the states are already merged the opposite way
+          dsu.unite(i, j + n);
+          dsu.unite(j, i + n);
+        }
+      } else  {
+        // Try to keep the elements the same
+        if (dsu.get(i) != dsu.get(j + n)) { // Check if the states are already merged the opposite way
+          dsu.unite(i, j);
+          dsu.unite(i + n, j + n);
+        }
+      }
+    }
+  }
+  // Construct the answer
+  int ans[n][n];
+  for (int i = 0; i < n; i++) {
+    for (int j = i; j < n; j++) {
+      // If a[i][j] is part of a diagonal or the pair is equal (a[i][j] == a[j][i]), manually set it to be the same as the original array
+      if (i == j) ans[i][j] = a[i][j];
+      else if (a[i][j] == a[j][i]) ans[i][j] = ans[j][i] = a[i][j];
+      else { // Otherwise check whether the i and j states are merged
+        if (dsu.get(i) == dsu.get(j)) {
+          ans[i][j] = a[i][j];
+          ans[j][i] = a[j][i];
+        } else {
+          ans[i][j] = a[j][i];
+          ans[j][i] = a[i][j];
+        }
+      }
+    }
+  }
+  // Print the answer
+  for (int i = 0; i < n; i++) {
+    for (int j = 0; j < n; j++) {
+      cout << ans[i][j] << " ";
+    }
+    cout << endl;
+  }
+}
+
+int main() { 
+  ios::sync_with_stdio(false); cin.tie(nullptr);
+  int T = 1; cin >> T;
+  while (T--) {
+    solve();
+  }
+}
+```
+
+</CPPSection>
+
+</LanguageSection>
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -82,7 +82,9 @@ void solve() {
 			cin >> a[i][j];
 		}
 	}
+	
 	DSU dsu(2 * n);
+	
 	for (int i = 0; i < n; i++) {
 		// Only check each pair once and diagnals don't matter
 		for (int j = i + 1; j < n; j++) {
@@ -105,6 +107,7 @@ void solve() {
 			}
 		}
 	}
+	
 	// Construct the answer
 	int ans[n][n];
 	for (int i = 0; i < n; i++) {
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -13,7 +13,7 @@ Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operati
 
 In addition, we want to lexicographically sort the array. Thus, we can iterate and greedily process the array.
 
-For all j$ such that $i < j$ (we only need to consider each case once), we can determine whether we want to swap them or keep them the same.
+For all $j$ such that $i < j$ (we only need to consider each case once), we can determine whether we want to swap them or keep them the same.
 
 If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to keep them the same. If they are equal it doesn't matter.
 
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -5,7 +5,7 @@ title: Cross Swapping
 author: Aakash Gokhale
 ---
 
-## Solution
+## Explanation
 
 [Official Analysis](https://codeforces.com/blog/entry/105419)
 
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -128,7 +128,7 @@ void solve() {
 				ans[i][j] = a[i][j];
 			} else if (a[i][j] == a[j][i]) {
 				ans[i][j] = ans[j][i] = a[i][j];
-			} else { // Otherwise check whether the i and j states are merged
+			} else {  // Otherwise check whether the i and j states are merged
 				if (dsu.get(i) == dsu.get(j)) {
 					ans[i][j] = a[i][j];
 					ans[j][i] = a[j][i];
@@ -149,7 +149,8 @@ void solve() {
 }
 
 int main() { 
-	ios::sync_with_stdio(false); cin.tie(nullptr);
+	ios::sync_with_stdio(false);
+	cin.tie(nullptr);
 	int T;
 	cin >> T;
 	for (int t = 0; t < T; t++) {
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -117,7 +117,7 @@ void solve() {
 	}
 	
 	// Construct the answer
-	int ans[n][n];
+	vector<vi> ans(n, vi(n));
 	for (int i = 0; i < n; i++) {
 		for (int j = i; j < n; j++) {
 			/*
@@ -144,7 +144,7 @@ void solve() {
 		for (int j = 0; j < n; j++) {
 			cout << ans[i][j] << " ";
 		}
-		cout << endl;
+		cout << '\n';
 	}
 }
 
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -42,7 +42,7 @@ using namespace std;
 
 using vi = vector<int>;
 
-//BeginCodeSnip{}
+//BeginCodeSnip{DSU}
 struct DSU {
 	vi e;
 	DSU(int n) {
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -90,8 +90,9 @@ void solve() {
 		for (int j = i + 1; j < n; j++) {
 			// If they are the same we should not merge either way
 			// because it could affect later merges
-			if (a[i][j] == a[j][i]) continue;
-			else if (a[i][j] > a[j][i]) {
+			if (a[i][j] == a[j][i]) {
+				continue;
+			} else if (a[i][j] > a[j][i]) {
 				// Swap the elements if the states are not already merged the opposite way
 				if (dsu.get(i) != dsu.get(j)) {
 					dsu.unite(i, j + n);
@@ -115,9 +116,11 @@ void solve() {
 			/* If a[i][j] is part of a diagonal or the pair is equal (a[i][j] == a[j][i]),
 			 * manually set it to be the same as the original array
 			 */
-			if (i == j) ans[i][j] = a[i][j];
-			else if (a[i][j] == a[j][i]) ans[i][j] = ans[j][i] = a[i][j];
-			else { // Otherwise check whether the i and j states are merged
+			if (i == j) {
+				ans[i][j] = a[i][j];
+			} else if (a[i][j] == a[j][i]) {
+				ans[i][j] = ans[j][i] = a[i][j];
+			} else { // Otherwise check whether the i and j states are merged
 				if (dsu.get(i) == dsu.get(j)) {
 					ans[i][j] = a[i][j];
 					ans[j][i] = a[j][i];
--- a/solutions/cf-1713E.mdx
+++ b/solutions/cf-1713E.mdx
@@ -117,7 +117,7 @@ void solve() {
 	}
 	
 	// Construct the answer
-	int ans[n][n];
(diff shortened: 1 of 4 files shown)


SansPapyrus683:
bro
i said notify me before adding a new problem

Keys9873:
Oh was it on discord? I'm not really active ...

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+## Solution
+
+[Official Analysis](https://codeforces.com/blog/entry/105419)
+
+Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operations are performed.
+
+In addition, we want to lexicographically sort the array. Thus, we can iterate and greedily process the array.
+
+For all $j$ such that $i < j$ (we only need to consider each pair once), we can determine whether we want to swap them or keep them the same.
```
Aadit-Ambadkar, line 16:
```suggestion
For all $j$ such that $i < j$ (we only need to consider each pair once and keeping $i < j$ will ensure this), we can determine whether we want to swap them or keep them the same.
```

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,134 @@
...
+using vi = vector<int>;
+
+struct DSU {
+	vi e;
+	DSU(int n) {
+		e = vi(n, -1);
+	}
+	int get(int x) {
+		return e[x] < 0 ? x : e[x] = get(e[x]);
+	}
```
Aadit-Ambadkar, line 52:
```suggestion
	int get(int x) {
		if (e[x] < 0) return x;
		else {
			e[x] = get(e[x]);
			return e[x];
		}
	}
```

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,134 @@
...
+	for (int i = 0; i < n; i++)
+	  for (int j = 0; j < n; j++) cin >> a[i][j];
+	DSU dsu(2 * n);
+	for (int i = 0; i < n; i++) {
+		for (int j = i + 1; j < n; j++) { // Only check each pair once and diagnals don't matter
+			// If they are the same we should not merge either way because it could affect later merges
+			if (a[i][j] == a[j][i]) continue;
+			else if (a[i][j] > a[j][i]) {
+				// Try to swap the elements
+				if (dsu.get(i) != dsu.get(j)) { // Check if the states are already merged the opposite way
```
Aadit-Ambadkar, line 81: LINE LENGTH EXCEEDS 80 CHARACTERS

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,134 @@
...
+					ans[i][j] = a[i][j];
+					ans[j][i] = a[j][i];
+				} else {
+					ans[i][j] = a[j][i];
+					ans[j][i] = a[i][j];
+				}
+			}
+		}
+	}
+  // Print the answer
```
Aadit-Ambadkar, line 114:
```suggestion
	// Print the answer
```

Aadit-Ambadkar (changes requested):

solutions/cf-1713E.mdx
Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,145 @@
...
+		return true;
+	}
+};
+
+void solve() {
+	int n;
+	cin >> n;
+	int a[n][n];
+	for (int i = 0; i < n; i++)
+	  for (int j = 0; j < n; j++) cin >> a[i][j];
```
SansPapyrus683, line 81: sus

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+In addition, we want to lexicographically sort the array. Thus, we can iterate and greedily process the array.
+
+For all $j$ such that $i < j$ (we only need to consider each pair once and keeping $i < j$ will ensure this), we can determine whether we want to swap them or keep them the same.
+
+If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to keep them the same. If they are equal it doesn't matter.
+
+However, it is possible that not all conditions can work together with each other and swapping something means we can't swap something else later on.
+
+We can use the [Amogus Trick](https://codeforces.com/blog/entry/106407) to represent this problem. Create a disjoint set union of $2n$ nodes where node $i$ represents the state where the operation is performed on $i$ and $i + n$ represents the state where operation is not performed on $i$.
```
SansPapyrus683, line 22: is there no other way to solve this problem?
because if the only way to solve this is with some meme-a** trick, i don't think this is a good problem

SansPapyrus683: or unless you can give a nice intuitive method for arriving at this trick

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,145 @@
...
+				// Keep the elements the same if the states
+				// are not already merged the opposite way
+				if (dsu.get(i) != dsu.get(j + n)) {
+					dsu.unite(i, j);
+					dsu.unite(i + n, j + n);
+				}
+			}
+		}
+	}
+	// Construct the answer
```
SansPapyrus683, line 105: separate relevant code segments w/ whitespace

SansPapyrus683 (changes requested):

solutions/cf-1713E.mdx
SansPapyrus683: braces + tabs

solutions/cf-1713E.mdx
SansPapyrus683: i wouldn't mention the trick then if it isn't really needed

solutions/cf-1713E.mdx
Keys9873: added

solutions/cf-1713E.mdx
Keys9873: fixed

solutions/cf-1713E.mdx
Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,152 @@
...
+	for (int i = 0; i < n; i++) {
+		for (int j = 0; j < n; j++) {
+			cout << ans[i][j] << " ";
+		}
+		cout << endl;
+	}
+}
+
+int main() { 
+	ios::sync_with_stdio(false); cin.tie(nullptr);
```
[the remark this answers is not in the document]
Keys9873: Yes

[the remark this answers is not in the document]
Keys9873: Even if it wasn't necessary, shouldn't you put fast io regardless?

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,152 @@
...
+	
+	// Construct the answer
+	int ans[n][n];
+	for (int i = 0; i < n; i++) {
+		for (int j = i; j < n; j++) {
+			/* If a[i][j] is part of a diagonal or the pair is equal (a[i][j] == a[j][i]),
+			 * manually set it to be the same as the original array
+			 */
+			if (i == j) ans[i][j] = a[i][j];
+			else if (a[i][j] == a[j][i]) ans[i][j] = ans[j][i] = a[i][j];
```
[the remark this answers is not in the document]
Keys9873: fixed

solutions/cf-1713E.mdx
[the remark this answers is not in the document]
SansPapyrus683: ? no?

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,155 @@
...
+using vi = vector<int>;
+
+struct DSU {
+	vi e;
+	DSU(int n) {
+		e = vi(n, -1);
+	}
+	
+	int get(int x) {
+		if (e[x] < 0) return x;
```
SansPapyrus683, line 52: use braces _everywhere_

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,155 @@
...
+
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+using vi = vector<int>;
+
+struct DSU {
```
SansPapyrus683, line 45: also put this in a codesnip or smth

solutions/cf-1713E.mdx
Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -42,6 +42,7 @@ using namespace std;
 
 using vi = vector<int>;
 
+//BeginCodeSnip{}
```
SansPapyrus683, line 45:
```suggestion
//BeginCodeSnip{DSU}
```

solutions/cf-1713E.mdx
Keys9873: done

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+	
+	// Construct the answer
+	int ans[n][n];
+	for (int i = 0; i < n; i++) {
+		for (int j = i; j < n; j++) {
+			/* If a[i][j] is part of a diagonal or the pair is equal (a[i][j] == a[j][i]),
+			 * manually set it to be the same as the original array
+			 */
```
SansPapyrus683, line 127:
```suggestion
			/*
			 * If a[i][j] is part of a diagonal or the pair is equal (a[i][j] == a[j][i]),
			 * manually set it to be the same as the original array
			 */
```

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+					ans[j][i] = a[j][i];
+				} else {
+					ans[i][j] = a[j][i];
+					ans[j][i] = a[i][j];
+				}
+			}
+		}
+	}
+	
+	// Print the answer
```
SansPapyrus683, line 144:
```suggestion
```
comments need not state the obvious

solutions/cf-1713E.mdx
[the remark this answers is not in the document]
SansPapyrus683: bump

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+			cout << ans[i][j] << " ";
+		}
+		cout << endl;
+	}
+}
+
+int main() { 
+	ios::sync_with_stdio(false); cin.tie(nullptr);
+	int T = 1; cin >> T;
+	while (T--) {
```
SansPapyrus683, line 156:
```suggestion
	int T;
	cin >> T;
	for (int t = 0; t < T; t++) {
```

SansPapyrus683 (changes requested):

solutions/cf-1713E.mdx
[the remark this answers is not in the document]
Keys9873: Yes you need it
I tried without and it tle'd

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+We can create a disjoint set union to model this problem. The DSU will have $2n$ nodes where node $i$ represents the state where the operation is performed on the $i$th row and node $i + n$ represents the state where operation is not performed on the $i$th row.
+
+As we are iterating greedily through the array, we can merge ($i$ and $j$) and ($i + n$ and $j + n$) if we don't want to swap. This works because if we swap neither or swap both the arrangement remains the same.
+If we do want to swap we can merge ($i$ and $j + n$) and ($i + n$ and $j$). This works because if we swap exactly one of these, then the arrangement will change. 
+
+After this we can construct the answer.
+
+We can iterate over the entire grid. If it is a diaganal it will never change. If $a[i][j] == a[j][i]$ then we can manually set them to normal (since we didn't merge them in the disjoint set union when comparing them the states might have not been merged). If it is not a diaganol we can check whether $i$ and $j$ are merged in the disjoint set union. If they aren't then we will swap $a[i][j]$ and $a[j][i]$. Otherwise we will keep them the same.
```
SansPapyrus683, line 29:
```suggestion
After this we can construct the answer by iterating over the grid. If it is a diagonal, it will never change. If $a[i][j] == a[j][i]$ then we can manually set them to normal (since we didn't merge them in the disjoint set union when comparing them the states might have not been merged). If it is not a diagonal, we can check whether $i$ and $j$ are merged in the disjoint set union. If they aren't then we will swap $a[i][j]$ and $a[j][i]$. Otherwise, we will keep them the same.
```
what does "iterating over the entire grid"
if it's cell by cell, what does "being a diagonal" mean?

Keys9873: iterating over grid =
for (int i = 0; i < n; i++) {
    for (int j = 0; j < n; j++) {
        // do stuff to a[i][j]
    }
}

diagonal = a[i][i]

SansPapyrus683: say this then
"We construct the answer by iterating over each cell of the grid."
"If the cell's row index is equal to its column index (i.e. it's on the diagonal of the grid"

Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,163 @@
...
+				if (dsu.get(i) != dsu.get(j + n)) {
+					dsu.unite(i, j);
+					dsu.unite(i + n, j + n);
+				}
+			}
+		}
+	}
+	
+	// Construct the answer
+	int ans[n][n];
```
SansPapyrus683, line 120: please use a vector

SansPapyrus683 (changes requested):

solutions/cf-1713E.mdx
Keys9873: vectors are too slow

Keys9873: i got inconsistent results when i used a vector
sometimes tle and sometimes ac

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+				}
+			}
+		}
+	}
+	
+	for (int i = 0; i < n; i++) {
+		for (int j = 0; j < n; j++) {
+			cout << ans[i][j] << " ";
+		}
+		cout << endl;
```
SansPapyrus683, line 147: perhaps try using `\n` here instead of `endl`, see if that fixes the vector issue

SansPapyrus683 (changes requested):

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+author: Aakash Gokhale
+---
+
+## Explanation
+
+[Official Analysis](https://codeforces.com/blog/entry/105419)
+
+Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operations are performed.
+
+In addition, we want to lexicographically sort the array. Thus, we can iterate and greedily process the array.
```
SansPapyrus683, line 14: > Thus, we can iterate and greedily process the array.

what does this mean

solutions/cf-1713E.mdx
Keys9873: yea it fixed the issue

solutions/cf-1713E.mdx
Keys9873: Since we want the lexicographically smallest answer, the earlier indices matter more than later indices so we find optimal row swaps for earlier indices first.

SansPapyrus683: then say that

Keys9873: fixed

SansPapyrus683:
@Aadit-Ambadkar review & i'll merge

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operations are performed.
+
+Since we want the lexicographically smallest answer, the earlier indices matter more than later indices so we find optimal row swaps for earlier indices first.
+
+For all $j$ such that $i < j$ (we only need to consider each pair once and keeping $i < j$ will ensure this), we can determine whether we want to swap them or keep them the same.
+
+If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to keep them the same. If they are equal it doesn't matter.
+
+However, it is possible that not all conditions can work together with each other and swapping something means we can't swap something else later on.
```
SansPapyrus683, line 20: so...how do we address this problem?

Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+[Official Analysis](https://codeforces.com/blog/entry/105419)
+
+Notice that $a[i][j]$ can only be swapped with $a[j][i]$ no matter which operations are performed.
+
+For all $j$ such that $i < j$ (we only need to consider each pair once and keeping $i < j$ will ensure this), we can determine whether we want to swap them or keep them the same.
+
+If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to keep them the same. If they are equal it doesn't matter.
+
+However, it is possible that not all conditions can work together with each other and swapping something means we can't swap something else later on. Since we want the lexicographically smallest answer, the earlier indices matter more than later indices so we find optimal row swaps for earlier indices first.
```
SansPapyrus683, line 18: > not all conditions can work together with each other

what does this mean exactly?

Keys9873: swapping/keeping the same $a[i][j]$ with $a[j][i]$ means we can't swap/keep the same $a[k][l]$ with $a[l][k]$ because $i = k$, $i = l$, $j = k$, $j = l$, or some combination.

Does this work as a replacement?

Keys9873: However, it is possible that not all conditions can work together with each other and swapping something/keeping something the same means we can't swap something else/keep something the same late on. For example it is impossible to swap $i$ and $j$, keep $j$ and $k$ the same, and keep $i$ and $k$ the same. We can first flip the $i$ith column to swap $i$ and $j$. We do nothing to $j$ or $k$ since we want to keep them the same. In order to keep $i$ and $k$ the same, we need to flip the $k$th column because we earlier swapped $i$ and two swaps will cancel out. However, swapping $k$ will mean that the second condition will fail and we have reached a scenario where we cannot satisfy all conditions.

Keys9873: is this better?

SansPapyrus683: at this point i'd just give an example

Keys9873: For example it is impossible to swap $i$ and $j$, keep $j$ and $k$ the same, and keep $i$ and $k$ the same.

Isn't this already an example?

SansPapyrus683: > swap $i$ and $j$

how can you swap $i$ and $j$? are these numbers? points?
instead of variables, just use concrete numbers, perhaps with the sample test case

Keys9873: However, it is possible that not all conditions can work together with each other and swapping something/keeping something the same means we can't swap something else/keep something the same late on. For example it is impossible to swap a[i][j] and a[j][i], keep a[j][k] and a[k][j] the same, and keep a[i][k] and a[k][i] the same. We can first flip the ith column/row which satisfies the first condition. We do nothing to the jth or kth column/row since we want to keep them the same which satisfies the second condition. In order to keep a[i][k] and a[k][i] the same, we need to flip the kth column/row and two swaps (i and k) will cancel out. However, swapping the kth column/row will cause the second condition to now fail and we have reached a scenario where we cannot satisfy all conditions.

Keys9873: I was unclear in my explanation. I was supposed to reference a[i][j], a[j][k], and a[i][k]

SansPapyrus683: this is pretty hard to follow
i'd use the example input or make an example array & say "it's possible to swap 3 & 4, but then this would [something bad]"

Keys9873: However, it is possible that not all conditions can work together with each other and swapping something/keeping something the same means we can't swap something else/keep something the same later on.
Consider to following input:
0 2 1
1 0 1
3 2 0
We want to first swap a[0][1] and a[1][0] because 1 is less than 2. We can do this by flipping the 0th row/col. This would give us the array:
0 1 3
2 0 1
1 2 0
Then we want to swap a[0][2] and a[2][0] because 1 is less than 3. To do this either we can either swap the 0th row/col or the 2nd row/col. However, swapping the 0th row/col would undo the efforts of our first operation so we will swap the 2nd row/col and get:
0 1 1
2 0 2
3 1 0
Now we want to swap a[1][2] and a[2][1] because 1 is less than 2 and we want the lexicographically minimum. However swapping the 1st row/col would mess up the swap we made in the first operation and swapping the 2nd row/col would mess up the swap we made in the second operation.
Thus, it is impossible to make all swaps optimally.

Is this better?

SansPapyrus683: much much better

Keys9873: fixed

solutions/cf-1713E.mdx
```diff
@@ -0,0 +1,184 @@
...
+For all $j$ such that $i < j$ (we only need to consider each pair once and keeping $i < j$ will ensure this), we can determine whether we want to swap them or keep them the same.
+
+If $a[i][j] > a[j][i]$, we want to swap them. If $a[i][j] < a[j][i]$ we want to keep them the same. If they are equal it doesn't matter.
+
+However, it is possible that not all conditions can work together with each other and swapping something/keeping something the same means we can't swap something else/keep something the same later on.
+Consider to following input:
+
+$0$ $2$ $1$
+$1$ $0$ $1$
+$3$ $2$ $0$
```
SansPapyrus683, line 23: did you check the formatting on the website? this seems kinda sus

SansPapyrus683:
hold on where's the problem in the guide itself? it doesn't appear when i search it

solutions/cf-1713E.mdx
Keys9873: fixed

SansPapyrus683:
strange, this problem's in a modulel but it doesn't appear in the search thing?

Keys9873:
I don't know why it isn't in the search thing. I added it to the extraPractice json file. Do I need to add it somewhere else?

SansPapyrus683:
idk, good enough i guess

SansPapyrus683:
image
typo

Keys9873:
Fixed
Istg I am so done with this PR

SansPapyrus683:
same, bro, same

Issue #3222, opened 2022-11-26
Profanity in user submissions?


SansPapyrus683:
so i was browsing the user solutions for [this](https://usaco.guide/silver/greedy-sorting/#problem-usaco-1158) problem
<img width="461" alt="image" src="https://user-images.githubusercontent.com/55369003/204071037-17d704dd-7bdb-42ae-8364-c6c262d144f2.png">
should we take any steps to implement some rudimentary filter, or just "let the viewer beware"

bqi343:
A filter seems reasonable. Though it's worth noting that you will occasionally encounter this when browsing public submissions on any site.

SansPapyrus683:
i mean we don't have to put too much effort into it, maybe just a regex replace will work

bqi343:
this should be extended to usernames (e.g., see https://usaco.guide/problems/cf-1478C/user-solutions)

PR #3238, opened 2022-11-30
added profanity filter to user sols

ady00:

Added a really simple regex profanity filter - let me know if I need to add/remove words. Tackles issue #3222.

- [x] I have tested my code.


bqi343:
it would be better to use a package like https://www.npmjs.com/package/bad-words 

(also, this would avoid needlessly injecting profanity into our codebase)

src/components/ProblemSolutions.tsx
```diff
@@ -203,7 +205,9 @@ export default function ProblemSolutions({
                             : undefined
                         }
                       >
-                        {submission.solutionCode}
+                        {filter.isProfane(submission.solutionCode)
+                          ? filter.clean(submission.solutionCode)
```
bqi343, line 209: why not just `filter.clean(submission.solutionCode)`?

bqi343 (changes requested):
don't commit package-lock

src/components/ProblemSolutions.tsx
ady00: Tried filter.clean and filter.isProfane on my personal build, filter.isProfane was significantly faster on sections with more than 30-40 public user solutions.

ady00:
final commit on #3242

PR #3240, opened 2022-12-01
Added python solution for Wormhole Sort

TamKengHong:

- [ ✓] I have tested my code.
- [ ✓] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [ ✓] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions), which include the following: 

### DIFF
--- a/content/1_General/Choosing_Lang.mdx
+++ b/content/1_General/Choosing_Lang.mdx
@@ -332,9 +332,54 @@ while lo != hi:
 print(-1 if lo == 0 else all_edges[lo-1][0])
 ```
 
-We are not sure whether it is possible to modify the above approach to receive
-full credit (please let us know if you manage to succeed)! Of course, it is
-possible to pass this problem in Python using DSU (a Gold topic):
+It is possible to optimise this approach even further to pass all testcases. This takes around 3.7s to run.
+
+```py
+def main():
+	f = open("wormsort.in","rb")
+	n,m = map(int, f.readline().split())
+	loc = [*map(int, f.readline().split())]
+	edges = [[] for _ in range(n)]
+	weights = []
+	lo,hi = 0, m - 1
+
+	def valid(loc, minW):
+		component = [-1]*n
+		numcomps = 0
+		for i in range(n):
+			if component[i] != component[loc[i] - 1]:
+				return False
+			elif component[i] == -1:
+				todo = [i]
+				component[i] = numcomps
+				for node in todo:
+					for child, weight in edges[node]:
+						if component[child] == -1 and weight >= minW:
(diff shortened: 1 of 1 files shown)


content/1_General/Choosing_Lang.mdx
```diff
@@ -332,9 +332,54 @@ while lo != hi:
...
+It is possible to optimise this approach even further to pass all testcases. This takes around 3.7s to run.
+
+```py
+def main():
+	f = open("wormsort.in","rb")
+	n,m = map(int, f.readline().split())
+	loc = [*map(int, f.readline().split())]
+	edges = [[] for _ in range(n)]
+	weights = []
+	lo,hi = 0, m - 1
```
bqi343, line 344: would suggest moving this line to just before `lo` and `hi` are used

content/1_General/Choosing_Lang.mdx
```diff
@@ -332,9 +332,54 @@ while lo != hi:
...
+	weights.sort()
+
+	while lo != hi:
+		mid = (lo + hi) // 2
+		if valid(loc, weights[mid]):
+			lo = mid + 1
+		else:
+			hi = mid
+
+	open("wormsort.out","w").write(f"{-1 if lo == 0 else weights[lo-1]}\n")
```
bqi343, line 377: can you fix this? (as mentioned in https://forum.usaco.guide/t/general-choosing-a-language-python-solution-for-usaco-silver-wormhole-sort/3699/3?u=benq)

bqi343 (changes requested):

bqi343:
btw should we remove the solution just above this one? seems like it only passes 5/10 test cases now ...

SansPapyrus683:
i mean the point is to demonstrate that python can be slow even when it meets the target complexity, right?

bqi343:
Yes, that's the point of the first Python solution in the module

SansPapyrus683:
oh wait there's already 2 python solutions, yeah, just have one that doesn't pass the test cases

TamKengHong:
Fixed all above
1. moved lo, hi to before bin search
2. changed to lo == m to output -1
3. removed the older solution

content/1_General/Choosing_Lang.mdx
TamKengHong: fixed

content/1_General/Choosing_Lang.mdx
TamKengHong: Fixed

bqi343 (approved):

PR #3242, opened 2022-12-01
uncomitted heavy files, added profanity filter

ady00:

Added a simple profanity filter through the bad-words lib. Extension to #3238.

- [x] I have tested my code.


src/components/ProblemSolutions.tsx
```diff
@@ -203,7 +205,9 @@ export default function ProblemSolutions({
                             : undefined
                         }
                       >
-                        {submission.solutionCode}
+                        {filter.isProfane(submission.solutionCode)
+                          ? filter.clean(submission.solutionCode)
```
ady00, line 209: Tried filter.clean and filter.isProfane on my personal build, filter.isProfane was significantly faster on sections with more than 30-40 public user solutions.

SansPapyrus683: using `filter.clean` on everything makes performance take a hit right?

ady00: I use `filter.isProfane` on everything, and `filter.clean` is only reserved for submissions that triggered `filter.isProfane`.
`Filter.isProfane` is run on every submission, and I don't think there's a way around that.

thecodingwizard: you can plausibly use React.useMemo() to avoid rerunning this expensive computation

ady00: Isn't React.useMemo() only beneficial when the page rerenders numerous times? AFAIK this page only renders once when it loads in all the solutions, so using useMemo wouldn't optimize that initial render.

thecodingwizard: if parent components rerender this component (might?) also rerender. you can probably test this just by logging to console during the render... (I could be wrong -- maybe it really only runs once?)

ady00: Yeah, the submissions don't rerender whenever the user upvotes a submission or adds a submission (which are the only components which can change states on the problem-solutions page), meaning that submissions only render once

thecodingwizard: UserDataContext can rerender (many) times I think

PR #3252, opened 2022-12-05
some maintenance stuff

SansPapyrus683:

- [x] I have tested my code.

### DIFF
--- a/solutions/cses-1640.mdx
+++ b/solutions/cses-1640.mdx
@@ -2,7 +2,7 @@
 id: cses-1640
 source: CSES
 title: Sum of Two Values
-author: Michael Cao, DRGSH, Benjamin Qi, Brad Ma, Ryan Chou, David Zhang
+author: Michael Cao, Benjamin Qi, Brad Ma, Ryan Chou, David Zhang
 ---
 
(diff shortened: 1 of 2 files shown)


ryanchou-dev (approved):
yap

solutions/cses-1640.mdx
```diff
@@ -41,132 +40,173 @@ One idea that comes to mind would be to used a boolean array to store the
...
+	cin >> n >> x;
+
+	vector<int> values(n);
 	for (int i = 0; i < n; i++) {
-		cin >> a[i];
+		cin >> values[i];
 	}
+	
+	// use a map to avoid using a very large array
+	map<int, int> val_to_ind;
```
ChuangSheep, line 87: any specific reason to use `map` instead of `unordered_map`? 
u don't necessarily have to change it, but in Java we also used a hashmap

ChuangSheep (approved):
there are some other styling problems but since the pr rather focuses on cleaning up & removing spoliers tag etc. I guess it's ok

solutions/cses-1640.mdx
SansPapyrus683: in c++ i just use map as default
it's actually faster than unordered_map sometimes

PR #3273, opened 2022-12-19
Update String_Hashing.mdx

### LINKED ISSUE #3195: Issues with String Hashing Module
https://usaco.guide/gold/string-hashing?lang=cpp

1. The first two resources say that the probability of the hashes of two different length-$n$ strings colliding is $\approx \frac{1}{MOD}$. Actually, Schwartz-Zippel only upper bounds this probability by $\frac{n}{MOD}$, and for certain prime moduli you can construct test cases that achieve this probability (see https://codeforces.com/blog/entry/60442). There are also other inaccuracies (e.g., $p$ needing to be prime in the cp-algo link)

2. > Since the test data for this particular problem is very strong, 

    Actually, the test data doesn't need to be "very strong" to fail solutions with a single modulus around $10^9+7$. All you need to do is construct a haystack with many different permutations of the needle, and then the solution will probably fail by the birthday paradox.

3. Why use a single base with two different moduli? Using two different bases with a single small modulus (or even better, just one base modulo a large modulus) have better theoretical guarantees ( $n^2/MOD^2$ and $n/MOD$, respectively), according to the blog above.

4. ~~Why is one solution labeled "one hash" and the other "two hashes?" They both use two hashes.~~ Also, it's probably worth noting that they have the same collision guarantees.

bqi343:
Closes #3195.

- [x] I have tested my code.

### DIFF
--- a/content/4_Gold/String_Hashing.mdx
+++ b/content/4_Gold/String_Hashing.mdx
@@ -16,7 +16,7 @@ prerequisites:
   <Resource source="CPH" title="26.3 - String Hashing"></Resource>
   <Resource source="cp-algo" title="String Hashing" url="string/string-hashing.html">code</Resource>
   <Resource source="PAPS" title="14.3 - Hashing">many applications</Resource>
-
+  <Resource source="rng-58" title="Hashing and Probability of Collision" url="http://rng-58.blogspot.com/2017/02/hashing-and-probability-of-collision.html" starred></Resource>
 </Resources>
 
 ## Implementation
@@ -157,7 +157,7 @@ polynomial will always be $B^{b - a}$.
 In general, when using polynomial hashing modulo a prime modulus $M$, 
 the probability of two distinct strings having equal hashing over all possible choices
 of the base $B$ can be up to $\frac{n}{M}$, where $n$ is the length of the longer
-of the two strings. See [rng-58's blog post about hashing](http://rng-58.blogspot.com/2017/02/hashing-and-probability-of-collision.html)
+of the two strings. See rng-58's blog post about hashing (linked above)
 for how to derive this probability using the Schwarz-Zippel lemma.
 
 </Info>
--- a/content/4_Gold/String_Hashing.mdx
+++ b/content/4_Gold/String_Hashing.mdx
@@ -158,7 +158,7 @@ In general, when using polynomial hashing modulo a prime modulus $M$,
 the probability of two distinct strings having equal hashing over all possible choices
 of the base $B$ can be up to $\frac{n}{M}$, where $n$ is the length of the longer
 of the two strings. See [rng-58's blog post about hashing](http://rng-58.blogspot.com/2017/02/hashing-and-probability-of-collision.html)
-for how this probability is derived.
+for how to derive this probability using the Schwarz-Zippel lemma.
 
 </Info>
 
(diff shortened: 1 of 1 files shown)


content/4_Gold/String_Hashing.mdx
```diff
@@ -13,26 +13,13 @@ prerequisites:
 ## Tutorial
 
 <Resources>
-  <Resource source="CPH" title="26.3 - String Hashing" starred>good intro</Resource>
-  <Resource source="cp-algo" title="String Hashing" url="string/string-hashing.html" starred>code</Resource>
-  <Resource source="PAPS" title="14.3 - Hashing" starred>many applications</Resource>
+  <Resource source="CPH" title="26.3 - String Hashing"></Resource>
```
bqi343, line 16: some of these resources may contain inaccuracies, so I unstarred them

content/4_Gold/String_Hashing.mdx
```diff
@@ -42,31 +29,26 @@ inverses.
...
 ```cpp
-#include <vector>
-#include <string>
-
-using namespace std;
-
 class HashedString {
 	private:
-		// change M and P if you want
+		// change M and B if you want
```
bqi343, line 34: use B for "base"(instead of "P" for prime - the base doesn't need to be prime)

content/4_Gold/String_Hashing.mdx
```diff
@@ -157,49 +139,83 @@ class HashedString:
...
+using prefix sums. This is nice because the highest power of $B$ in that
+polynomial will always be $B^{b - a}$.
+
+
+<Info title="Collision Probability">
+
+In general, when using polynomial hashing modulo a prime modulus $M$, 
+the probability of two distinct strings having equal hashing over all possible choices
+of the base $B$ can be up to $\frac{n}{M}$, where $n$ is the length of the longer
+of the two strings. See [rng-58's blog post about hashing](http://rng-58.blogspot.com/2017/02/hashing-and-probability-of-collision.html)
```
SansPapyrus683, line 160: why not just put the link in resources

bqi343: updated

content/4_Gold/String_Hashing.mdx
```diff
@@ -214,26 +230,56 @@ that). However, we need to count the **distinct** permutations of $N$ in $H$, so
...
 ```cpp
 #include <bits/stdc++.h>
-typedef long long ll;
 using namespace std;
 
-const ll P = 9973, M1 = 1e9 + 9, M2 = 1e9 + 7;
+using ll = long long;
+//BeginCodeSnip{HashedString}
+class HashedString {
+  private:
```
SansPapyrus683, line 250: these are spaces?

bqi343: that's intended

bqi343: well, that's just how my auto-formatter formats it

content/4_Gold/String_Hashing.mdx
```diff
@@ -157,49 +139,83 @@ class HashedString:
...
+
+In particular, a virtually unhackable way of choosing $B$ in the implementation
+above is as follows:
+
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+mt19937 rng((uint32_t)chrono::steady_clock::now().time_since_epoch().count());
```
SansPapyrus683, line 199: maybe explain what this gibberish means?

SansPapyrus683 (approved):
yeah

PR #3343, opened 2023-01-10
remove skipped/ignored from active dashboard

aaronthechen:
removes skipped and ignored tags from active modules/problems on the dashboard

- [x] I have tested my code.


bqi343:
why? use "ignored" instead of "skipped" if you don't want a problem to show up on the dashboard

SansPapyrus683:
TIL that the "ignored" option exists
i'm probably not the only one

SansPapyrus683:
i feel like we should make it more explicit as to what marking each problem does...

bqi343:
Can add to “using this guide”

PR #3436, opened 2023-02-16
god i want to eat strawberries now

ryanchou-dev:

- [ ] I have tested my code.

### DIFF
--- a/solutions/silver/usaco-990.mdx
+++ b/solutions/silver/usaco-990.mdx
@@ -5,11 +5,9 @@ title: Berry Picking
 author: Qi Wang, Daniel Suh, Juheon Rhee
 ---
 
-[Official Analysis](http://www.usaco.org/current/data/sol_berries_silver_jan20.html)
-
 <Spoiler title="Hint 1">
-Let $k$ equal the maximum number of berries in one basket,
-what can we say about the capacity of the rest of the baskets?
+Let $k$ equal the maximum number of berries in one basket.
+What can we say about the capacity of the rest of the baskets?
 </Spoiler>
 
 <Spoiler title="Hint 1 (answer)">
@@ -17,6 +15,9 @@ You want to maximize the number of baskets with $k$ berries.
 </Spoiler>
 
 <Spoiler title="Solution">
+
+[Official Analysis](http://www.usaco.org/current/data/sol_berries_silver_jan20.html)
+
 ## Implementation
 
 **Time Complexity:** $\mathcal{O}(max(B_i) \cdot N\log N)$
--- a/solutions/silver/usaco-990.mdx
+++ b/solutions/silver/usaco-990.mdx
@@ -2,10 +2,23 @@
 id: usaco-990
 source: USACO Silver 2020 January
 title: Berry Picking
-author: Qi Wang (Java), Daniel Suh (C++), Juheon Rhee (Python)
+author: Qi Wang, Daniel Suh, Juheon Rhee
 ---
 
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
uh could you follow the template i gave in my pr

ryanchou-dev:
oops

solutions/silver/usaco-990.mdx
```diff
@@ -2,20 +2,31 @@
 id: usaco-990
 source: USACO Silver 2020 January
 title: Berry Picking
-author: Qi Wang (Java), Daniel Suh (C++), Juheon Rhee (Python)
+author: Qi Wang, Daniel Suh, Juheon Rhee
 ---
 
+<Spoiler title="Hint 1">
+Let $k$ equal the maximum number of berries in one basket.
+What can we say about the capacity of the rest of the baskets?
```
SansPapyrus683, line 10: i might just be rusty w/ this problem but
what does max # of berries? isn't it just the largest tree?

ryanchou-dev: you don't have to take the entire tree ?

and motivation is that if you have a $k$-basket, then you should try to make as many $k$-baskets as possible.

SansPapyrus683: hb uh
"say our baskets were [1,2,3,4,5]. is there any difference between this basket set and [1, 2, 3, 3, 3]?"
not sure if this encourages the mindset for the solution but

ryanchou-dev: yeah that works. it does kinda give it away though. maybe we should leave drawing out the cases to the reader?

SansPapyrus683: "is it _really_ always the best for bessie to take all the berries from the trees?"

ryanchou-dev: ummm wait how does that relate

we want the reader to figure out that given an upper bound for each basket, we want as many baskets to reach this bound. the trees are just an interference in achieving this.

SansPapyrus683: how would they come upon the idea for an upper bound though, that seems so jank

ryanchou-dev: well if we give the maximum half of the baskets to bessie, then we want to other half to contain as much as possible, which implies that.

SansPapyrus683: well, my point is when i read this hint at first it's just like "k is just the largest tree bruh"

ryanchou-dev: oh ok, wdyt abt something like:

---

Let's say Bessie and Elsie get the same number of berries, what's a case that give us this? What can we do if this construction doesn't exist?

SansPapyrus683: how can a construction suddenly not exist

ryanchou-dev: Let's say Bessie and Elsie get the same number of berries, what's a case that give us this? How can we increase this value?

ryanchou-dev: [oopsies i forgot we weren't fixing any values]

SansPapyrus683: like bessie & elsie both get k berries?

ryanchou-dev: yea

SansPapyrus683: yeah that seems nice

Issue #3425, opened 2023-02-11
All problem statuses disappeared


Yash-Singh1:
I've been using USACO Guide for quite a while now and I had several problems marked as complete in the problem tracker, but today the site started refreshing randomly and then the site went into light mode for a second, then all the problems marked as complete became marked as "not started" for every division. The modules' and resources' progress did not change though.

bqi343:
Same as #3396?

bqi343:
@Yash-Singh1 Could you also answer the questions in https://github.com/cpinitiative/usaco-guide/issues/3396#issuecomment-1414102550?

Yash-Singh1:
- Are you signing in with Google / Github (and if so, which one)?

I sign in with Google, but I have both accounts linked

- How long ago did you create your account (or do you not have an online account)?

I am not sure

- Did you switch devices at any point?

I use the Guide on several devices

- When did this issue happen?

Early today morning

- Do you remember what you were doing before you lost all your data? Did you just sign out / sign in / etc?

The page was reloading so I closed it and reopened it then the problem data got erased.

- Is all your data lost (including module progress, streak, etc) or just your problem data?

No, it's just my problem data, I still have my streak, module progress, resource progress, etc.

bqi343:
@Yash-Singh1 some followups:

> I use the Guide on several devices

Were you using the Guide on several devices just before the disappearance?

> The page was reloading so I closed it and reopened it 

What were you doing before the page was reloading?

> No, it's just my problem data, I still have my streak, module progress, resource progress, etc.

Is your problem activity still present? (check https://usaco.guide/dashboard)

Yash-Singh1:
> Were you using the Guide on several devices just before the disappearance?

No, the last time I used it on another device was yesterday.

> What were you doing before the page was reloading?

I was solving a problem, and came back to the Guide page to mark it as complete.

> Is your problem activity still present? (check [usaco.guide/dashboard](usaco.guide/dashboard))

No, it's no longer there.

bqi343:
> I was solving a problem, and came back to the Guide page to mark it as complete.

Did the page start reloading after you attempted to mark the problem as complete?

Yash-Singh1:
> Did the page start reloading after you attempted to mark the problem as complete?

I don't think so, I came back to mark the problem as complete, when I noticed the page wasn't interactive, and I saw that it was reloading.

marcliang:
[...]

PR #3492, opened 2023-03-04
(try to) fix user progress on modules / problems deletion issue

### LINKED ISSUE #3396: User problem data getting erased
Hi, I'm the user that submitted the issue #3389.
I saw bqi343's comment on my issue.
image
Does this imply that the same problem had happened before?
If so, can you guys give me the link of the previous issue or maybe tell me how the issue got solved?
Thank you.

thecodingwizard:
closes #3396, #3425

I don't actually know if this works lol, but hopefully if data is lost we will only lose the data on that particular problem / module progress which is fine.

I think other places `setDoc` is used include daily # of pageviews, but I don't think it's that bad if that data is lost.

The scary part of using `updateData` is if the doc doesn't exist then it will crash and die. But this would only happen when the user account is first created, and I think once the account is created data is written anyway...

### DIFF
--- a/src/components/markdown/ResourceStatusCheckbox.tsx
+++ b/src/components/markdown/ResourceStatusCheckbox.tsx
@@ -6,6 +6,7 @@ import 'tippy.js/themes/light.css';
 import ConfettiContext from '../../context/ConfettiContext';
 import { useDarkMode } from '../../context/DarkModeContext';
 import MarkdownLayoutContext from '../../context/MarkdownLayoutContext';
+import { replaceIllegalFirebaseCharacters } from '../../context/UserDataContext/properties/userProgressOnResources';
 import UserDataContext from '../../context/UserDataContext/UserDataContext';
 import {
   ResourceInfo,
@@ -20,9 +21,8 @@ const StyledTippy = styled(Tippy)`
 `;
 
 const ProgressDropdown = ({ onProgressSelected, currentProgress }) => {
-  const [activeProgress, setActiveProgress] = useState<ResourceProgress>(
-    currentProgress
-  );
+  const [activeProgress, setActiveProgress] =
+    useState<ResourceProgress>(currentProgress);
 
   const icon = (status: ResourceProgress, equal: boolean) => {
     const colorMap: { [key in ResourceProgress]: string } = {
@@ -137,12 +137,10 @@ export default function ResourcestatusCheckbox({
 }): JSX.Element {
   const darkMode = useDarkMode();
   const markdownLayoutContext = useContext(MarkdownLayoutContext);
-  const { userProgressOnModules, setModuleProgress } = useContext(
-    UserDataContext
-  );
-  const { userProgressOnResources, setUserProgressOnResources } = useContext(
-    UserDataContext
-  );
+  const { userProgressOnModules, setModuleProgress } =
+    useContext(UserDataContext);
+  const { userProgressOnResources, setUserProgressOnResources } =
+    useContext(UserDataContext);
   const updateResourceProgressToPracticing = () => {
     if (
       markdownLayoutContext === null ||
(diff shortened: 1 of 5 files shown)


thecodingwizard:
Please help test!! Otherwise I will just randomly merge this at some point :)

src/components/markdown/ResourceStatusCheckbox.tsx
```diff
@@ -6,6 +6,7 @@ import 'tippy.js/themes/light.css';
 import ConfettiContext from '../../context/ConfettiContext';
 import { useDarkMode } from '../../context/DarkModeContext';
 import MarkdownLayoutContext from '../../context/MarkdownLayoutContext';
+import { replaceIllegalFirebaseCharacters } from '../../context/UserDataContext/properties/userProgressOnResources';
```
bqi343, line 9: mind explaining the changes you made?

src/context/UserDataContext/properties/userProgressOnResources.ts
```diff
@@ -52,30 +56,26 @@ export default class UserProgressOnResourcesProperty extends UserDataPropertyAPI
...
-            setDoc(
-              this.firebaseUserDoc,
-              {
-                [this.progressStorageKey]: {
-                  [problemId]: status,
-                },
-              },
-              { merge: true }
-            );
+            updateDoc(this.firebaseUserDoc, {
```
bqi343, line 69: what's the difference?

thecodingwizard: Actually, I realized there is no difference 🤡 

I thought `setDoc` might not work with nested objects but actually that is incorrect (it does work with nested objects). But I think `arrayUnion` only works with `updateDoc` (but I could be wrong)

src/components/markdown/ResourceStatusCheckbox.tsx
thecodingwizard: I read the problem progress from local storage every time before writing to Firebase, changed `setDoc` to `updateDoc`, and used `arrayUnion` to add an element into the progress array rather than override the entire array. Reasoning for each change:

- local storage: this shouldn't do anything, especially after the other two changes. (In fact I am like pretty sure this change does nothing)
- setDoc => updateDoc: this allows us to use `arrayUnion`. It's possible that a poor internet connection could cause user activity array to be overwritten by an outdated user activity array (this would be fixed by arrayUnion), but this doesn't explain why user problem progress is being lost.

thecodingwizard:
I think it's possible to write a firebase function that runs every time a user document changes. If we do this, we can check to see if the user progress on problems was being overwritten and try to reproduce the issue / debug from there. It seems like a nontrivial amount of work though :/

bqi343:
Screen Shot 2023-03-04 at 3 58 47 PM

this happens when I try to mark problems on the deployed URL for the first time. I can't reproduce it though ...

src/components/markdown/ResourceStatusCheckbox.tsx
bqi343: ok, then should we leave the issue open?

thecodingwizard:
@bqi343 can you check the dev console?

thecodingwizard:
Also can you reproduce this issue in the latest commit? https://usaco-guide-git-fix-data-deletion-cpinitiative.vercel.app/

there was a bug but I fixed it in the last commit I think

bqi343:
whoops I was pressing the wrong URL, seems okay in latest

src/components/markdown/ResourceStatusCheckbox.tsx
thecodingwizard: Probably. Maybe we can make one mega issue for reporting lost data and pin it?

bqi343: ok

bqi343 (approved):

PR #3497, opened 2023-03-05
remove duplicate appears in

### LINKED ISSUE #3495: Duplicate Appears In When Problem Appears Twice in a Module
https://usaco.guide/problems/cses-1753-string-matching/solution

Screen Shot 2023-03-04 at 9 48 09 PM

Also in problem search:

Screen Shot 2023-03-04 at 10 44 00 PM

bqi343:
closes #3495

- [ ] I have tested my code.


src/utils/algolia-queries.ts
```diff
@@ -135,7 +135,22 @@ const queries = [
...
+            const removeDuplicates = arrayOfObjects => {
+              return arrayOfObjects.filter(
+                (object, index) =>
+                  index ===
+                  arrayOfObjects.findIndex(
+                    obj => JSON.stringify(obj) === JSON.stringify(object)
+                  )
+              );
+            };
+            existingProblem.problemModules = removeDuplicates([
```
bqi343, line 149: for some reason, this didn't have any effect locally. is this expected? @thecodingwizard

thecodingwizard: Yes. The process to update algolia is separate (and I forgot how... I think it updates on `gatsby build` if certain environment variables are set. See https://github.com/algolia/gatsby-plugin-algolia and https://github.com/cpinitiative/usaco-guide/blob/master/gatsby-config.ts#L116 . There should be a fast way to test this but I forgot...)

thecodingwizard: I think a more efficient way to do this is to check if `moduleInfo` exists in `existingProblem.problemModules` already; if so, do nothing, if not, add. Can probably be done with something like `existingProblem.problemModules.find(module => module.id === moduleInfo.id)` or something (`find` returns `undefined` if no match is found)

thecodingwizard (approved):
(I didn't test)

bqi343:
looks like it worked

PR #3502, opened 2023-03-06
fix katex warnings

### LINKED ISSUE #3307: LaTeX Build Warnings
When running `yarn develop`:

> warn LaTeX-incompatible input and strict mode is set to 'warn': In LaTeX, \\ or \newline does nothing in display mode [newLineInDisplayMode]

How to determine which file is resulting in this warning?

bqi343:
closes #3307

I just looked through all the occurrences of `\\` and `\newline` ... 

is there a better way (e.g., get gatsby to print locations of warnings)? @thecodingwizard 

- [ ] I have tested my code.

### DIFF
--- a/content/4_Gold/Intro_Bitwise.mdx
+++ b/content/4_Gold/Intro_Bitwise.mdx
@@ -104,10 +104,12 @@ For simplicity, we will use the `sum` functions defined above. If we divide up
 $b$ into $2^{b_1}+2^{b_2}+\dots+2^{b_n}$, we get the following:
 
 $$
-a \times b \\
-= a \times (2^{b_1}+2^{b_2}+\dots+2^{b_n}) \\
-= a2^{b_1}+a2^{b_2}+\dots+a2^{b_n} \\
-= \sum_{\text{bits in b}} {a\texttt{<<}b_i}
+\begin{align*}
+&a \times b \\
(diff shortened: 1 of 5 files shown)


thecodingwizard:
I don't know; I think this is fine as-is (since editors presumably will get latex errors as they're editing a file anyway?). If we really wanted to, I think https://github.com/cpinitiative/usaco-guide/blob/master/src/mdx-plugins/rehype-math.js#L47 we could modify something here to print a line number but I'm not sure.

thecodingwizard:
could this be because gatsby caches the resuts of compiling files? ie. if you change and save the file, that would force gatsby to recompile (which you can test by adding a console.log to verify katex is actually running) and see if it outputs an error?

thecodingwizard:
also what do you mean by "an error is never thrown"? In the rehype-math file I linked it seems like we are setting `throwOnError` to fale

bqi343:
I mean that the code in the `catch` block never runs, because an error is never thrown in `try`. btw I just noticed that https://katex.org/docs/options.html allows you to pass a function in place of `strict`, will see if that works

bqi343:
huh for some reason they made it so that it's impossible to throw an error even if you pass a custom handler ...

https://github.com/KaTeX/KaTeX/blob/b58a43212e4178ed46cf8e42d67e88b9c8df9c24/src/Settings.js#L300

PR #3635, opened 2023-04-26
sz() function now takes reference to const container

drdilyor:
In the previous pull I made the sz function take reference, but now I realize that const reference also allows to use sz for rvalues, `sz(some_func_returning_vector())`

- [x] I have tested my code.


SansPapyrus683:
seems legit

PR #3739, opened 2023-06-11
Update to Google Analytics 4

thecodingwizard:
Apparently the older google analytics is ending on July 1, 2023. But maybe we should just use Vercel analytics instead...? Does anyone have thoughts?


SansPapyrus683:
damn i've never seen so many checks fail before

SansPapyrus683 (approved):
do we even look @ the analytics

PR #3746, opened 2023-06-13
Update Tree_Euler.problems.json

MariaChrysafis:
added new problem

(My first pull request, idk how this works. I have no code, just a problem proposition; since parenthesization is like an Euler tour of a tree, you can think of the problem as trees.)

- [x] I have tested my code.

### DIFF
--- a/content/4_Gold/Tree_Euler.problems.json
+++ b/content/4_Gold/Tree_Euler.problems.json
@@ -83,6 +83,18 @@
       }
     },
     {
+      "uniqueID": "cf-1821E",
+      "name": "Rearrange Brackets",
+      "url": "https://codeforces.com/contest/1821/problem/E",
+      "source": "Codeforces",
+      "difficulty": "Normal",
+      "isStarred": false,
+      "tags: ["Euler Tour"],
+      "solutionMetadata": {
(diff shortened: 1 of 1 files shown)


envyaims:
i dont think euler tour is really necessary no?

SansPapyrus683:
> since parenthesization is like an Euler tour of a tree, you can think of the problem as trees

apparently this is their reasoning

MariaChrysafis:
> i dont really get why this is euler tour... I think it's stacks/dfs + greedy at most.

Yeah, it's a bit of a stretch. But one solution is to think of parenthesization of as an Euler tour of a forest. Then, the cost function is just the sum of the depths of all nodes in the tree. Moving a bracket optimally is equivalent to extracting a node from the forest and making it its own connected component (this decreases the cost of the forest by the subtree size). Therefore, the problem is actually asking us to find the k largest subtrees in the forest and to remove them.

envyaims:
I'm sorry but there are a lot of other problems out there that fits the category a lot better. This problem isn't bad or anything, but I fail to see something like finding the k largest subtrees in the forest really fits with the euler tour ideology

PR #3770, opened 2023-06-23
Update Centroid.problems.json

MariaChrysafis:
I am not sure what to put as the year. The contest was during the 2016-2017 season, and it took place in 2017. But DMOJ marks it as 2016, since it was 2016-2017 season. I put it down as 2017.

- [x] I have tested my code.

### DIFF
--- a/content/5_Plat/Centroid.problems.json
+++ b/content/5_Plat/Centroid.problems.json
@@ -183,6 +183,18 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


dustin-miao:
In my opinion, I don't think we need any more normal difficulty problems in the CD module. Besides, the solution is nearly the same as [an existing problem](https://usaco.guide/plat/centroid#problem-cf-715C).

PR #3896, opened 2023-08-23
Simplify UserDataContext

### LINKED ISSUE #3511: [WIP] Simplify UserDataContext
Currently, `UserDataContext` feels quite complicated and unmaintainable.

# Current Situation

Every "API" (`userProgressOnProblems`, `adSettingsProperty`, etc) is responsible for writing to `localStorage`, reading from `localStorage`, erasing from `localStorage`, reading from a user data object from firebase, writing to a user data object from firebase, and creating a setter function that 

(todo finish writing this lol)

# New System

Everything happens through hooks.

## Reading

```jsx
const useUserLang = createUserDataGetter((userData) => {
    return userData.userLang;
});
```

## Writing

```jsx
const useSetUserLang = createUserDataMutation((userData, updatePayload) => {
    return {
        userLang: updatePayload
    };
});
```

A more complicated example:

```jsx
const useSetUserProgressOnProblem = createUserDataMutation((userData, updatePayload) => {
  const newActivityData = getNewActivityData(updatePayload);
  return {
    userProgressOnProblems: {
      ...userData.userProgressOnProblems,
      [updatePayload.problemId]: updatePayload.problemStatus,
    },
    userActivity: [...userData.userActivity, newActivityData],
  };
});

```

## Important Note

Both reading and writing functions must be pure (no side effects); they can be called randomly, multiple times, etc. They also should not mutate userData.

## Writing Implementation

Note: The below writing implementation is outdated. I realized as I was implementing this that we don't actually have a need for "transactions;" most of our updates can be done through the `updateDoc` function. Transactions had some annoying limits like a hard limit of 1 transaction / second, which could easily be exceeded by a user switching pages pretty fast.

### Case 1: User isn't signed in

1. Get the latest copy of userData by reading from localStorage
2. Run the setter function, passing in userData and the update payload
3. Merge the returned mutation with userData
4. Write to localStorage
5. Update userData state (triggering rerender)

### Case 2: User is signed in

Do all of Case 1 (to have an optimistic UI update where the user can see UI feedback before Firebase is updated since the network might be slow). Then:

1. Start a firebase transaction on userData.
2. Run the setter function, passing in the latest copy of userData and the update payload.
3. Write the returned mutation to firebase.
4. After the transaction completes, do nothing since our `onSnapshot` query on `userData` should get called.
5. If the transaction fails, show an error message and tell the user to reload the page.

### Misc Notes

A loading indicator should be shown beginning on part 2 and ending after part 2 is complete. In the future, if we want to add support for "pending" states, we should skip part 1 and only do part 2 if the user is signed in.

Todo: test to make sure that firebase transactions don't automatically lead to optimistic UI updates.

I think with this implementation of writing data, user data is even more unlikely to be lost (since we are using transactions, so there won't be an issue of outdated data overriding new data).

Warning: We'll have to migrate all the local storage keys to the new format (which assumes it to be one massive object). That doesn't seem like a fun rewrite :/

## Logging in for the first time

If remote account exists, override local data with remote account and save to localStorage. If remote account does not exist, upload local data to remote. Our current system is fine.

thecodingwizard:
Closes #3511 

This PR simplifies the implementation of UserDataContext, hopefully making it less buggy and easier to maintain / change in the future.

Future improvements:
- ~~loading indicators~~ never mind, I added a warning if user tries to exit before firebase finishes writing; I think this is good enough for now. I can't really think of a good way to show a non-intrusive loading indicator
- e2e testing (ex: https://firebase.google.com/docs/emulator-suite/connect_auth#web-modular-api) see #3902

Checkpoint where we used transactions instead of update doc: https://github.com/cpinitiative/usaco-guide/pull/3896/commits/e3d9ef788831f9db528c65e124f7d9120360bacd


thecodingwizard:
It is done! Just needs to be tested

SansPapyrus683:
uuuuhhh nathan?

thecodingwizard:
oops i didn't actually find time to test it, so maybe I'll just merge it LOL

PR #4063, opened 2023-12-02
Update Conclusion.problems.json

Pramad712:
(Move to functional graphs section, it has only 3 problems)

_Place an "x" in the corresponding checkbox if it is done or does not apply to this pull request._

- [x ] I have tested my code.
- [x ] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [x ] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions).
  - I understand that if it is clear that I have not attempted to follow these conventions, my PR will be closed.
  - If changes are requested, I will re-request a review after addressing them.
- [x ] I have linked this PR to any issues that it closes.

PR #4064, opened 2023-12-02
Add Maximize Value of Function for a Ball Passing Game

Pramad712:
This module has just 3 problems, it's best if it's moved here.

- [x ] I have tested my code.
- [ x] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [ x] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions).

### DIFF
--- a/content/3_Silver/Func_Graphs.problems.json
+++ b/content/3_Silver/Func_Graphs.problems.json
@@ -66,6 +66,18 @@
       "solutionMetadata": {
         "kind": "internal"
       }
+    },
+    {
+      "uniqueId": "lc-MaximizeValueOfFunctionInABallPassingGame",
+      "name": "Maximize Value of Function in a Ball Passing Game",
+      "url": "https://leetcode.com/contest/weekly-contest-360/problems/maximize-value-of-function-in-a-ball-passing-game/",
+      "source": "LC",
+      "difficulty": "Hard",
+      "isStarred": false,
+      "tags": ["Functional Graph"],
+      "solutionMetadata": {
+        "kind": "none"
+      }
     }
   ]
 }


Pramad712:
see https://github.com/cpinitiative/usaco-guide/pull/4063 to find where it'll be deleted (these two pull requests go with each other)

ryanchou-dev:
just close the other pr and do the change here.

you can switch to this branch after cloning / or even through the github web app and find the file and use the web edit tool.

ryanchou-dev:
also, taking a quick look -- this problem seems more like a standard binary lifting problem than a functional graph problem, so likely out of scope for this module

Pramad712:
binary lifting is technically functional graphs

dongliu0426:
also I believe there exists a solution without binary lifting that involves doing casework on the functional graph, which might be a bit more appropriate for silver

SansPapyrus683:
@dongliu0426 @ryanchou-dev so what yall think of this problem
put in bin lifting or use the scuffed casework sol

ryanchou-dev:
i think if the author can put in an internal solution for the one with casework, that'd be preferred...

otherwise, i feel like a lot of people attempting the problem will get lost doing it + reading the editorial (?)

Pramad712:
I'll close it for now until I get the solution.

Pramad712:
Actually, I don't think it's possible without binary lifting. The problem is that you can have a component with:

1. A cycle
2. Multiple lines of nodes that connect to this cycle
3. There are multiple connecting nodes

For example:
image

It's very hard to find and make sure the component is completely found. For example, doing a dfs or bfs starting from node 5 in this graph won't detect the other "branches" from the cycle (e.g. 11, 12, 8, 9). This is because functional graphs are directed.

Pramad712:
So, probably the binary lifting plat module.

PR #4086, opened 2023-12-15
automating problem updates

### LINKED ISSUE #4027: Automate USACO Problem Updates
Any time USACO releases new problems, the following things need to be done (potentially incomplete):

- Update USACO Monthlies table (maybe look at https://github.com/cpinitiative/usaco-guide/blob/8dc5cbf71f7ad4a5274b2912fbcc1d0c90f180e1/src/components/markdown/ProblemsList/DivisionList/scripts/README.md?plain=1#L8)?
- Update Algolia problem search for USACO Guide (this may be synced with USACO Monthlies table? I don't remember)
- Update https://github.com/cpinitiative/vjudge-usaco with new problems
- Update Algolia `usacoProblems` search for USACO Guide IDE to include new problems
- Update Serverless OJ with new problem data (this is ambitious and probably doesn't need to be done...)

They're done manually right now, but we should figure out how to do it automatically (or at least make it easier to do everything manually).

devo1ution:
closes #4027 
- [x] I have tested my code.


devo1ution:
Nevermind...it went back to the old behavior? Not sure what's going on 😭

SansPapyrus683:
so uh what's this tryna do

devo1ution:
umm the goal is to eventually resolve #4027 but I'm also figuring out how to get algolia to work locally rn (and also how the gatsby stuff works 😭 )

SansPapyrus683:
i mean another issue is how to actually scrape the usaco site

devo1ution:
We probably don't need that atm
https://github.com/cpinitiative/usaco-guide/issues/4027#issuecomment-1836604673

devo1ution:
@thecodingwizard so I added the prettier organize imports hook to pre-commit, but for whatever reason it ran against all the repo files (i was under the impression it would only run on modified ones) so this pull request has an insanely large diff 😓 is that ok or should I put it into a different pr or just get rid of the hook entirely

thecodingwizard:
It would be better to move the prettier changes into a different PR if that's possible. thank you!

devo1ution:
hey @thecodingwizard so my idea for future automation was something like:
1. we keep track of the last added id (`usaco_info.json`)
2. we schedule github action workflows a week after each contest window (although I'm not sure if this changes each year so the precise dates of these workflows are still tentative) that scans up from this last id

Here's the relevant commit: https://github.com/cpinitiative/usaco-guide/pull/4086/commits/e0f4f3f9a046f02aade018458af4b4201a2c854b
Relevant files: `update-usaco.yml`, `usaco_info.json`, `usaco_util.mjs` (there are some minor errors that get patched in the later comments)

what I'm still unsure of:
- [ ] the precise schedule (not sure if you have more experience with this?)
- [ ] how to implement the `Commit and push changes` step (I believe we need some kind of Github token?)

(also, I assume workflows can only be run once merged into master? So I'm not sure if we have any way of testing this one)

This code also assumes that between the end of each contest and the scheduled workflow date, none of the problems from that contest have been added to a module (although that shouldn't be too egregious an assumption and we can probably workaround this assumption too if needed, but it would be tedious)

Any help for any of the aforementioned questions would be appreciated!

thecodingwizard:
Thanks for all your work here @devo1ution !

It slipped my mind, but we actually have a `vjudge-usaco` repository that powers the USACO Guide IDE submission system that also needs a list of every USACO problem + its corresponding information. I'm thinking it may be a good idea to maintain one centralized `.json` file containing all the information for all the USACO problems, and any project that might need this information can just use this one JSON file. I've added you to the repository (the repo is private). Relevant files are `problems_all.ts`, `ProblemData.ts`, and `getProblemData.ts`.

The information vjudge-usaco needs is the following:

```ts
export type Sample = {
  input: string;
  output: string;
};

export type ProblemDataAll = {
  errorMsg: string;
  status: number;
  id: number;
  url: string;
  source: {
    sourceString: string;
    year: number;
    contest: string;
    division: string;
  };
  submittable: boolean;
  title: {
    titleString: string;
    place: number;
    name: string;
  };
  input: string;
  output: string;
  samples: Sample[];
};

export type ProblemData = {
  id: number;
  submittable: boolean;
  url: string;
  source: string;
  title: string;
  input: string;
  output: string;
  samples: Sample[];
};

const example: Partial<ProblemDataAll> = {
  id: 1142,
  url: "http://www.usaco.org/index.php?page=viewproblem2&cpid=1142",
  source: {
    sourceString: "2021 US Open Platinum",
    year: 2021,
    contest: "US Open",
    division: "Platinum",
  },
  submittable: true,
  title: {
    titleString: "3. Balanced Subsets",
    place: 3,
    name: "Balanced Subsets",
  },
  input: "stdin",
  output: "stdout",
  samples: [
    {
      input: "2\nGG\nGG\n",
      output: "13\n",
    },
    {
      input: "4\nGGGG\nGGGG\nGG.G\nGGGG\n",
      output: "642\n",
    },
  ],
};
```

Maybe we can combine the info vjudge-usaco and the info USACO Guide needs into one JSON file, and have each project extract the information it needs.

If we do this, then I would imagine next steps would look something like:

- Update `usaco_util.mjs` to fetch the information needed by both USACO Guide and vjudge-usaco. Parts of `getProblemData.ts` may be helpful here. `usaco_util.mjs` saves the output into a massive JSON file.
- Create a script that extracts the relevant information from this JSON file that USACO Guide needs.
- Add workflows, etc following the plan you laid out above. the LAST_ID can probably be computed by looking at the JSON file.

>  the precise schedule (not sure if you have more experience with this?)

I think running it every Tuesday, Wednesday, and Thursday every week December through May at 4am PT will probably be fine (ie. just avoid running this during contest window).

> how to implement the Commit and push changes step (I believe we need some kind of Github token?)

Unfortunately I'm not sure either. I would try to see if it can automatically open a GitHub PR, that would be most ideal. Maybe something like this? https://stackoverflow.com/questions/68057744/create-pull-request-with-github-action

Thanks again!

devo1ution:
> What do you think? It's also possible for us to move the USACO scraping code into a separate repository independent of USACO Guide / vjudge repos if we wanted to.

Hmm, I think having a separate repo would definitely make the codebase look cleaner, although do you know if we can/how we would reference files across repos? 

> Unfortunately I'm not sure either. I would try to see if it can automatically open a GitHub PR, that would be most ideal. Maybe something like this? https://stackoverflow.com/questions/68057744/create-pull-request-with-github-action

Sure, I'll try something like that! Although could you check this option is toggled:
> Make sure you also enable Allow GitHub Actions to create and approve pull requests under Settings > Actions in your repository

as per the stackoverflow article?

Thanks!

thecodingwizard:
> Hmm, I think having a separate repo would definitely make the codebase look cleaner

Sounds good! I made one here and gave you write access: https://github.com/cpinitiative/usaco-problems

> although do you know if we can/how we would reference files across repos?

Maybe have one workflow in the USACO problems repo to scrape problems, and another workflow in every other repository that fetches problem info from the public USACO problems repo (ie. with a simple GET request). So we would have three workflows, one for `usaco-problems`, one for `usaco-guide`, and another for `vjudge-usaco`.

Alternatively, in the `usaco-problems` repo, we can make a workflow to submit a PR to the USACO Guide / vjudge repositories, but this seems annoying: https://github.com/orgs/community/discussions/49745. It seems much easier for workflows to submit PR's to the repo they are part of.

> Although could you check this option is toggled:

I think it is, but if something's not working, please let me know!

devo1ution:
Also, I noticed that problems in `vjudge-usaco` are stored in a typescript file (`problems_all.ts`); would you prefer keep it like so or migrating it to json?

thecodingwizard:
Doesn't matter to me -- whatever you think makes more sense!

src/components/markdown/ProblemsList/DivisionList/div_to_probs.json
```diff
@@ -193,7 +193,19 @@
...
+    ["[object Undefined]", "2023 US Open", "Pareidolia"],
+    ["[object Undefined]", "2023 US Open", "Milk Sum"],
+    ["[object Undefined]", "2023 US Open", "Field Day"],
+    ["[object Undefined]", "2023 US Open", "Pareidolia"],
+    ["[object Undefined]", "2023 US Open", "Milk Sum"],
+    ["[object Undefined]", "2023 US Open", "Field Day"],
+    ["[object Undefined]", "2023 US Open", "Pareidolia"],
+    ["[object Undefined]", "2023 US Open", "Milk Sum"],
+    ["[object Undefined]", "2023 US Open", "Field Day"],
+    ["[object Undefined]", "2023 US Open", "Pareidolia"]
```
ryanchou-dev, line 208: ono

ryanchou-dev (commented):
great work so far! let mew know if you need any help :)

src/components/markdown/ProblemsList/DivisionList/div_to_probs.json
devo1ution: oh shoot i thought I fixed that wait

devo1ution:
@thecodingwizard significant progress!
I created a workflow that can now open prs to add missing problems (https://github.com/cpinitiative/usaco-problems/pull/2)
I did exclude a lot of checks present in `getProblemData.ts` from the `vjudge-usaco` repo, since that code accounted for a lot of misformatted/non-standard old problems, but this code should work into the forseeable future.
All that's left is to integrate this with `usaco-guide` and `vjudge-usaco`!

thecodingwizard:
Wow, this looks great. Thank you!

Actually, what are your thoughts on having the workflow in `usaco-problems` auto-commit straight to the repo rather than opening a PR? The `usaco-guide` and `vjudge-usaco` workflows would still create PR's.

The reason I bring this up is because the `usaco-guide` and `vjudge-usaco` workflows will poll the `usaco-problems` repo, which means the current setup would require us to merge the PR from `usaco-problems`, wait for the workflow for `usaco-guide` and `vjudge-usaco` to run (and these workflows are only set to run once a day), and then finally merge the PR's from `usaco-guide` and `vjudge-usaco`. Ideally we shouldn't have to wait a day for the workflow for `usaco-guide` / `vjudge-usaco` to run.

devo1ution:
oops, I just realized this script exists 😓 https://github.com/cpinitiative/usaco-guide/blob/master/src/components/markdown/ProblemsList/DivisionList/scripts/update.py
so uhh i think i may have just reinvented the wheel 😅 
> Actually, what are your thoughts on having the workflow in usaco-problems auto-commit straight to the repo rather than opening a PR? The usaco-guide and vjudge-usaco workflows would still create PR's.

Yeah, that would definitely make everything more convenient; although it might be possible the script generates faulty metadata, in which case it would be slightly more annoying to undo the changes across all 3 repos. If we do continue with the PR route, I think we could still do this:
> Alternatively, in the usaco-problems repo, we can make a workflow to submit a PR to the USACO Guide / vjudge repositories, but this seems annoying: https://github.com/orgs/community/discussions/49745. It seems much easier for workflows to submit PR's to the repo they are part of.

The code is a bit lengthy but doesn't seem too egregious 😅  However, if we're not worried about the code generating faulty metadata (I might actually need a second opinion on my code, actually, as I haven't tested it on very many ids yet and I'm also not sure what kind of formatting is standard and what isn't) having commits is definitely the best option!

bqi343:
>  I haven't tested it on very many ids yet

can't you test it on all ids and compare it to the output of `problems_all.ts`?

devo1ution:
I didn't want to test it on all ids as I wasn't sure if that would be too many burst requests to the usaco server, although its output does match `problems_all.ts` for the ~40 recent problems I tested; I was mostly just wondering whether I could rely on the h4/pre for sample input/outputs to extract the samples for each problem 😅

thecodingwizard:
> are div_to_prob and id_to_sol necessary?

I'm not sure -- it would be nice if they weren't!

> if we're not worried about the code generating faulty metadata

I think it's definitely possible even if your code is correct (ie. imagine USACO changes their website someday). I think it's OK to directly commit to usaco-problems repo, but we should definitely have PR's for usaco guide / vjudge-usaco so a human can review.

The link you sent for sending a trigger looks pretty cool! I'll trust your judgement on what to do, but I think both the trigger link you sent and making a PR to a different repo are very valid options.

devo1ution:
@thecodingwizard do you know if the `yarn build` check run in the pull request workflows pushes to dev_problems or prod_problems? I'm asking because I plan to implement merging `usaco-problems` into `usaco-guide` like this:
1. run `yarn build` to push all problems to algolia (probably `dev_problems`)
2. for each (new? or maybe just all, not sure if we have bandwidth) problem in `usaco-problems`, run an algolia query to check if it already exists
3. if it doesn't, add it to `extraProblems.json`

It would probably be more convenient if we query the graphql directly rather than routing the queries to algolia, although I'm not sure how we can access the graphql data outside the application.

thecodingwizard:
It runs on `dev_problems` (since it's a preview environment in Vercel).

> for each (new? or maybe just all, not sure if we have bandwidth) problem in usaco-problems, run an algolia query to check if it already exists

if it's a one-time thing, we have sufficient bandwidth probably.

> if it doesn't, add it to extraProblems.json

Is this just to get `extraProblems.json` to be up-to-date with all the past contests that have been released so far?

> It would probably be more convenient if we query the graphql directly rather than routing the queries to algolia, although I'm not sure how we can access the graphql data outside the application.

Gatsby has `graphiql` that might be helpful (lets you run queries in a UI interface): https://www.gatsbyjs.com/docs/how-to/querying-data/running-queries-with-graphiql/

devo1ution:
@thecodingwizard this took way too long but I finally got this:
https://github.com/devo1ution/usaco-guide/pull/2
This was auto-generated by this workflow: https://github.com/devo1ution/usaco-guide/actions/workflows/update-usaco.yml
Which was in turn called by this one: https://github.com/cpinitiative/usaco-problems/actions/workflows/autofetch.yml
Here are my todos for tomorrow 😅 :
- [ ] see if `autofetch` in `usaco-problems` can be run without PAT and just the basic GITHUB_TOKEN (as you can see, [this config](https://github.com/cpinitiative/usaco-problems/actions/runs/7320146775) failed, but I think we can modify the GITHUB_TOKEN permissions [like so](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token))
- [ ] tidy up code and remove extra debugging output

thecodingwizard:
whoa, that's great! The whole process of automating problem updates was a little harder than I expected it to be, so it's awesome that you're making so much progress on this. Thanks again :D

devo1ution:
OK @thecodingwizard I think the PR is finally ready 😓 
I haven't added integration with vjudge-usaco yet but I think we should get this out of the way first 😅 The only slight issue rn is that the `update-usaco` workflow doesn't exist on the master branch of this repo yet so I can't test it 100%, although it worked flawlessly on my forks, so hopefully it works the same here!

SansPapyrus683:
so where's the script that actually scrapes usaco

devo1ution:
it's in a separate repo: https://github.com/cpinitiative/usaco-problems

src/components/markdown/ProblemsList/DivisionList/div_to_probs.json
ryanchou-dev: still unfixed?

devo1ution: i think its fixed now 😭

devo1ution:
autofetch now makes edits rather than closing and reopening:
https://github.com/devo1ution/usaco-guide/pull/7

.github/workflows/update-usaco.yml
```diff
@@ -0,0 +1,49 @@
...
+          owner: ${{ github.repository_owner }}
+      - name: Build list of problem ids
+        run: |
+          git branch auto
+          git checkout auto
+          yarn
+          yarn develop &
+          sleep 20
+          pkill -f 'yarn develop'
+          ls
```
thecodingwizard, line 31: maybe we should add a comment here? it's not immediately obvious to me why `yarn develop` is needed to do this...

(Also, `yarn build` instead of `yarn develop` might be more reliable, but if it works as is then I guess it's fine...)

gatsby-node.ts
```diff
@@ -24,7 +24,7 @@ try {
 
 // ideally problems would be its own query with
 // source nodes: https://www.gatsbyjs.com/docs/reference/config-files/gatsby-node/#sourceNodes
-
+let stream = fs.createWriteStream('ids.log', { flags: 'a' });
```
thecodingwizard, line 27: maybe add a comment here explaining what this does / why this is needed?

.github/workflows/update-usaco.yml
devo1ution: yeah I think we should probably switch to yarn build rather than yarn develop (since killing the process after 20 secs is a bit sketchy); I just used develop so my workflows would run faster so I could test them quicker 😅

gatsby-node.ts
devo1ution: yeah, it's intentional bc `ids.log` is gitignored, so it starts empty at the start of every workflow run :)

PR #4157, opened 2023-12-30
migrate to instantsearch v7

danielzsh:

- [x] I have tested my code.


SansPapyrus683:
what this do

devo1ution:
um so the version of instantsearch we're currently using (v6) is deprecated so I'm trying to migrate to v7, but theres a few components I'm not sure how to migrate

src/components/ProblemAutocompleteModal/ProblemAutocomplete.tsx
```diff
@@ -38,42 +37,36 @@ const ProblemAutocompleteHit = ({
...
-export const ProblemAutocomplete = ({
-  hits,
-  currentRefinement,
-  refine,
-  onProblemSelect,
-  modalIsOpen,
-}) => (
-  <div>
+export const indexName =
+  process.env.NODE_ENV === 'production' ? 'prod_problems' : 'dev_problems';
```
thecodingwizard, line 41: should this use `GATSBY_ALGOLIA_INDEX_NAME`?

thecodingwizard (commented):
also, what was the change in `gatsby/xdm.js`?

devo1ution:
> also, what was the change in `gatsby/xdm.js`?

Hm, I'm actually not sure, I think it may have been from the algolia codemod; I just restored the original :)

src/components/ProblemAutocompleteModal/ProblemAutocomplete.tsx
devo1ution: Oh right 😅

src/pages/problems.tsx
```diff
@@ -143,3 +144,36 @@ export const pageQuery = graphql`
     }
   }
 `;
+
+// TODO (Codemod generated): ensure your usage correctly maps the props from the connector to the hook
```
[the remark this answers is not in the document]
devo1ution: oops wait

thecodingwizard:
On the problems page: https://usaco-guide-git-fork-devo1ution-migrate-ins-09f1bf-cpinitiative.vercel.app/problems/ If I try to set "Section" to "Bronze," all the problems disappear...

devo1ution:
Yeah, working on that rn 👍🏻

devo1ution:
Alright, this is kinda bad...
Instantsearch v7 no longer supports refining a list of filters (`refine(string[])`) and forces to refine one filter at a time (`refine(string)`) so the RefinementLists for module and problems are horrifically slow (try filtering to "Not Attempted" 😭); I haven't been able to find a workaround yet but lmk if you have any ideas @thecodingwizard 🙏🏻

SansPapyrus683:
so you have to filter things one attribute at a time or
(context: i have no clue what you're talking about)

devo1ution:
> so you have to filter things one attribute at a time or (context: i have no clue what you're talking about)

Yeah pretty much so its really unoptimized bc every time you add a filter it also triggers a re-render

I just pushed an update that uses indexUiState to hopefully bypass adding each filter one by one, but it unfortunately doesn't allow filtering by Object ID (which is needed for status filter) which is pretty annoying

devo1ution:
@thecodingwizard i think it should work now?

src/pages/problems.tsx
```diff
@@ -1,47 +1,126 @@
...
+        'Not Attempted',
+        'Solving',
+        'Reviewing',
+        'Skipped',
+        'Ignored',
+        'Solved',
+      ].map(label => ({
+        label,
+        value: problemIds.filter(
+          id => (userProgress[id] ?? 'Not Attempted') == label
```
thecodingwizard, line 119: `Not Attempted` seems like it might be an expensive query (ie. the list will be very long). Does it make more sense to *exclude* ID's specifically for `Not Attempted`?

devo1ution: Hm, I don't think we can use exclusion in conjunction with inclusion but my refactor to use `setIndexUiState` vs multiple `refine` calls should mitigate this issue

thecodingwizard: > I don't think we can use exclusion in conjunction with inclusion

Why not? Does adding a `-` in front of every refinement not work (https://discourse.algolia.com/t/facet-filter-negative-and-not-in-react-instant-search/4470)? (I haven't actually tried it so it's quite possible it doesn't work...)

> should mitigate this issue

Is there a limit of 1000 refinements for Algolia? If so, would this break if we exceed 1000 problems in the USACO Guide? (If we aren't that close to hitting the limit, it's probably fine as-is though.)

If we are close to hitting the limit, I think the "proper" solution will be to not allow searching by name in addition to searching by other properties. If we only search by name, we can use Algolia; if we only search by other properties like module, difficulty, and solve status, we can just use plain Javascript to make a list of all problem IDs that meet the criterion.

I can't think of a use case where a user will search by both name and other properties. The most likely use case I'm imagining is searching for a specific problem (searching by name) or searching to create a problemset (searching by properties like module, difficulty, solve status).

(We don't need to build this out now though)

src/components/ProblemsPage/Selection.tsx
```diff
@@ -1,26 +1,62 @@
...
+  const [refinements, setRefinements] = useState<string[]>([]);
+  const { setIndexUiState } = useInstantSearch();
+  useEffect(() => {
+    setIndexUiState(prevIndexUiState => ({
+      refinementList: {
+        ...prevIndexUiState.refinementList,
+        [attribute]: refinements,
+      },
+    }));
+  }, [refinements]);
```
thecodingwizard, line 44: on page load, will this be called once per `<Selection>` component? If so, will this result in many "operations"?

devo1ution: Hm, I'm not sure since refinements starts as an empty array; not sure if that counts towards operations? Is there any API to track operations (or for example log whenever an "operation" occurs)?

src/pages/problems.tsx
devo1ution: > Why not? Does adding a - in front of every refinement not work (https://discourse.algolia.com/t/facet-filter-negative-and-not-in-react-instant-search/4470)? (I haven't actually tried it so it's quite possible it doesn't work...)

I think this filters syntax ("AND", "NOT", "-", etc.) is unfortunately not generally available in Algolia's React/Hooks API (`useSearchBox`, `useRefinementList`, etc).

> If we are close to hitting the limit, I think the "proper" solution will be to not allow searching by name in addition to searching by other properties. If we only search by name, we can use Algolia; if we only search by other properties like module, difficulty, and solve status, we can just use plain Javascript to make a list of all problem IDs that meet the criterion.

Hm, we could probably build the filters directly into `ProblemHits` if necessary, although that does get rid of some small Algolia features like filter options changing dynamically based on the other filter states (in either case, we can definitely allow both searching by name and by property 😄 )

src/components/ProblemsPage/Selection.tsx
thecodingwizard: Maybe try monitoring the chrome network tab? Or try googling, I vaguely recall reading some post about how to do this...

devo1ution: I think according to [this algolia page](https://support.algolia.com/hc/en-us/articles/4406975243921-What-is-an-operation-)
> If you use the [disjunctive faceting](https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/#conjunctive-and-disjunctive-faceting) feature in your search, the number of operations for each query depends on the number of disjunctive facets you use.

It seems like each filter we use counts as an operation? (so I guess even with `setIndexUiState`, we're still taking 1000 operations 😭)

devo1ution:
also @thecodingwizard any clue what's going on with this lone typescript error? I can't figure out how to fix it 😭

src/components/ProblemsPage/Selection.tsx
bqi343: if the only issue is filtering by status, can we disable it for now and merge the rest if it's working?

danielzsh: Hm in that case the remaining refinement lists are still a bit unoptimized; apparently if you call `refine` on an already refined filter, it unrefines that filter (which isn't really mentioned anywhere in the docs). So in our current case right now if a user already has 2 refinements active and adds a 3rd one, this will take us 3 operations (i believe?), but theoretically it should only take 1. However, I'm still not quite sure how `setIndexUiState` counts operations yet, so I'm just waiting for algolia comes back online to verify 😓

danielzsh:
Alright, good news; Algolia provides an API monitoring service [here](https://dashboard.algolia.com/apps/3CFULMFIDW/monitoring/logs) so I tested filtering to not attempted, and the "nb operations" is only 2! (although I'm not sure why it's 2 rather than 1 yet)
image
So filtering by status should no longer be a problem.

SansPapyrus683:
what’s “nb” mean

danielzsh:
Number probably?

danielzsh:
I think we've already optimized ops as much as we can (see comment above) so can you all review one more time 🙏🏻

SansPapyrus683:
no need to request review from me

package.json
```diff
@@ -83,13 +85,13 @@
     "react-full-screen": "^1.1.0",
     "react-helmet": "^6.1.0",
     "react-hot-toast": "^2.2.0",
+    "react-instantsearch": "^7.5.0",
     "react-instantsearch-dom": "^6.28.0",
```
thecodingwizard, line 89: why are these version numbers different? (do we still need `react-instantsearch-dom`?)

src/components/ProblemsPage/Selection.tsx
thecodingwizard, line 44: if `refinements` is only used in `useEffect`, it may make more sense to just define a `setRefinements` function that directly calls `setIndexUiState` without having the `useEffect` wrapper. It doesn't really matter though.

thecodingwizard (approved):
I didn't look too carefully, if you think it's fine I trust you!

On the preview page, the page does look a bit squashed:

image


Also, if you select "Status = Solving" with no problems marked as "Solving," it doesn't actually return an empty list but rather just doesn't filter anything. That's not really a big problem though.

danielzsh:
> Also, if you select "Status = Solving" with no problems marked as "Solving," it doesn't actually return an empty list but rather just doesn't filter anything. That's not really a big problem though.

Hm, this was the original behavior but I can definitely change it!

> On the preview page, the page does look a bit squashed:

It looks normal to me?
image

src/components/ProblemsPage/Selection.tsx
danielzsh: Yeah I think we can just keep it as is for now 😅

package.json
danielzsh: oh oops i forgot to remove it 😅

danielzsh:
@thecodingwizard I added a "No Problems Found" info block, but when opening the selection menu it causes the whole page to scroll rather than just the menu, which is a bit annoying; any idea how to fix this?
image

thecodingwizard:
> It looks normal to me?

Ok maybe it's just my browser's issue then...

> any idea how to fix this?

hmm... maybe just set a minimum height on the search interface? You could make it span at least the remaining height, or just be lazy and hardcode some reasonable value?

danielzsh:
Alr, I think I fixed it 👍🏻

thecodingwizard:
Thanks @danielzsh !

I think I still see the sizing issue though:

image

However, it only appears if the window is not too wide (ie. try 1250px or less).

danielzsh:
> However, it only appears if the window is not too wide (ie. try 1250px or less).

Oh hm, let me see if I can fix that 😅 
Also, @thecodingwizard, any idea what's causing the sentry error above? I've never seen anything like it before.

thecodingwizard:
Not sure... seems like it only affected one person and is not related to this PR so I just ignored it xD

Issue #904, opened 2021-03-05
Default solution links for OI problems


dolphingarlic:
Some OIs (besides the IOI) have (somewhat) centralized pages with editorials/code solutions. It'd be nice if the solutions linked there by default.

- CEOI: http://ceoi.inf.elte.hu/tasks-archive/
- Baltic OI: https://boi.cses.fi/tasks.php (mostly code in a zip file)
- COCI: https://hsin.hr/coci/
- COI: https://hsin.hr/hio2020/ (replace 2020 with the desired year)

thecodingwizard:
This should probably be done after #930 is resolved...

SansPapyrus683:
oh wait crap forgot this pr wasn't really

bqi343:
it should be straightforward to do this if we want to (just modify https://github.com/cpinitiative/usaco-guide/blob/master/src/models/problem.ts#L323)

PR #4254, opened 2024-01-26
support adding arbitrary file path + refactor modals

danielzsh:
Adds an option to create a new solution from an arbitrary path:
1. 
image
2. 
image
3. profit
- [x] I have tested my code.


SansPapyrus683:
really appreciate this editor stuff
but could you possibly work on the issues here sometime :pleading_face:

danielzsh:
After I finish this I promise 🙏🙏

src/components/Editor/EditorFileModalInterface.tsx
```diff
@@ -45,11 +45,17 @@ const indexName = `${
...
   onSelect,
+  openAddFile,
 }: {
   onSelect: (file: AlgoliaEditorFile) => void;
+  openAddFile: () => void;
 }) => {
   const { query, refine: setQuery } = useSearchBox();
   const { hits } = useHits() as { hits: AlgoliaEditorFileHit[] };
+  console.log(hits.map(hit => hit.__position));
+  if (!hits.find(hit => hit.id === 'none')) {
```
thecodingwizard, line 56:
```suggestion
  const { hits } = useHits() as { hits: AlgoliaEditorFileHit[] };
  if (!hits.find(hit => hit.id === 'none')) {
```

src/components/Editor/EditorFileModalInterface.tsx
```diff
@@ -70,26 +76,38 @@ const FileSearch = ({
...
+                        {hit.path == null ? (
+                          'Create New Internal Solution'
+                        ) : (
+                          <Highlight hit={hit} attribute="path" />
+                        )}
+                      </SearchResultDescription>
+                    </button>
+                  ))
+                : 'test'}
             </div>
```
thecodingwizard, line 111: ?

src/components/Editor/AddProblemModal.tsx
```diff
@@ -31,98 +21,67 @@ async function addProblem(
       difficulty: 'N/A',
       isStarred: false,
       tags: ['Add Tags'],
-      solutionMetadata: parsed.solutionMetadata,
+      solutionMetadata: { kind: 'internal' },
     };
```
thecodingwizard, line 25: this modal is also used for the "Add Problem" button in `module.problems.json`; in that use case, I think we do want the solutionMetadata to be autofilled to whatever the source-specific solution metadata is.

thecodingwizard (commented):
nice!! what do you think of also adding a "Add New Problem (Solution)" button on the sidebar, similar to the "Login with Github" button?

It also seems like it might be nice to support adding internal solutions for generic problems (ie. problems that don't have a parser). we could do this in a separate PR though.

there's also a bunch of unnecessary console.log's...

src/components/Editor/AddFileModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    control: provided => ({
+                      ...provided,
+                      backgroundColor: '#111827',
+                      borderColor: '#374151',
+                    }),
+                    menuList: provided => ({
+                      ...provided,
+                      borderColor: '#374151',
+                      borderWidth: '1px',
+                      borderRadius: '6px',
```
thecodingwizard, line 56: I feel like this code is duplicated in multiple places in the codebase (https://github.com/cpinitiative/usaco-guide/blob/bbc8e5b44a4eb3f6f080672b02f76a950801045a/src/components/ProblemSuggestionModal.tsx#L250), (https://github.com/cpinitiative/usaco-guide/blob/bbc8e5b44a4eb3f6f080672b02f76a950801045a/src/components/ProblemsPage/Selection.tsx#L73), maybe more. we may want to clean it up sometime

thecodingwizard: it's not immediately related to this PR though, so feel free to do this in another PR if you don't want to do it now

src/components/Editor/EditorFileModalInterface.tsx
```diff
@@ -45,11 +45,17 @@ const indexName = `${
...
 }: {
   onSelect: (file: AlgoliaEditorFile) => void;
+  openAddFile: () => void;
 }) => {
   const { query, refine: setQuery } = useSearchBox();
   const { hits } = useHits() as { hits: AlgoliaEditorFileHit[] };
+  console.log(hits.map(hit => hit.__position));
+  if (!hits.find(hit => hit.id === 'none')) {
+    hits.push({ id: 'none' } as AlgoliaEditorFileHit); // blank hit
+  }
```
thecodingwizard, line 58: Instead of doing this, maybe it's neater to just add another Button after `hits.map` for "Add New Problem".

src/components/QuizGeneratorModal.tsx
```diff
@@ -122,196 +122,157 @@ export default function QuizGeneratorModal(): JSX.Element {
     return code;
   };
   return (
```
thecodingwizard, line 124: not related to your PR, but for performance reasons I think it's better of `getCode()` was rewritten as `const code = useMemo(..., [quiz])` or something. probably not super important though

src/components/Editor/AddFileModal.tsx
danielzsh: refactored!

src/components/Editor/EditorFileModalInterface.tsx
danielzsh: That actually might be slightly more annoying, since styling will have to unnecessarily copied (if someone in the future wants to modify it, they'll have two things to modify...); gonna keep it like this for now!

src/components/Editor/AddProblemModal.tsx
danielzsh: sure! I'm not sure what to put for sources that aren't cf or usaco though (AC, CSES); any ideas?

danielzsh:
Oh oops, just realized i never replied to this...
> nice!! what do you think of also adding a "Add New Problem (Solution)" button on the sidebar, similar to the "Login with Github" button?

I think I'm gonna keep it more discreet for now, since I believe usaco.guide does require approvals for adding new problems so really it should be a feature reserved for people who know what they're doing :)

src/components/Editor/EditorFileModalInterface.tsx
thecodingwizard: Hmm the styling being unnecessarily copied is a good point, but I think in this case it's still better to have the duplication:

- The duplication is fairly short; it's not a large amount of duplication
- The duplication is right next to each other and it's obvious that it's duplicated, so making changes to both at the same time (ex. if we wanted to change some styles) is pretty fast (just ctrl+d to make multiple cursors on VSCode). Tailwind CSS wrote a blog post about this at some point (duplicated HTML classes was fine because of multiple cursors or something)
- Without the duplication, there's extra logic that's not immediately obvious. In particular, the onClick() handler now needs a separate branch for `if (hit.id === 'none')`, and it's not immediately obvious why that's there and what it's doing. With the duplication, the logic can be different and it's very clear why (one button is handling a normal hit and the other button is adding an option to create a new solution). We don't need any special checks.
- It's also strange that hits[] now has an extra item at the end of it. Imagine one day we wanted to make a "hits" counter in the popup -- we would have to remember to subtract 1 from the length of hits[].
- creating a new solution does not always have to have the same styling as the rest of the hits. For example, imagine we wanted to add a "plus" icon next to create new solution -- with the current approach, we would have to add another `hits.id === 'none'` check to conditionally render the icon. Or, we might want to make the new solution button have static position in the modal.

If you really didn't want to duplicate the button though, maybe you could change `hits.id` from `none` to `CREATE_NEW_SOLUTION` or something to make it super obvious what that's for? and / or change `// blank hit` to a comment explaining that this is a hit placeholder used to show a create new solution button. Or, you could refactor the button into a new component, but I think this is probably overkill.

(If you strongly want to keep it as is though, that's fine, I would just be curious as to why -- I haven't thought too much about this so maybe there's something I missed)

src/components/Editor/AddProblemModal.tsx
thecodingwizard: Maybe just "kind: none"? IDK what the current default is

thecodingwizard: Also see https://github.com/cpinitiative/usaco-guide/issues/904  and https://github.com/cpinitiative/usaco-guide/blob/master/src/models/problem.ts#L323; there's some discussion about auto-generated solutions

src/components/Editor/AddFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+import Select from '../Select';
+import parse from './parsers/parse';
+export default function AddFileModal(props) {
+  const darkMode = useContext(DarkModeContext);
+  const [division, setDivision] = useState('');
+  const [fileStatus, setFileStatus] = useState('Create File');
+  const fileURLRef = useRef<HTMLInputElement>(null);
+  const createSol = useSetAtom(createNewInternalSolutionFileAtom);
+  console.log(division);
+  return (
```
thecodingwizard, line 17:
```suggestion
  const createSol = useSetAtom(createNewInternalSolutionFileAtom);
  return (
```

src/components/Editor/AddFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { createNewInternalSolutionFileAtom } from '../../atoms/editor';
+import { DarkModeContext } from '../../context/DarkModeContext';
+import { AlgoliaEditorSolutionFile } from '../../models/algoliaEditorFile';
+import Modal from '../Modal';
+import Select from '../Select';
+import parse from './parsers/parse';
+export default function AddFileModal(props) {
+  const darkMode = useContext(DarkModeContext);
+  const [division, setDivision] = useState('');
+  const [fileStatus, setFileStatus] = useState('Create File');
```
thecodingwizard, line 13: it would be nice to add types to these

src/components/Editor/AddFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <Modal {...props}>
+      <Dialog.Panel className="bg-white dark:bg-black w-full max-w-xl dark:text-white p-5 rounded-lg shadow-lg flex flex-col items-start">
+        <h3 className="text-lg font-bold">Enter Problem URL</h3>
+        <input
+          type="url"
+          className="shadow-sm focus:ring-blue-500 focus:border-blue-500 block w-full sm:text-sm border-gray-300 rounded-md dark:bg-gray-900 dark:border-gray-700"
+          placeholder="e.g. https://codeforces.com/contest/1920/problem/C"
+          ref={fileURLRef}
+        />
```
thecodingwizard, line 26: why ref and not controlled input (ie. `const [fileURL, setFileURL] = useState('')`?)

src/components/Editor/AddFileModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+            ].map(div => ({
+              label: div,
+              value: div.toLowerCase(),
+            }))}
+            onChange={e => setDivision(e.value)}
+          />
+        </div>
+        <button
+          className="btn mt-2"
+          onClick={async () => {
```
thecodingwizard, line 46: maybe add `disabled={fileStatus !== 'Create File'}` or something like that

src/components/Editor/AddProblemModal.tsx
```diff
@@ -33,96 +23,64 @@ async function addProblem(
...
+          Add Problem
+        </Dialog.Title>
+        <div className="mt-2 relative rounded-md shadow-sm">
+          <input
+            type="text"
+            className="shadow-sm focus:ring-blue-500 focus:border-blue-500 block w-full sm:text-sm border-gray-300 rounded-md dark:bg-gray-900 dark:border-gray-700"
+            placeholder="Enter Problem URL"
+            onChange={e => console.log(e.target.value)}
+            ref={linkRef}
+          />
```
thecodingwizard, line 58: same here -- why ref instead of `useState`?

src/components/Editor/AddProblemModal.tsx
```diff
@@ -33,96 +23,64 @@ async function addProblem(
...
-            </Transition.Child>
-          </div>
+        <div className="mt-4">
+          <button
+            className="btn"
+            onClick={() =>
+              linkRef.current &&
+              addProblem(linkRef.current.value, setMetadata, setStatus)
+            }
+          >
```
thecodingwizard, line 68: should probably add `disabled={status!=...}` here too, otherwise if someone clicks the button twice `addProblem` might get run twice

src/components/Editor/AddProblemModal.tsx
```diff
@@ -33,96 +23,64 @@ async function addProblem(
...
+        </div>
+        <div className="mt-4 relative">
+          <pre className="bg-gray-900 p-4 rounded-md text-white text-sm whitespace-pre-wrap">
+            {metadata}
+          </pre>
+          <CopyButton
+            className="btn absolute top-2 right-2"
+            onClick={() => {
+              navigator.clipboard.writeText(metadata);
+            }}
```
thecodingwizard, line 80: nice animation :)

src/components/Editor/parsers/cses.ts
```diff
@@ -7,8 +7,6 @@ export default function parseCses(url: string, html: string) {
     uniqueId: `cses-${problemId}`, // e.g. cses-1083
     name: html.match(titleRegex)?.[1] ?? 'Unknown',
     source: 'CSES',
-    solutionMetadata: {
-      kind: 'CHANGE THIS',
-    },
+    solutionMetadata: { kind: 'internal' },
```
thecodingwizard, line 10: i think there's a default CF solution metadata

src/components/Editor/parsers/parse.ts
```diff
@@ -8,11 +8,26 @@ export const parsers = {
...
-export default function parse(url: string, html: string) {
+export async function getHtml(url: string): Promise<string> {
+  const res = await fetch('/api/fetch-html', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ url }),
+  }).then(res => res.json());
+  return res.data;
```
thecodingwizard, line 19: I feel like the logic of parsing the HTML page belong on the server, not the client. It feels a bit weird to just have an `/api/fetch-html` endpoint like this. I'm fine with leaving it as-is though because it seems like a lot of effort to change it and I don't think it's a big deal

src/components/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  children: React.ReactNode;
+  isOpen: boolean;
+  onClose: () => void;
+  bg?: string;
+};
+export default function Modal({
+  children,
+  isOpen,
+  onClose,
+  bg = 'bg-black/25',
```
thecodingwizard, line 13: nice, thanks for refactoring this!

thecodingwizard: You might be interested in this: https://catalyst.tailwindui.com/

src/components/ProblemsPage/Selection.tsx
```diff
@@ -4,8 +4,8 @@ import {
   useInstantSearch,
   useRefinementList,
 } from 'react-instantsearch';
-import Select from 'react-select';
 import { useDarkMode } from '../../context/DarkModeContext';
+import Select from '../Select';
```
thecodingwizard, line 8: thanks for refactoring this!

src/components/QuizGeneratorModal.tsx
```diff
@@ -93,7 +93,7 @@ export default function QuizGeneratorModal(): JSX.Element {
     });
   };
 
-  const getCode = () => {
+  const getCode = useCallback(() => {
```
thecodingwizard, line 96: I think this should be `useMemo`, not `useCallback`, and rather than calling `getCode` we should just say `const code = useMemo(...)`. Otherwise we'll end up calling the `getCode()` function every render anyway.

src/components/QuizGeneratorModal.tsx
```diff
@@ -120,198 +120,159 @@ export default function QuizGeneratorModal(): JSX.Element {
...
+            </Dialog.Title>
+            <button className={'btn'} onClick={() => addQuestion()}>
+              Add question
+            </button>
+          </div>
+          <div className={'space-y-8 mt-4'}>
+            {quiz.map((question, idx) => (
+              <div
+                key={idx}
+                className="flex border-l-4 focus-within:border-blue-500 dark:border-gray-800 dark:focus-within:border-blue-900"
```
thecodingwizard, line 143: A key of idx is kind of sketchy (I think there may be some issues if you try deleting a quiz question in the middle of the list). The key should be unique per-question; one way to do this is to add a unique key to every question when you create it (ex. in `addQuestion()`, add an `id` property to the question that's `Math.random()` or `Date.now()`, and make that your key)

src/components/QuizGeneratorModal.tsx
```diff
@@ -120,198 +120,159 @@ export default function QuizGeneratorModal(): JSX.Element {
...
+                    onChange={e => changeQuestion(e.target.value, idx)}
+                    className={
+                      'mt-2 w-full shadow-sm focus:ring-blue-500 focus:border-blue-500 block text-sm border-gray-300 rounded-md dark:bg-gray-900 dark:border-gray-700'
+                    }
+                  />
+                  <div className={'ml-4'}>
+                    <h5 className={'font-medium my-2'}>Answers</h5>
+                    <div className={'space-y-4'}>
+                      {question.answers.map((answer, i) => (
+                        <div key={i} className={'flex'}>
```
thecodingwizard, line 166: same here -- indexed key is sketchy especially when you are able to remove items

thecodingwizard (commented):
looks good other than a few minor changes; thanks for all your work here!

src/components/Modal.tsx
danielzsh: Oh wow, it looks really nice...will definitely consider it moving forward! (not too willing to use it rn since its still in dev 😅 )

src/components/QuizGeneratorModal.tsx
danielzsh: Ah right, mb!

src/components/Editor/AddFileModal.tsx
danielzsh: Not sure, is there a significant difference? I always thought useRef on inputs would be faster, since typing wouldn't trigger a re-render.

src/components/Editor/EditorFileModalInterface.tsx
danielzsh: icic, thanks for the detailed explanation; will be changing this!

src/components/Editor/AddFileModal.tsx
thecodingwizard: I don't think the performance overhead will be that high... I think React generally prefers `useState`: https://stackoverflow.com/questions/76086163/what-is-better-usestate-or-useref-when-handling-multiple-form-inputs-in-react

danielzsh: got it 👍🏻

SansPapyrus683:
hey how's this going

danielzsh:
pretty well, js need a few more fixes :)

src/components/Editor/parsers/cses.ts
danielzsh: isn't this the cses parser?

src/components/Editor/AddProblemModal.tsx
danielzsh: hm, none of these apply to the currently available parsers; will be putting kind: none for now!

thecodingwizard (approved):
lgtm! if you've tested it and you think it works, feel free to merge whenever you're ready.

Thanks for adding this feature!!

SansPapyrus683:
💀

PR #4300, opened 2024-02-10
CSES Sliding Window Median - Java solution using Red-Black Order Statistics Tree

andreionoie:
Since there is no "official" order statistics library in Java and the code is not insanely long, we can fit an implementation along with the solution. Instead of storing Pairs of <int, int>, we can modify the behaviour of each node to store the count of that key and correctly handle it when updating the `size` field (which is used in the order statistics operations). This is a little bit more efficient than storing each duplicate node separately along with its index.

Code is directly adapted from CLRS Introduction to Algorithms 3rd ed.

- [x] I have tested my code.

### DIFF
--- a/solutions/gold/cses-1076.mdx
+++ b/solutions/gold/cses-1076.mdx
@@ -209,7 +209,7 @@ OS-SELECT(x, i):
 ```
 
 <Warning>
-Due to the tight time limit, the following Java solution still receives TLE on a few of the test cases.
+Due to the tight time constraints, the following Java solution still receives TLE on a few of the test cases.
 </Warning>
 
 ```java
@@ -241,6 +241,7 @@ public class Main {
 
 	// CodeSnip{Kattio}
 
+	// BeginCodeSnip{OrderedMultiset}
 	static class OrderedMultiset<K extends Comparable<K>> {
 		private class RBNode {
 			K key;
@@ -522,6 +523,7 @@ public class Main {
 			return rank;
 		}
 	}
+	// EndCodeSnip
 }
 ```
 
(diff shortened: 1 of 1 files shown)


andreionoie:
I guess it would be nice to have a separate page where we present Java alternatives to GNU PBDS or other cpp-native structures (e.g. multiset).

SansPapyrus683:
that's a good idea actually!

also uh i'm gonna need a screenshot of AC for merge

andreionoie:
Had to use an optimized `nextInt()` since the normal one from Kattio gave TLE:
<details><summary>Screenshot</summary>



</details>

SansPapyrus683:
hm it seems your code here still uses the normal kattio stuff

andreionoie:
Well yes, I thought it'd be more relevant to use the standard Kattio and have 2 test cases TLE instead of cluttering the code even more. (kinda sucks that you need to code the most optimized I/O for many CSES problems in Java)

SansPapyrus683:
hm if that's the case add a warning tag

solutions/gold/cses-1076.mdx
```diff
@@ -191,4 +191,339 @@ int main() {
...
+
+			io.print(" " + slidingWindow.findByOrder(medianRank));
+		}
+
+		io.close();
+	}
+
+	// CodeSnip{Kattio}
+
+	static class OrderedMultiset<K extends Comparable<K>> {
```
SansPapyrus683, line 244: codesnip this maybe?

SansPapyrus683 (approved):

Issue #4165, opened 2024-01-02
requested missing solutions


SansPapyrus683:
- [x] [poi sums](https://usaco.guide/gold/gold-conclusion/?lang=cpp#problem-poi-sums), external sol in [cf link](https://codeforces.com/blog/entry/18193?#comment-230790)
- [x] [fair nut & rectangles](https://usaco.guide/plat/convex-hull-trick#problem-cf-1083E)
- [ ] [examination](https://usaco.guide/plat/sqrt/#problem-joi-19-examination) (the user wants both the sqrt decomp sol & the non-decomp sol)
- [x] [vertex add component sum](https://usaco.guide/adv/offline-del#problem-ys-VertexAddComponentSum), someone has [a python sol](https://github.com/cpinitiative/usaco-guide/issues/3957#issuecomment-1738344895)
- [x] [angry cows gold](https://usaco.guide/silver/binary-search#problem-usaco-597)
- [x] [maximum product](https://codeforces.com/gym/100886/problem/G)
- [x] [police chase](https://usaco.guide/adv/min-cut/#problem-cses-1695)
- [x] [mobile O(N)](https://usaco.guide/problems/baltic-oi-2012mobile/solution)- sol sketch can be found [here](https://forum.usaco.guide/t/boi12-mobile/5660/6?u=benq). also be sure to add this problem to the stacks module
- [x] [hot & cold](https://usaco.guide/plat/binary-jump/#problem-bts-hotcold), solution sketches given [here](https://dmoj.ca/problem/bts17p7/editorial)

if anyone wants to work on any one of these solutions, feel free to!

ryanchou-dev:
ill get started on some of these 👍

danielzsh:
Ill work on the sol for maximum product :)

danielzsh:
also solved poi sums just now, will work on sol today or tmrw

ryanchou-dev:
ill do the mobile o(n) solution over the next few days!

danielzsh:
I feel like we should link the original issue next to each item in the list above as well

SansPapyrus683:
why's that

danielzsh:
since we closed the sub-issues before their corresponding internal sols were actually written, i feel like commenting again on each issue once the sol is finished would be helpful to let OP know

bqi343:
why can't we leave the sub-issues open and only close them when they're completed?

danielzsh:
[...]

PR #4321, opened 2024-02-17
Solution for POI Sums

danielzsh:
Addresses #4165 
- [x] I have tested my code.

### DIFF
--- a/solutions/gold/poi-sums.mdx
+++ b/solutions/gold/poi-sums.mdx
@@ -59,7 +59,7 @@ using namespace std;
 
 const int N = 5000;
 const int K = 50000;
-const int inf = 1e9;
+const int INF = 1e9;
 
 int a[N], d[K], vis[K];
 
@@ -69,7 +69,7 @@ int main() {
 	cin >> n;
 	for (int i = 0; i < n; i++) { cin >> a[i]; }
 	// d[i] -> min value we can make congruent to i (mod a[0])
-	fill(d, d + a[0], inf);
+	fill(d, d + a[0], INF);
 	d[0] = 0;
 	for (int i = 0; i < n; i++) {
 		// important observation: edges of a single weight will form several
--- a/solutions/gold/poi-sums.mdx
+++ b/solutions/gold/poi-sums.mdx
@@ -17,7 +17,7 @@ Building off the previous hint, consider writing numbers in terms of $x$ and $a_
 ## Explanation
 A crucial observation: if we can make a number $x$, **we can make any number $x + k\cdot a_0$**, where k is a non-negative integer. We may be tempted to rewrite this as so: if we can make $x$, we can make all $y \equiv x \mod a_0$. If this is indeed the case, we can run a simple 0/1 DP to figure out what mods we can assemble, and to answer a query $b_i$, we can simply output $\texttt{dp}[b_i \mod a_0]$. However, this solution has a flaw; try to think about why before moving on!
 
-An issue occurs when $x > a_0$. For example, if $n = 2$ and $a = \{3, 5\}$, we
+The issue occurs when $x > a_0$. For example, if $n = 2$ and $a = \{3, 5\}$, we
 can construct 8, but we cannot construct 2, even though $2 \equiv 8 \mod a_0$.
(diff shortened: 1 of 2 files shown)


solutions/gold/poi-sums.mdx
```diff
@@ -0,0 +1,103 @@
...
+<LanguageSection>
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+const int N = 5000;
+const int K = 50000;
+const int inf = 1e9;
```
SansPapyrus683, line 62: int32_max?

solutions/gold/poi-sums.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+int a[N], d[K], vis[K];
+
+int smin(int &a, int b) { return a = min(a, b); }
```
SansPapyrus683, line 66: you use this function only once, might as well take it out

solutions/gold/poi-sums.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+int main() {
+	cin.tie(0)->sync_with_stdio(0);
```
SansPapyrus683, line 69: needed?

solutions/gold/poi-sums.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+const int N = 5000;
+const int K = 50000;
+const int inf = 1e9;
+
+int a[N], d[K], vis[K];
```
SansPapyrus683, line 64: can u use vectors?

SansPapyrus683 (changes requested):
ik u said time limit was tight, so just asking here

solutions/gold/poi-sums.mdx
danielzsh: uhhh usually i never use int32_max to avoid potential overflow issues

solutions/gold/poi-sums.mdx
danielzsh: yep

solutions/gold/poi-sums.mdx
danielzsh: i got tle with vectors :(

solutions/gold/poi-sums.mdx
SansPapyrus683: seems like that's a problem w your implementation then lowk

danielzsh: ? it's just annoying to deal with; for example, inf + 1 should be greater than whatever you compare it too, but if you use int32_max inf + 1 will end up being some really negative number

SansPapyrus683: you should check if it's inf in the first place before doing such operations

danielzsh: bro this problem is actually tripping
I modified it like so and it TLE'd: https://szkopul.edu.pl/s/3224355/
but this submission, which is nearly identical aside from `inf` value and a single if statement, passes: https://szkopul.edu.pl/s/3224360/
???

SansPapyrus683: idk man i'm getting 403's for both of these submissions
but if that's so it doesn't really matter
just rename inf to `INF` and we good

danielzsh: done 👍

SansPapyrus683 (commented):
kk
just need proof of ac w/ the updated code

after that @ryanchou-dev can merge whenerver

SansPapyrus683 (approved):
github stinks

danielzsh:
since sending submission links doesnt work here's a screenshot
image

SansPapyrus683:
wth why do you only get 10 seconds

envyaims:
> wth why do you only get 10 seconds

0.1 seconds

SansPapyrus683:
i misspoke

danielzsh:
uhh @envyaims would you mind looking at this since ryan seems to be afk

ryanchou-dev (approved):
love this :D

PR #4325, opened 2024-02-18
add strictNullChecks

danielzsh:

- [x] I have tested my code.


danielzsh:
oops mb 😅

SansPapyrus683:
is this to get typescript to stop complaining

danielzsh:
actually we're getting typescript to complain more and then fix those complaints lol

danielzsh:
@thecodingwizard i've added a few non-null assertions to quiet tsc but before i keep adding more, would you mind taking a quick look and see if that's actually the best thing to do?

src/atoms/editor.ts
```diff
@@ -47,16 +47,13 @@ const baseActiveFileAtom = atomWithStorage(
   'guide:editor:activeFile',
   null as string | null
 );
-// Writable type needed if strictNullChecks is false
-// TODO: remove this when strictNullChecks is true
-export type Writable<T> = WritableAtom<T, any[], unknown>;
 export const branchAtom = atomWithStorage('guide:editor:branch', null);
-export const tokenAtom = atom(null) as Writable<string | null>;
+export const tokenAtom = atom(null);
```
thecodingwizard, line 51: will this cause `tokenAtom` to lose the "string" typing? should we do something like `atom<string | null>(null)` instead?

src/pages/editor.tsx
```diff
@@ -19,7 +19,7 @@ export default function EditorPageContainer(props: PageProps): JSX.Element {
     setHasMounted(true);
   }, []);
   if (!hasMounted) {
-    return null;
+    return <></>;
```
thecodingwizard, line 22: It may be better to modify the return type of `EditorPageContainer` to include `null`. Not a big deal though

src/stories/Tooltip.stories.tsx
```diff
@@ -22,12 +22,12 @@ const Template: Story<ComponentProps<typeof Tooltip>> = args => {
   const [show, setShow] = React.useState<boolean>(false);
   React.useEffect(() => {
     if (ref.current) setShow(true);
-  }, [ref.current]);
+  }, []);
```
thecodingwizard, line 25: why remove the check? if `ref.current` is always true then should we remove the if condition as well?

src/utils/utils.ts
```diff
@@ -17,7 +17,7 @@ export const getModulesForDivision = (
       }
       return {
         ...allModules[k2 as string],
-        slug: `/${division}/${allModules[k2 as string].frontmatter.id}`,
+        slug: `/${division}/${allModules[k2 as string].frontmatter?.id}`,
```
thecodingwizard, line 20:
```suggestion
        slug: `/${division}/${allModules[k2 as string].frontmatter!.id}`,
```

We probably just want to throw an error here? Otherwise it'll be `/{division}/null` or something (?)

thecodingwizard (approved):
I think these changes make sense to me, though I'm not an expert on this. Thanks for working on this!

src/stories/Tooltip.stories.tsx
danielzsh: eslint told me not to put ref.current in the dep array since it's not reactive?

src/atoms/editor.ts
danielzsh: ah right, I missed that 😅

danielzsh (commented):
btw, saw some comments in the codebase about "getting rid of classes"---i assume this refers to class components? I've noticed a few here and there, are we still hoping to migrate them?

thecodingwizard:
Yes, but I don't think this is high priority unless these are components that will be updated in the future. Feel free to use your best judgement!

danielzsh:
alright, I fixed all but three errors that I'm not quite sure how to deal with; @thecodingwizard would you mind taking a look at them?

thecodingwizard:
I'll take a look tomorrow. Please remind me if I forget!

src/components/MarkdownLayout/MarkdownLayout.tsx
```diff
@@ -119,8 +119,8 @@ export default function MarkdownLayout({
   //   document.querySelector('html').style.scrollBehavior = 'smooth';
   //   return () => (document.querySelector('html').style.scrollBehavior = 'auto');
   // }, []);
-  const problemSolutionContext = React.useContext(ProblemSolutionContext);
-  let activeIDs = [];
+  const problemSolutionContext = useProblemSolutions();
+  let activeIDs: string[] = [];
```
thecodingwizard, line 123: `useProblemSolutions()` throws an error if `ProblemSolutionContext` does not exist. `ProblemSolutionContext` exists only for solution pages, but `MarkdownLayout` is rendered both for solution pages and module pages (this should maybe be refactored).

`React.useContext()` returns null / undefined if `ProblemSolutionContext` does not exist, which is what we want here. We can assert that `problemSolutionContext` is defined if `markdownData` is not an instance of `ModuleInfo`.

thecodingwizard (dismissed):
This is a rubber-stamp approval -- I didn't actually look through these changes. Let me know if there's any part you want an actual review on.

Some misc TS errors:

```
src/components/Groups/PostLeaderboardPage/PostLeaderboardPage.tsx:55:36 - error TS2345: Argument of type '{ id: string; } | { id: string; language: string; problemID: string; score: number; submissionID: string; userID: string; type: string; verdict: string; timestamp: any; } | { id: string; score: number; ... 4 more ...; link: string; }' is not assignable to parameter of type 'FirebaseSubmission'.
  Type '{ id: string; }' is not assignable to type 'FirebaseSubmission'.

55         openProblemSubmissionPopup(submission);
```

need to remove the ID property from `submission`, or add ID to `FirebaseSubmission`. I actually don't know what the correct fix is -- I would definitely test this since removing the ID property might break things.

```
src/hooks/groups/useLeaderboardData.tsx:64:40 - error TS2345: Argument of type 'DocumentData' is not assignable to parameter of type 'LeaderboardEntry'.
  Type 'DocumentData' is missing the following properties from type '{ totalPoints: number; userInfo: { displayName: string; photoURL: string; uid: string; }; details: { [postId: string]: { [problemId: string]: { bestScore: number; bestScoreStatus: string; bestScoreSubmissionId: string; bestScoreTimestamp: Timestamp; }; }; }; }': totalPoints, userInfo, details

64       snap.forEach(doc => newData.push(doc.data()));
```

might need to cast `doc.data()` here? not sure

danielzsh:
Thanks for the help, everything should be fixed!

thecodingwizard (approved):

SansPapyrus683:
man why does sentry always complain

danielzsh:
sentry hates me for some reason 😭

PR #4333, opened 2024-02-20
fix flicker on "Select" elements

danielzsh:

- [x] I have tested my code.

### DIFF
--- a/src/context/DarkModeContext.tsx
+++ b/src/context/DarkModeContext.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useIsUserDataLoaded } from './UserDataContext/UserDataContext';
 
 // Note: this is separate from DarkModeProvider.tsx because this is imported by Storybook.
 
@@ -7,9 +6,6 @@ export const DarkModeContext = React.createContext<boolean>(null);
 
 export const useDarkMode = (): boolean => {
   const context = React.useContext(DarkModeContext);
-  if (!useIsUserDataLoaded()) {
-    return document.documentElement.classList.contains('dark');
-  }
   if (context === null) {
     throw new Error('useDarkMode() must be used inside a DarkModeProvider');
   }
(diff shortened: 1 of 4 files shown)


danielzsh:
This solution seems to work, but idk if its valid 😅

danielzsh:
wait nvm, i forgot to test `yarn build` 😭

src/context/DarkModeContext.tsx
```diff
@@ -1,11 +1,15 @@
...
 
 // Note: this is separate from DarkModeProvider.tsx because this is imported by Storybook.
 
 export const DarkModeContext = React.createContext<boolean>(null);
 
 export const useDarkMode = (): boolean => {
   const context = React.useContext(DarkModeContext);
+  if (!useIsUserDataLoaded()) {
+    return document.documentElement.classList.contains('dark');
+  }
```
thecodingwizard, line 12: Maybe this should be moved into `DarkModeProvider`?

Note that any access to document must be inside a `useEffect`. I think a better approach would be to leave this function alone and modify `DarkModeProvider` instead. DarkModeProvider waits for the user data to load before setting dark mode, which is slow. Instead, before `isLoaded` is set, it should do something similar to `gatsby-ssr.tsx` (ie. try loading from `themeKey` or default to system).

thecodingwizard (approved):
lgtm other than the comment I left!

src/context/DarkModeContext.tsx
danielzsh: Curious; even though the first value is likely still wrong, there's no flicker at all anymore? Do you know what's going on?

danielzsh: Ah I see, it's probably because everything on the problems page is only rendered client side 😅

danielzsh:
Btw, would you mind re-approving? I think any commits made after an approving review auto-dismisses it as stale, which is kinda annoying...

thecodingwizard (approved):

PR #4365, opened 2024-03-04
fix warnings

### LINKED ISSUE #4363: Warning EVERYWHERE
Why does every module have a warning in it? Even stranger is that the additional practice modules don't ...

danielzsh:
Closes #4363

- [ ] I have tested my code.

### DIFF
--- a/src/components/Dashboard/ModuleLink.tsx
+++ b/src/components/Dashboard/ModuleLink.tsx
@@ -203,7 +203,7 @@ const ModuleLink = ({ link }: { link: ModuleLinkInfo }): JSX.Element => {
     py: link.pyOc,
     showAll: maxLangOc,
   };
-  const isMissingLang = langToOc[userLang] ?? 0 < maxLangOc;
+  const isMissingLang = langToOc[userLang]! < maxLangOc;
   return (
     <LinkWithProgress
       lineColorStyle={lineColorStyle}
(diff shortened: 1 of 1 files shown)


src/components/Dashboard/ModuleLink.tsx
```diff
@@ -203,7 +203,7 @@ const ModuleLink = ({ link }: { link: ModuleLinkInfo }): JSX.Element => {
     py: link.pyOc,
     showAll: maxLangOc,
   };
-  const isMissingLang = langToOc[userLang] ?? 0 < maxLangOc;
+  const isMissingLang = langToOc[userLang]! < maxLangOc;
```
bqi343, line 206: @danielzsh hm I haven't seen this syntax before, can you explain what changed?

bqi343 (dismissed):

src/components/Dashboard/ModuleLink.tsx
danielzsh: I'm actually not quite sure why this fix works, as the only case where `langToOc[userLang] ?? 0` differs from `langToOc[userLang]!` is when `langToOc[userLang]` is null; however, `null` and `0` should compare the same? (as in `null < x` is the same as `0 < x`)

bqi343: can you print out exactly what's being compared before and after the change?

danielzsh: Oh yikes, it was an operator precendence issue...thanks for helping me catch that 😅 (btw would you mind re-approving)

bqi343 (approved):

PR #4382, opened 2024-03-13
aliens trick 👽

A1exL1ang:

- [x] I have tested my code.

### DIFF
--- a/content/6_Advanced/Lagrange.mdx
+++ b/content/6_Advanced/Lagrange.mdx
@@ -66,7 +66,7 @@ because we either begin a new subarray or we continue an existing subarray.
 
 Let $v$ be the maximal achievable sum with $\lambda$ penalty and $c$ be the number of subarrays used to achieve $v$. Then the **maximal possible sum achievable if we use exactly $c$ subarrays is $v+\lambda c$**. Note that we add $\lambda c$ to undo the penalty.
 
-Our goal is to find some $\lambda$ such that $c=k$. As we increase $\lambda$, it makes sense for $c$ to decrease since we are penalizing subarrays more. Thus, we can try to binary search for $\lambda$ to make $c=k$ and set our answer to be $v+\lambda c$ at the optimal $\lambda$.
+Our goal is to find some $\lambda$ such that $c=K$ (assuming $K$ is at most the number of positive elements). As we increase $\lambda$, it makes sense for $c$ to decrease since we are penalizing subarrays more. Thus, we can try to binary search for $\lambda$ to make $c=K$ and set our answer to be $v+\lambda c$ at the optimal $\lambda$.
 
 This idea almost works but there are still some very important caveats and conditions that we have not considered.
 
--- a/content/6_Advanced/Lagrange.mdx
+++ b/content/6_Advanced/Lagrange.mdx
@@ -116,51 +116,50 @@ Because calculating $v(\lambda)$ and $c(\lambda)$ with the dynamic programming s
 ```cpp
 #include <bits/stdc++.h>
 using namespace std;
- 
+
 #define ll long long
 
 const int MAX = 3e5 + 5;
 
-int n, k, A[MAX]; pair<ll, ll> dp[MAX][2];
- 
-pair<ll, ll> better(pair<ll, ll> a, pair<ll, ll> b){
-    if (a.first == b.first) 
-        return (a.second < b.second ? a : b);
+int n, k, A[MAX];
+pair<ll, ll> dp[MAX][2];
 
-    return (a.first > b.first ? a : b);
+pair<ll, ll> better(pair<ll, ll> a, pair<ll, ll> b) {
+	if (a.first == b.first) return (a.second < b.second ? a : b);
+
+	return (a.first > b.first ? a : b);
 }
 
-bool solve(ll lmb){
-    dp[0][0] = {0, 0}; 
-    dp[0][1] = {-1e18, 0};
- 
-    for (int i = 1; i <= n; i++){
-        dp[i][0] = better(dp[i - 1][0], dp[i - 1][1]);
+bool solve(ll lmb) {
+	dp[0][0] = {0, 0};
+	dp[0][1] = {-1e18, 0};
 
-        dp[i][1] = better(
-            {dp[i - 1][0].first + A[i] - lmb, dp[i - 1][0].second + 1}, 
-            {dp[i - 1][1].first + A[i], dp[i - 1][1].second}
-        );
-    }
-    return better(dp[n][0], dp[n][1]).second <= k;
+	for (int i = 1; i <= n; i++) {
+		dp[i][0] = better(dp[i - 1][0], dp[i - 1][1]);
+
+		dp[i][1] =
+		    better({dp[i - 1][0].first + A[i] - lmb, dp[i - 1][0].second + 1},
+		           {dp[i - 1][1].first + A[i], dp[i - 1][1].second});
+	}
+	return better(dp[n][0], dp[n][1]).second <= k;
 }
- 
-int main(){
-    ios_base::sync_with_stdio; cin.tie(0);
-    cin >> n >> k;
 
-    for (int i = 1; i <= n; i++) 
-        cin >> A[i];
- 
-    ll L = 0, H = 1e15;
+int main() {
+	ios_base::sync_with_stdio;
+	cin.tie(0);
+	cin >> n >> k;
 
-    while (L < H){
-        ll M = (L + H) / 2;
-        solve(M) ? H = M : L = M + 1;
-    }
-    solve(L);
-    
-    cout << better(dp[n][0], dp[n][1]).first + L * k << "\n";
+	for (int i = 1; i <= n; i++) cin >> A[i];
+
+	ll L = 0, H = 1e15;
+
+	while (L < H) {
+		ll M = (L + H) / 2;
+		solve(M) ? H = M : L = M + 1;
+	}
+	solve(L);
+
+	cout << better(dp[n][0], dp[n][1]).first + L * k << "\n";
 }
 ```
 ## Problems
--- a/content/6_Advanced/Lagrange.mdx
+++ b/content/6_Advanced/Lagrange.mdx
@@ -121,41 +121,43 @@ using namespace std;
 #define ll long long
 
 int main() {
-	int n, k;
-	cin >> n >> k;
+    int n, k;
+    cin >> n >> k;
 
-	int a[n];
-	for (int &i : a) cin >> i;
+    int a[n];
+    for (int &i : a) { cin >> i; }
 
-	/*
-	 * Returns {maximum sum, subarrays used (max if ties)}
-	 * if creating a subarray penalizes the sum by "lmb" and
-	 * there is no limit to the number of subarrays you can create
-	 */
-	auto solveLambda = [&](ll lmb) {
-		pair<ll, ll> dp[n][2];
+    /*
+     * Returns {maximum sum, subarrays used (max if ties)} 
+     * if creating a subarray penalizes the sum by "lmb" and 
+     * there is no limit to the number of subarrays you can create
+    */
+    auto solve_lambda = [&](ll lmb) {
+        pair<ll, ll> dp[n][2];
 
-		dp[0][0] = {0, 0};
-		dp[0][1] = {a[0] - lmb, 1};
+        dp[0][0] = {0, 0}; 
+        dp[0][1] = {a[0] - lmb, 1};
+    
+        for (int i = 1; i < n; i++){
+            dp[i][0] = max(dp[i - 1][0], dp[i - 1][1]);
 
-		for (int i = 1; i < n; i++) {
-			dp[i][0] = max(dp[i - 1][0], dp[i - 1][1]);
+            dp[i][1] = max(
+                make_pair(dp[i - 1][0].first + a[i] - lmb, dp[i - 1][0].second + 1), 
+                make_pair(dp[i - 1][1].first + a[i], dp[i - 1][1].second)
+            );
+        }
 
-			dp[i][1] =
-			    max(make_pair(dp[i - 1][0].first + a[i] - lmb,
-			                  dp[i - 1][0].second + 1),
-			        make_pair(dp[i - 1][1].first + a[i], dp[i - 1][1].second));
-		}
-		return max(dp[n - 1][0], dp[n - 1][1]);
-	};
+        return max(dp[n - 1][0], dp[n - 1][1]);
+    };
+    
+    ll lo = 0; 
+    ll hi = 1e18;
+    while (lo < hi) {
+        ll mid = (lo + hi) / 2;
+        solve_lambda(mid).second <= k ? hi = mid : lo = mid + 1;
+    }
 
-	ll lo = 0, hi = 1e18;
-
-	while (lo < hi) {
-		ll mid = (lo + hi) / 2;
-		solveLambda(mid).second <= k ? hi = mid : lo = mid + 1;
-	}
-	cout << solveLambda(lo).first + lo * k << endl;
+    cout << solve_lambda(lo).first + lo * k << endl;
 }
 ```
 
--- a/content/6_Advanced/Lagrange.mdx
+++ b/content/6_Advanced/Lagrange.mdx
@@ -122,37 +122,37 @@ using namespace std;
 
 const ll INF = (ll)300000 * 1000000000;
 
-int main(){
-    int n, k;
-    cin >> n >> k;
+int main() {
+	int n, k;
+	cin >> n >> k;
 
-    int a[n + 1];
-    for (int i = 1; i <= n; i++) cin >> a[i];
+	int a[n + 1];
+	for (int i = 1; i <= n; i++) cin >> a[i];
 
-    auto solveLambda = [&](ll lmb){
-        pair<ll, ll> dp[n + 1][2];
+	auto solveLambda = [&](ll lmb) {
+		pair<ll, ll> dp[n + 1][2];
 
-        dp[0][0] = {0, 0}; 
-        dp[0][1] = {-INF, 0};
-    
-        for (int i = 1; i <= n; i++){
-            dp[i][0] = max(dp[i - 1][0], dp[i - 1][1]);
+		dp[0][0] = {0, 0};
+		dp[0][1] = {-INF, 0};
 
-            dp[i][1] = max(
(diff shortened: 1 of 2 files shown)


envyaims (dismissed):

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+int n, k, A[MAX];
+pair<ll, ll> dp[MAX][2];
+
+pair<ll, ll> better(pair<ll, ll> a, pair<ll, ll> b) {
+	if (a.first == b.first) return (a.second < b.second ? a : b);
+
+	return (a.first > b.first ? a : b);
+}
+
+bool solve(ll lmb) {
```
SansPapyrus683, line 133: move this into main

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+
+#define ll long long
+
+const int MAX = 3e5 + 5;
+
+int n, k, A[MAX];
+pair<ll, ll> dp[MAX][2];
+
+pair<ll, ll> better(pair<ll, ll> a, pair<ll, ll> b) {
+	if (a.first == b.first) return (a.second < b.second ? a : b);
```
SansPapyrus683, line 128: braces

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+
+Because calculating $v(\lambda)$ and $c(\lambda)$ with the dynamic programming solution described above will take $\mathcal{O}(N)$ time, this solution runs in $\mathcal{O}(N\log{\sum A[i]})$ time.
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+#define ll long long
+
+const int MAX = 3e5 + 5;
```
SansPapyrus683, line 122: why +5

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+
+int n, k, A[MAX];
+pair<ll, ll> dp[MAX][2];
```
SansPapyrus683, line 125: better names

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+		dp[i][1] =
+		    better({dp[i - 1][0].first + A[i] - lmb, dp[i - 1][0].second + 1},
+		           {dp[i - 1][1].first + A[i], dp[i - 1][1].second});
+	}
+	return better(dp[n][0], dp[n][1]).second <= k;
+}
+
+int main() {
+	ios_base::sync_with_stdio;
+	cin.tie(0);
```
SansPapyrus683, line 149: remove if runs w/o

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+	cin >> n >> k;
+
+	for (int i = 1; i <= n; i++) cin >> A[i];
+
+	ll L = 0, H = 1e15;
```
SansPapyrus683, line 154: use `INT64_MAX` instead

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+
+	while (L < H) {
+		ll M = (L + H) / 2;
+		solve(M) ? H = M : L = M + 1;
+	}
+	solve(L);
+
+	cout << better(dp[n][0], dp[n][1]).first + L * k << "\n";
```
SansPapyrus683, line 162:
```suggestion
	cout << better(dp[n][0], dp[n][1]).first + L * k << endl;
```

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,145 @@ adding lambda\*smth
...
+	cin.tie(0);
+	cin >> n >> k;
+
+	for (int i = 1; i <= n; i++) cin >> A[i];
+
+	ll L = 0, H = 1e15;
+
+	while (L < H) {
+		ll M = (L + H) / 2;
+		solve(M) ? H = M : L = M + 1;
```
SansPapyrus683, line 158: also names should be in lowercase at least

SansPapyrus683 (changes requested):
cool pr

content/6_Advanced/Lagrange.mdx
A1exL1ang: Huh wdym? Like do you mean create a lambda in main or move it directly inside the binary search? It feels kind of weird and I edited my code so can you check if I still need to do this?

content/6_Advanced/Lagrange.mdx
A1exL1ang: ok i changed it to MAX=3e5 and declared my arrays with MAX+1 (since everything is 1 indexed)

content/6_Advanced/Lagrange.mdx
A1exL1ang: Wait is it fine if I don't since n, k, and A are the variables used in the problem and dp is like dp? Idk since the problem is themed around food and I don't want my variables to be about food lol

content/6_Advanced/Lagrange.mdx
A1exL1ang: ok i just used INF instead lol

content/6_Advanced/Lagrange.mdx
A1exL1ang: ok i changed it to lo and hi

content/6_Advanced/Lagrange.mdx
SansPapyrus683: i mean why can't int64_max work

content/6_Advanced/Lagrange.mdx
SansPapyrus683: alright fine but

> I don't want my variables to be about food lol

lmao what

content/6_Advanced/Lagrange.mdx
SansPapyrus683: move it directly inside the binsearch

while you're at it make `n`, `k`, `A`, and `dp` all local variables

A1exL1ang:
Can someone also check if the explanation is correct lol 😨 💀 :trust:

content/6_Advanced/Lagrange.mdx
A1exL1ang: uhh its fine but ig having INF is more convenient since i also initialize dp with INF. and ig another issue is that (lo+hi)/2 will overflow so u are forced to do the lo + (hi - lo) / 2 thingy

envyaims: why use int64_max though?

content/6_Advanced/Lagrange.mdx
A1exL1ang: wait hollup u need to run the same dp again after the binary search tho to get ur answer so shud i do it as a lambda lol

content/6_Advanced/Lagrange.mdx
SansPapyrus683: > uhh its fine but ig having INF is more convenient since i also initialize dp with INF. and ig another issue is that (lo+hi)/2 will overflow so u are forced to do the lo + (hi - lo) / 2 thingy

i'm pretty sure it wouldn't? i've done int64_max before and it worked just fine

content/6_Advanced/Lagrange.mdx
SansPapyrus683: come again? i don't see why you can't just extract the code from the function into the for loop
nothing in it seems to depend on anything

A1exL1ang: The function solveLambda is called in two separate places -- once inside the binary search and once in:
cout << solveLambda(lo).first + lo * k << endl;

So I will have to paste the same dp stuff into 2 separate places which is 😳

content/6_Advanced/Lagrange.mdx
A1exL1ang: oh wait yeah nvm it works in this case. ig ill still stick with inf since it is used to declare the dp and the dp cannot be initialized with int64_min

content/6_Advanced/Lagrange.mdx
SansPapyrus683: you could also store the result in a variable `last_run` or whatever

content/6_Advanced/Lagrange.mdx
SansPapyrus683: > the dp cannot be initialized with int64_min

why's that

A1exL1ang: the transition (specifically dp[i - 1][0].first + a[i] - lmb) will underflow

SansPapyrus683: i feel like you should check for the smallest value in that case

content/6_Advanced/Lagrange.mdx
A1exL1ang: Hmm i guess u can do that but u would have to make sure that you only store the result from the last true run (assuming its smth like FFFFTTTT) and you have to make sure 'hi' is greater than the max opt lambda. I think it makes the code feel strange or smth. 

Can I just leave it as a lambda in main? tbh i think it looks the best and is most natural when it's as a function outside of main.

content/6_Advanced/Lagrange.mdx
A1exL1ang: wait like set it to int64min and check if (dp[i - 1][0].first + a[i] - lmb) will underflow?

SansPapyrus683: on second thought just keep it as inf

content/6_Advanced/Lagrange.mdx
SansPapyrus683: alright then
but at least add some comments about what this function actually does, what it returns, etc.

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,141 @@ adding lambda\*smth
...
+
+#define ll long long
+
+const ll INF = (ll)300000 * 1000000000;
+
+int main() {
+	int n, k;
+	cin >> n >> k;
+
+	int a[n + 1];
```
SansPapyrus683, line 129: do you ever touch element 0 of this array?

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,141 @@ adding lambda\*smth
...
+
+Because calculating $v(\lambda)$ and $c(\lambda)$ with the dynamic programming solution described above will take $\mathcal{O}(N)$ time, this solution runs in $\mathcal{O}(N\log{\sum A[i]})$ time.
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+#define ll long long
+
+const ll INF = (ll)300000 * 1000000000;
```
SansPapyrus683, line 123: why not just smth like 1e18

SansPapyrus683 (changes requested):
almost done

content/6_Advanced/Lagrange.mdx
A1exL1ang: uhh no. i one indexed to make the dp transitions cleaner as i don't have to worry about i-1 being sus

content/6_Advanced/Lagrange.mdx
A1exL1ang: uhh do you think i should change it to 1e18? right now its that because i said the time complexity is \sum{A[i]} and making it 1e18 will make it like slightly slower i guess

content/6_Advanced/Lagrange.mdx
SansPapyrus683: > worry about i-1 being sus

if the array was 0-indexed, should your dp access -1 in the first place?
right now if your dp is doing so, currently this code is silently erroring

content/6_Advanced/Lagrange.mdx
SansPapyrus683: if it still ACs then yeah

content/6_Advanced/Lagrange.mdx
A1exL1ang: > right now if your dp is doing so, currently this code is silently erroring

wait my for loop starts at 1 so im not accessing -1. like basically everything is one indexed in the code. surely its good :trust:

> if the array was 0-indexed, should your dp access -1 in the first place?

if the array is 0 indexed i would be accessing -1 with the current code (since i would have to start at i=0) but ig it can be fixed by initializing the base case using A[0] but that seems kinda strange and less straight forward.

SansPapyrus683: idk i feel like making initializing the base case explicitly would be kinda nice

A1exL1ang: wait like make everything zero indexed and initialize the first entry with A[0]:
dp[0][0] = {0, 0};
dp[0][1] = {A[0] - lambda, 1};

instead of what I'm currently doing now:
dp[0][0] = {0, 0};
dp[0][1] = {-INF, 0};

SansPapyrus683: yeah i don't see anything wrong with that

content/6_Advanced/Lagrange.mdx
```diff
@@ -120,41 +120,42 @@ using namespace std;
...
 			        make_pair(dp[i - 1][1].first + a[i], dp[i - 1][1].second));
 		}
-
-		return max(dp[n][0], dp[n][1]);
+		return max(dp[n - 1][0], dp[n - 1][1]);
 	};
 
-	ll lo = 0;
-	ll hi = INF;
+	ll lo = 0, hi = 1e18;
```
SansPapyrus683, line 152: :skull: why'd you overwrite my changes

A1exL1ang: oh wait oops i copy pasted from the wrong thing 💀

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,144 @@ adding lambda\*smth
...
+
+int main() {
+	int n, k;
+	cin >> n >> k;
+
+	int a[n];
+	for (int &i : a) { cin >> i; }
+
+	/*
+	 * Returns {maximum sum, subarrays used (max if ties)}
```
SansPapyrus683, line 131:
```suggestion
	/**
	 * @returns the maximum sum along with the number of subarrays used
```

SansPapyrus683 (dismissed):
there we go
just show a screenshot of this current solution AC-ing and we're good to go

A1exL1ang:
image

wait has anyone reviewed if the geo description is correct lol

SansPapyrus683:
haha no lol
ping benq if you're feeling up to it

A1exL1ang:
@bqi343 are u willing to check if the explanations are correct

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,144 @@ adding lambda\*smth
...
+
+We construct a flow graph with source $S$, sink $T$, and $N+1$ additional vertices numbered $1$ to $N+1$. We will have the following edges.
+
+- A directed edge from $S$ to $i$ ($1 \le i \le N+1$) with weight $0$ and capacity $1$.
+
+- A directed edge from $i$ ($1 \le i \le N+1$) to $T$ with weight $0$ and capacity $1$.
+
+- A bidirectional edge from $i$ ($1 \le i \le N$) to $i+1$ with weight $A[i]$ and capacity $1$.
+
+$f(x)$ will be the maximum cost $x$-flow through the graph. We can repeatidly find the maximum cost augmenting path $x$ times to get our answer. Because the maximum cost as a function of flow is concave, $f(x)$ will be concave. You can read more about simulating cost flows [here](https://codeforces.com/blog/entry/118391).
```
bqi343, line 89:
```suggestion
$f(x)$ will be the maximum cost $x$-flow through the graph. We can repeatedly find the maximum cost augmenting path $x$ times to get our answer. Because the maximum cost as a function of flow is concave, $f(x)$ will be concave. You can read more about simulating cost flows [here](https://codeforces.com/blog/entry/118391).
```

content/6_Advanced/Lagrange.mdx
```diff
@@ -27,6 +23,144 @@ adding lambda\*smth
...
+
+<FocusProblem problem="sample" />
+
+The problem gives us a length $N$ ($1 \le N \le 3 \cdot 10^5$) array of integers in the range $[-10^9,10^9]$. We are given some $K$ ($1 \le K \le N$) and are asked to choose at most $K$ disjoint subarrays such that the sum of elements included in a subarray is maximized.
+
+### Intuition
+
+The main bottleneck of any dynamic programming solution to this problem is having to store the number of subarrays we have created so far.
+
+Let's try to find a way around this. Instead of storing the number of subarrays we have created so far, we assign a penalty of $\lambda$ for creating a new subarray (i.e. everytime we create a subarray we penalize our sum by $\lambda$).
```
bqi343, line 38:
```suggestion
Let's try to find a way around this. Instead of storing the number of subarrays we have created so far, we assign a penalty of $\lambda$ for creating a new subarray (i.e. every time we create a subarray we penalize our sum by $\lambda$).
```

bqi343 (changes requested):
I didn't read that carefully but looks correct

A1exL1ang:
made a small change to the binary search. everything should be good now :trust:

content/6_Advanced/Lagrange.mdx
```diff
@@ -153,8 +153,8 @@ int main() {
 	ll lo = 0;
 	ll hi = 1e18;
 	while (lo < hi) {
-		ll mid = (lo + hi) / 2;
-		solve_lambda(mid).second <= k ? hi = mid : lo = mid + 1;
+		ll mid = (lo + hi + 1) / 2;
+		solve_lambda(mid).second >= k ? lo = mid : hi = mid - 1;
```
bqi343, line 157: so was it wrong before?

bqi343 (approved):

content/6_Advanced/Lagrange.mdx
A1exL1ang: yeah i think it was wrong before (somehow ACs due to weak tests or smth). i stress tested and found some cases where the code before WAs. hopefully it's correct now lol

SansPapyrus683:
OOPS forgot to pay you
can you contact @thecodingwizard to figure out payment or whatever

PR #4411, opened 2024-03-27
small fix to workflow (pray it works)

danielzsh:


SansPapyrus683:
you just removed an assertion?

danielzsh:
yeah because that assertion isn't always true anymore (last results table didnt list observers for some reason)

SansPapyrus683 (approved):
haha

PR #4455, opened 2024-04-27
Update Graph_Traversal.mdx

MFael1:
Refactor BFS algorithm to ensure nodes are visited only once

This commit refines the BFS algorithm implementation to correctly handle cases where a node has multiple parents at the same level. Previously, nodes could be added to the queue multiple times, leading to potential inefficiencies or incorrect traversal results.

Changes include:
- Marking a node as visited immediately after it is added to the queue, ensuring it is not processed again if encountered through another parent at the same level.
- Ensuring that each node is visited exactly once by checking the `visited` status before adding a neighbor to the queue.

- [x] I have tested my code.

### DIFF
--- a/content/3_Silver/Graph_Traversal.mdx
+++ b/content/3_Silver/Graph_Traversal.mdx
@@ -476,15 +476,15 @@ int main() {
 		if (!visited[i]) {
 			queue<int> q;
 			q.push(i);
-
+			visited[i] = true;
 			while (!q.empty()) {
 				int current_node = q.front();
 				q.pop();
-
-				visited[current_node] = true;
-
 				for (int neighbor : adj[current_node]) {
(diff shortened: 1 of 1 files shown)


content/3_Silver/Graph_Traversal.mdx
```diff
@@ -476,15 +476,15 @@ int main() {
 		if (!visited[i]) {
 			queue<int> q;
 			q.push(i);
-
+			visited[i] = true;
 			while (!q.empty()) {
 				int current_node = q.front();
 				q.pop();
-
-				visited[current_node] = true;
```
SansPapyrus683, line 484: i mean didn't this work?

SansPapyrus683 (approved):

PR #4508, opened 2024-06-09
spruce up binjump module

SansPapyrus683:

- [x] I have tested my code.


content/5_Plat/Binary_Jump.mdx
```diff
@@ -31,77 +35,117 @@ prerequisites:
...
+	<source src="/animations/binary_jumping.mp4" type="video/mp4"/>
+</video>
 
-</Spoiler>
+To actually compute our binary jumping table, we start with the $2^0=1$st
+parents of each node, which is their direct parent.
+We then move on and calculate $2^1=2$nd parents, using the fact that
+the $2$nd parent can be calculated as the $1$st parent of the $1$st parent.
+Using similar logic, we move on to $2^2$, $2^3$, and so on.
+We stop when $2^n$ is greater than the size of the tree, since at that point
```
TheGamingMousse, line 63: i think we should keep the part where it talks about how to calculate the number of rows for the binlift table

SansPapyrus683: > We stop when $2^n$ is greater than the size of the tree

unless you think this isn't enough?

ryanchou-dev (approved):
looks great!

content/5_Plat/Binary_Jump.mdx
```diff
@@ -331,68 +481,136 @@ int main() {
...
+### Explanation
 
-Find LCA of node $a$ and $b$ as described above. Then, the distance between the
-two nodes would be the
+Since we have the depth of all the nodes, the distance between two nodes $a$ and $b$ is
 $\texttt{depth}[a] + \texttt{depth}[b] - 2 \cdot \texttt{depth}[\textrm{lca}(a, b)]$.
 
-<LanguageSection>
+Here's some intuition if you're confused about how this formula works.
+To get from node $a$ to $b$, one way would be to go to the root of the tree
```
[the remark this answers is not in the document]
SansPapyrus683: thanks haha

content/5_Plat/Binary_Jump.mdx
TheGamingMousse: fair enough 

also another thing, setting the number of rows to ceil(log2(n)) is only enough for 2^n - 1, which doesn’t matter for LCA but can matter for using binlift for range queries

SansPapyrus683: why's that? if it goes up to 2^n, then you can have at most 2^n parents

TheGamingMousse: it doesn’t matter cuz for most tree problems you only need to lift up n-1 times (eg if the tree is a line). but if you are using binlift for range queries, i’m p sure ceil(log2(n)) is problematic. i’ll try to find an example later (or i’m just tripping)

SansPapyrus683: so uh
did you find a sussy test case yet

TheGamingMousse: an example of when ceil(log2(n)) does not work for range queries:

ac: https://codeforces.com/contest/1142/submission/261280040
rte: https://codeforces.com/contest/1142/submission/265726905

granted, the module primarily talks about binlift for tree/graph stuff. but I think using something like 1 + (int) log2(n) is a better way to write it in general (also because you can write it as 1 + __lg(n), which doesn't deal with decimal precision stuff)

SansPapyrus683: wait no now i'm curious about why this is the case, think you could explain why?

TheGamingMousse: > wait no now i'm curious about why this is the case, think you could explain why?

ok actually the submissions i linked are a BIT of a bad example. in this case, it RTEs because ceil(log2(n)) evaluates to 0 when n = 1, so accessing the binlift array in any capacity RTEs cuz there aren't any rows...

a better example of when this would WA is attached [here](https://ide.usaco.guide/O-L4nyUb3Rzd1lL1q3S) (you can obv write ur code so that ceil(log2(n)) would work, but it's easier to just use 1 + (int)log2(n) or smth like that)

below is the write up i did for why it can be wrong:
<details><summary>

</summary>

Let $x$ be the number of rows in the table. Because the highest bit it covers is $x - 1$, at most we can lift $2^{x} - 1$ times. Now, say we set $x$ to be ceil(log2(n)). If n is a perfect power of 2, then $2^x $is equal to exactly $n$. So, this means that we can only lift $n-1$ times. However, if we need to lift $n$ times (e.g. in a range query), this is problematic.

</details>

SansPapyrus683: then why doesn't it happen when doing ancestor queries

SansPapyrus683: ok so i figured it out- this current method works for just lca so i'll leave it be

PR #4526, opened 2024-06-16
Update Digit_DP.problems.json

coolavy:
I removed Perfect Number because it has nothing to do with the topic and can be easily solved without it (as it's Div. 2 B).


envyaims (approved):
would be a good digit dp problem if n <= 10^18, what a shame

PR #4599, opened 2024-07-09
Update LIS.mdx

TheScarletPhoenix:
added code in Python

- [X] I have tested my code.

### DIFF
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -591,31 +591,34 @@ class PCB {
 
 ```py
 import sys
+
 input = sys.stdin.read
 
+
 def find_lis_length():
-    lis = 0
-    a = []
-    s = set()
+	lis = 0
+	a = []
+	s = set()
 
-    lines = input().strip().split('\n')
-    n = int(lines[0].strip())
-    for i in range(1, n+1):
-        x, y = map(int, lines[i].strip().split())
-        a.append((x, y))
-    a.sort(reverse=True, key=lambda x: x[1])
+	lines = input().strip().split("\n")
+	n = int(lines[0].strip())
+	for i in range(1, n + 1):
+		x, y = map(int, lines[i].strip().split())
+		a.append((x, y))
+	a.sort(reverse=True, key=lambda x: x[1])
 
-    for i in range(n):
-        if not s or max(s) < a[i][1]:
-            lis += 1
-        else:
-            s.discard(next(x for x in s if x > a[i][1]))
-        s.add(a[i][1])
+	for i in range(n):
+		if not s or max(s) < a[i][1]:
+			lis += 1
+		else:
+			s.discard(next(x for x in s if x > a[i][1]))
+		s.add(a[i][1])
 
-    return lis
+	return lis
+
 
 if __name__ == "__main__":
-    print(find_lis_length())
+	print(find_lis_length())
 ```
 
 </PySection>
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -593,16 +593,18 @@ class PCB {
 import sys
 
 input = sys.stdin.read
-
 lis = 0
 a = []
 s = set()
 
 lines = input().strip().split("\n")
 n = int(lines[0].strip())
+
 for i in range(1, n + 1):
 	x, y = map(int, lines[i].strip().split())
 	a.append((x, y))
+
+# finding the LIS of a reversed array = finding the LDS
 a.sort(reverse=True, key=lambda x: x[1])
 
 for i in range(n):
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -594,31 +594,25 @@ import sys
 
 input = sys.stdin.read
 
+lis = 0
+a = []
+s = set()
 
-def find_lis_length():
-	lis = 0
-	a = []
-	s = set()
+lines = input().strip().split("\n")
+n = int(lines[0].strip())
+for i in range(1, n + 1):
+    x, y = map(int, lines[i].strip().split())
+    a.append((x, y))
+a.sort(reverse=True, key=lambda x: x[1])
 
-	lines = input().strip().split("\n")
-	n = int(lines[0].strip())
-	for i in range(1, n + 1):
-		x, y = map(int, lines[i].strip().split())
-		a.append((x, y))
-	a.sort(reverse=True, key=lambda x: x[1])
+for i in range(n):
+    if not s or max(s) < a[i][1]:
+        lis += 1
+    else:
+        s.discard(next(x for x in s if x > a[i][1]))
+    s.add(a[i][1])
 
-	for i in range(n):
-		if not s or max(s) < a[i][1]:
-			lis += 1
-		else:
-			s.discard(next(x for x in s if x > a[i][1]))
-		s.add(a[i][1])
-
-	return lis
-
-
-if __name__ == "__main__":
-	print(find_lis_length())
+print(lis)
 ```
 
 </PySection>
--- a/content/4_Gold/LIS.mdx
+++ b/content/4_Gold/LIS.mdx
@@ -591,36 +591,31 @@ class PCB {
 
 ```py
 import sys
-
 input = sys.stdin.read
 
-lis = 0
-a = []
-s = set()
+def find_lis_length():
+    lis = 0
+    a = []
+    s = set()
 
+    lines = input().strip().split('\n')
+    n = int(lines[0].strip())
+    for i in range(1, n+1):
+        x, y = map(int, lines[i].strip().split())
+        a.append((x, y))
+    a.sort(reverse=True, key=lambda x: x[1])
 
-def main():
-	global lis, a, s
-	lines = input().strip().split("\n")
-	n = int(lines[0].strip())
-	for i in range(1, n + 1):
-		x, y = map(int, lines[i].strip().split())
-		a.append((x, y))
-	a.sort(reverse=True, key=lambda x: x[1])
+    for i in range(n):
+        if not s or max(s) < a[i][1]:
+            lis += 1
+        else:
+            s.discard(next(x for x in s if x > a[i][1]))
+        s.add(a[i][1])
 
-	# finding the LIS of a reversed array = finding the LDS
-	for i in range(n):
-		if not s or max(s) < a[i][1]:
-			lis += 1
-		else:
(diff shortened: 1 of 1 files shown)


content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,43 @@ class PCB {
...
+import sys
+
+input = sys.stdin.read
+
+lis = 0
+a = []
+s = set()
+
+
+def main():
```
SansPapyrus683, line 602: is the main method necessary

content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,43 @@ class PCB {
...
+	global lis, a, s
```
SansPapyrus683, line 603: no need to be global

content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,43 @@ class PCB {
...
+		x, y = map(int, lines[i].strip().split())
+		a.append((x, y))
+	a.sort(reverse=True, key=lambda x: x[1])
+
+	# finding the LIS of a reversed array = finding the LDS
+	for i in range(n):
+		if not s or max(s) < a[i][1]:
+			lis += 1
+		else:
+			s.discard(next(x for x in s if x > a[i][1]))
```
SansPapyrus683, line 616: how does this emulate sorted set behavior in python

SansPapyrus683 (changes requested):

content/4_Gold/LIS.mdx
TheScarletPhoenix: no, just included it for readability.
i can delete it though.

content/4_Gold/LIS.mdx
TheScarletPhoenix: ok, will change

content/4_Gold/LIS.mdx
TheScarletPhoenix: same logic as c++.

content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,41 @@ class PCB {
...
 </JavaSection>
+<PySection>
+
+```py
+import sys
+
+input = sys.stdin.read
+
+
+def find_lis_length():
```
SansPapyrus683, line 598: my guy you just changed the name :sob:

TheScarletPhoenix: wdym?

SansPapyrus683: execute the code directly
don't put it in some function then call the function
that's so roundabout

content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,37 @@ class PCB {
 ```
 
 </JavaSection>
+<PySection>
+
+```py
+import sys
+
+input = sys.stdin.read
```
SansPapyrus683, line 595: just use raw `input()`

content/4_Gold/LIS.mdx
SansPapyrus683: python sets are literally unordered

content/4_Gold/LIS.mdx
TheScarletPhoenix: ok

content/4_Gold/LIS.mdx
SansPapyrus683: wait can you explain this further
how do you manage to do this without bricking the time complexity

TheScarletPhoenix: Input Reading and Array Initialization: O(n)
Sorting: O(n log n)
Finding LIS using a Set: O(n)
So, the overall time complexity is O(n log n), same as the c++ code.

SansPapyrus683: please for the love of god can you not resolve the convo each time
> Finding LIS using a Set: O(n)

i see you're iterating throug the set each time tho

TheScarletPhoenix: yeah

envyaims (approved):
rawr

content/4_Gold/LIS.mdx
SansPapyrus683: you
you can't just gloss past that...

TheScarletPhoenix: i tbh don't know what u're looking for, but:

Insertion (s.add(a[i][1])): Inserting an element into a set takes average O(1) time, but in the worst case (due to rehashing), it can be O(k) where k is the size of the set. Since k <= n, this operation amortizes to O(1).

Deletion (s.discard(next(x for x in s if x > a[i][1]))): Deleting an element from a set can take up to O(k) time in the worst case, where k is the size of the set. Again, this amortizes to O(1) on average.

Overall: The for loop iterates n times, and for each iteration, operations on the set (s.add, s.discard) are O(1) on average. Therefore, the total time complexity for this part is O(n).

SansPapyrus683: > Since k <= n, this operation amortizes to O(1).

ok you're right, but i'm pretty sure this line of logic is flawed :skull: 

> Deleting an element from a set can take up to O(k) time in the worst case, where k is the size of the set. Again, this amortizes to O(1) on average.

elaborate

TheScarletPhoenix: 1. Average Case Time Complexity: O(1)

- When you delete an element from a set using s.remove(element) or s.discard(element), Python typically achieves this in constant time on average, O(1). This efficiency is due to the hash table's ability to directly locate and remove the element using its hash value.

2. Worst Case Time Complexity: O(k)

- In the worst-case scenario, the time complexity for deleting an element from a set can be O(k), where k is the size of the set. This situation arises when the hash table needs to be resized (rehashed) to accommodate more elements or when elements hash to the same slot, causing a collision chain to traverse.

- Hash Table Resizing (Rehashing): If the deletion operation triggers a resize of the hash table, all elements in the original hash table must be rehashed and inserted into a larger or differently structured hash table. This operation is linear with respect to the number of elements currently in the set (k).

- Collision Chains: In a hash table, elements that hash to the same index are stored in a linked list or similar structure (a collision chain). If many elements collide and are stored in the same bucket, removing an element from this chain may involve traversing multiple elements in the worst case.

3. Amortized Time Complexity: O(1)

- Despite the potential worst-case scenario of O(k) for deletion, in practice, deletions in Python sets are generally O(1) on average due to:
    - Efficient hash function distribution.
    - Hash table resizing being infrequent and amortized over many operations.
- Over many deletion operations, the occasional worst-case scenario is balanced out by numerous average-case O(1) operations. This amortized analysis is why Python sets are considered to provide O(1) average-time complexity for deletion operations.

ryanchou-dev (approved):

content/4_Gold/LIS.mdx
envyaims: The latter explanation by TheScarletPhoenix, which states that the deletion operation in a Python set has an average-case time complexity of \(O(1)\) and a worst-case time complexity of \(O(k)\), where \(k\) is the size of the set, contains some inaccuracies and flawed reasoning. Let's dissect it and point out where the logic falters:

\subsection*{1. Average-Case Time Complexity: \(O(1)\)}
The explanation correctly states that deletion in a hash table (which underlies Python sets) is typically \(O(1)\) on average. This is because hash tables allow for direct access to elements via their hash values.

\subsection*{2. Worst-Case Time Complexity: \(O(k)\)}
The claim that the worst-case time complexity for deleting an element from a set is \(O(k)\) is incorrect. The worst-case time complexity for deletion should be \(O(n)\), where \(n\) is the number of elements in the set.
The worst case occurs not due to hash table resizing but because of potential hash collisions. If many elements hash to the same index (forming a collision chain), traversing and finding the element to delete could theoretically take \(O(n)\) time, where \(n\) is the total number of elements in the set, not just the elements in the collision chain.

\subsection*{3. Hash Table Resizing (Rehashing)}
The explanation mistakenly associates the cost of resizing with the deletion operation. Resizing (rehashing) usually occurs during insertion when the load factor of the hash table exceeds a certain threshold. During deletion, resizing is less common and usually only happens when the hash table becomes very sparse, which is rare.
When resizing does occur, its cost is \(O(n)\) because every element must be rehashed and placed into a new table. However, this cost is amortized over many insertions or deletions, making the amortized cost per operation \(O(1)\).

\subsection*{4. Collision Chains}
While it is true that collisions can occur, the explanation should clarify that even in the worst-case scenario, where all elements hash to the same index, the cost of deletion is \(O(n)\) because the collision chain can contain all \(n\) elements of the set.
Modern hash tables, including Python's, use techniques to minimize collisions and keep chains short, but the worst-case complexity is still \(O(n)\).

\subsection*{5. Amortized Time Complexity: \(O(1)\)}
The reasoning that deletions are generally \(O(1)\) on average due to efficient hash function distribution and infrequent resizing is correct.
The explanation fails to adequately connect the \(O(k)\) claim with the amortized analysis. The correct analysis should emphasize that while individual operations can be expensive, the amortized cost over many operations remains low due to efficient hash table management.

\subsection*{Corrected Analysis}
\begin{itemize}
    \item \textbf{Average-Case Time Complexity}: \(O(1)\) for both insertion and deletion in a well-implemented hash table.
    \item \textbf{Worst-Case Time Complexity}: \(O(n)\) due to potential collisions and the need to traverse long chains in the worst-case scenario.
    \item \textbf{Amortized Time Complexity}: \(O(1)\) for both insertion and deletion, considering the infrequent and amortized cost of resizing.
\end{itemize}

By understanding these nuances, we can appreciate the performance characteristics of Python sets more accurately.

\end{document}

content/4_Gold/LIS.mdx
```diff
@@ -587,6 +587,32 @@ class PCB {
...
+n = int(input().strip())
+
+for i in range(n):
+	x, y = map(int, input().strip().split())
+	a.append((x, y))
+
+a.sort(reverse=True, key=lambda x: x[1])
+
+for i in range(n):
+	if not s or max(s) < a[i][1]:
```
dongliu0426, line 606: Since python sets are unordered, `max(s)` takes $O(|s|)$ time...

dongliu0426 (changes requested):

dongliu0426:
Did you test the code? Just submitted this on CSES and your code is outputting all 1s.

dongliu0426:
Hi!

Unfortunately, it appears that your code hasn't been tested and has several issues.

We'll have to close the PR for now--we strongly encourage you to create another PR following our guidelines.

If you have any further questions, feel free to reply.

PR #4615, opened 2024-07-15
fix proof in small to large merging

jasonzeng124:

- [x] I have tested my code.

### DIFF
--- a/content/5_Plat/Merging.mdx
+++ b/content/5_Plat/Merging.mdx
@@ -67,11 +67,14 @@ larger one of size $n$ takes $\mathcal{O}(m\log n)$ time.
 
 **Claim:** The solution runs in $\mathcal{O}(N\log^2N)$ time.
 
-**Proof:** When merging two sets, you move from the smaller set to the larger
-set. If the size of the smaller set is $X$, then the size of the resulting set
-is at least $2X$. Thus, an element that has been moved $Y$ times will be in a
-set of size at least $2^Y$, and since the maximum size of a set is $N$ (the
-root), each element will be moved at most $\mathcal{O}(\log N$) times.
+**Proof:** Suppose that you always merge from the smaller subtree into the
+larger subtree. Then, each element will be merged at most $O(\log N)$ times,
+because the subtree size will double every time the element is merged into a
+larger subtree. In total, we process at most $O(N\log N)$ elements.
+
+Note that when we use the merging technique in the code, each merge will take
+less time than in our proof. So, we will process at most $O(N \log N)$ elements,
+each taking at most $O(\log N)$ time.
 
 ### Full Code
 


TheGamingMousse (approved):
alter orz

SansPapyrus683:
ngl this seems about the same? idk

TheGamingMousse:
> ngl this seems about the same? idk

he explained it on discord but like, each element wouldn't necessarily be in a set of size $2^{x}$ (x = # of times moved) because not all elements are distinct

ryanchou-dev (approved):

content/5_Plat/Merging.mdx
```diff
@@ -67,11 +67,14 @@ larger one of size $n$ takes $\mathcal{O}(m\log n)$ time.
...
 
-**Proof:** When merging two sets, you move from the smaller set to the larger
-set. If the size of the smaller set is $X$, then the size of the resulting set
-is at least $2X$. Thus, an element that has been moved $Y$ times will be in a
-set of size at least $2^Y$, and since the maximum size of a set is $N$ (the
-root), each element will be moved at most $\mathcal{O}(\log N$) times.
+**Proof:** Suppose that you always merge from the smaller subtree into the
+larger subtree. Then, each element will be merged at most $O(\log N)$ times,
+because the subtree size will double every time the element is merged into a
+larger subtree. In total, we process at most $O(N\log N)$ elements.
```
SansPapyrus683, line 73: how is this more accurate then

jasonzeng124: Well, it is *true* (the subtree size will double every time an element is merged, under the formulation that small subtrees are merged into large subtrees)

content/5_Plat/Merging.mdx
```diff
@@ -67,11 +67,14 @@ larger one of size $n$ takes $\mathcal{O}(m\log n)$ time.
...
+
+Note that when we use the merging technique in the code, each merge will take
+less time than in our proof. So, we will process at most $O(N \log N)$ elements,
```
dongliu0426, line 76: > each merge will take less time than in our proof

why?

dongliu0426 (commented):
also, I don't see the issue with the original proof. It seems to be using the same analysis as yours?

content/5_Plat/Merging.mdx
jasonzeng124: > also, I don't see the issue with the original proof. It seems to be using the same analysis as yours?

I guess it's a really subtle (and kinda pedantic) difference lol.

To answer your question: 
Each merge will take less time than in our proof, because in our proof merges take O(min(s1,s2) log n) (s1,s2 are subtree sizes), and the size of the set is less than the size of the subtree, so therefore O(min(|set1|, |set2|) log n) must be faster.

content/5_Plat/Merging.mdx
SansPapyrus683: >  each element wouldn't necessarily be in a set of size $2^{x}$ (x = # of times moved) because not all elements are distinct

ryanchou-dev:
> > ngl this seems about the same? idk
> 
> he explained it on discord but like, each element wouldn't necessarily be in a set of at least size 2x (x = # of times moved) because not all elements are distinct

oops just re-read, didn't it say that each element would be in a size of **at least** $2^x$, not that it's exactly $2^x$?

jasonzeng124:
> > > ngl this seems about the same? idk
> > 
> > 
> > he explained it on discord but like, each element wouldn't necessarily be in a set of at least size 2x (x = # of times moved) because not all elements are distinct
> 
> oops just re-read, didn't it say that each element would be in a size of _at least_ 2x, not that it's exactly 2x?

if you merge two sets, the result is <= 2x, not >=2x :skull:

ryanchou-dev:
> > > > ngl this seems about the same? idk
> > > 
> > > 
> > > he explained it on discord but like, each element wouldn't necessarily be in a set of at least size 2x (x = # of times moved) because not all elements are distinct
> > 
> > 
> > oops just re-read, didn't it say that each element would be in a size of _at least_ 2x, not that it's exactly 2x?
> 
> if you merge two sets, the result is <= 2x, not >=2x 💀

wait don't the subtree sizes double when merging stl?

jasonzeng124:
lol

PR #4736, opened 2024-08-31
RURQ Focus Problem Solution Code

eysbutno:

- [x] I have tested my code.

i personally think this solution is way cleaner than the prev one
image

### DIFF
--- a/content/5_Plat/RURQ.mdx
+++ b/content/5_Plat/RURQ.mdx
@@ -169,13 +169,13 @@ struct Query {
 
 /**
  * T = tree node type, which will be long long
- * U = update type, which will be the Query struct
+ * We use the Query struct to manage updates.
  */
-template <class T, class U> class LazySegtree {
+template <class T> class LazySegtree {
   private:
 	const int sz;
-	vector<T> tree;  // tree[i] = sum of this node's range
-	vector<U> lazy;  // lazy[i] = lazy update for the range
+	vector<T> tree;      // tree[i] = sum of this node's range
+	vector<Query> lazy;  // lazy[i] = lazy update for the range
 
 	/** @return result of joining two tree nodes together */
 	inline T comb(T a, T b) { return a + b; }
@@ -193,7 +193,7 @@ template <class T, class U> class LazySegtree {
 	}
 
 	/** applies lazy update to t[v], places update at lz[v] */
-	void apply(int v, int len, U x) {
+	void apply(int v, int len, Query x) {
 		if (x.type == ADD) {
 			// if lazy[v]'s type is NONE or ADD, then we add to the range
 			// otherwise, we add to our lazy set value
@@ -203,7 +203,7 @@ template <class T, class U> class LazySegtree {
 				lazy[v] = Query{SET, lazy[v].val + x.val};
 			}
 			tree[v] += x.val * len;
-		} else if (query.type == SET) {
+		} else if (x.type == SET) {
 			// lazy set overrides any previous update
 			tree[v] = x.val * len;
 			lazy[v] = x;
@@ -220,7 +220,7 @@ template <class T, class U> class LazySegtree {
 		lazy[v] = Query{NONE, 0};
 	}
 
-	void range_update(int v, int l, int r, int ql, int qr, U x) {
+	void range_update(int v, int l, int r, int ql, int qr, Query x) {
 		if (qr < l || ql > r) { return; }
 		if (ql <= l && r <= qr) {
 			apply(v, r - l + 1, x);
@@ -248,7 +248,7 @@ template <class T, class U> class LazySegtree {
 	}
 
 	/** updates [ql, qr] with the update x */
-	void range_update(int ql, int qr, U x) {
+	void range_update(int ql, int qr, Query x) {
 		range_update(1, 0, sz - 1, ql, qr, x);
 	}
 
@@ -262,7 +262,7 @@ int main() {
 	vector<ll> a(n);
 	for (ll &i : a) { cin >> i; }
 
-	LazySegtree<ll, Query> st(a);
+	LazySegtree<ll> st(a);
 	for (int t = 0; t < q; t++) {
 		int type, a, b;
 		cin >> type >> a >> b;
--- a/content/5_Plat/RURQ.mdx
+++ b/content/5_Plat/RURQ.mdx
@@ -156,9 +156,11 @@ range-sum segment tree to solve the question.
 using namespace std;
 using ll = long long;
 
-// T = tree node type, which will be long long
-// U = update type, which will be array<ll, 2>
-// Updates are: {lazy_set, lazy_add}
+/**
+ * T = tree node type, which will be long long
+ * U = update type, which will be array<ll, 2>
+ * Updates are: {lazy_set, lazy_add}
+*/
 template <class T, class U> class LazySegtree {
   private:
 	static constexpr T ID = 0;          // neutral tree node value
@@ -210,7 +212,7 @@ template <class T, class U> class LazySegtree {
 	}
 
 	void range_update(int v, int tl, int tr, int ql, int qr, U x) {
-		if (qr < tl || ql > tr) return;
+		if (qr < tl || ql > tr) { return; }
 		if (ql <= tl && tr <= qr) {
 			apply(v, tr - tl + 1, x);
 		} else {
@@ -223,8 +225,8 @@ template <class T, class U> class LazySegtree {
 	}
 
 	T range_sum(int v, int l, int r, int ql, int qr) {
-		if (qr < l || ql > r) return ID;
-		if (l >= ql && r <= qr) return t[v];
+		if (qr < l || ql > r) { return ID; }
+		if (l >= ql && r <= qr) { return t[v]; }
 		pushdown(v, l, r);
 		int m = (l + r) / 2;
 		return comb(range_sum(2 * v, l, m, ql, qr),
--- a/content/5_Plat/RURQ.mdx
+++ b/content/5_Plat/RURQ.mdx
@@ -244,7 +244,7 @@ template <class T, class U> class LazySegtree {
 
   public:
 	LazySegtree(const vector<T> &a)
-	    : sz((int)a.size()), tree(4 * sz), lazy(4 * sz) {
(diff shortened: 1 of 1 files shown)


content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,125 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
+		if (type == 1) {
+			int x;
+			cin >> x;
+			st.range_update(a, b, {0, x});
+		} else if (type == 2) {
+			int x;
+			cin >> x;
+			st.range_update(a, b, {x, 0});
+		} else {
+			cout << st.range_sum(a, b) << "\n";
```
ryanchou-dev, line 271: space out some of this code

ryanchou-dev (commented):
looks good!

content/5_Plat/RURQ.mdx
TheGamingMousse: good point

TheGamingMousse: done

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,124 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
-	ll lzAdd;
-	ll lzSet;
-	node(){};
-} tree[maxN << 2];
-
-#define lc p << 1
-#define rc (p << 1) + 1
+// T = tree node type, which will be long long
+// U = update type, which will be array<ll, 2>
+// Updates are: {lazy_set, lazy_add}
```
SansPapyrus683, line 161: turn into doc comment

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,124 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
 		}
-		if (tree[rc].lzSet == 0) tree[rc].lzAdd += tree[p].lzAdd;
-		else {
-			tree[rc].lzSet += tree[p].lzAdd;
-			tree[rc].lzAdd = 0;
+		lz[v] = LZ_ID;
+	}
+
+	void range_update(int v, int tl, int tr, int ql, int qr, U x) {
+		if (qr < tl || ql > tr) return;
```
SansPapyrus683, line 213: braces

content/5_Plat/RURQ.mdx
SansPapyrus683: ok so upon further inspection it seems like you're trying to do two things at once with `U` which is... kinda sus
why not just have separate functions for setting and adding
yk, single responsibility and whatnot

SansPapyrus683 (changes requested):
\*kisses you on the tongue\*

content/5_Plat/RURQ.mdx
TheGamingMousse: shoot my b

content/5_Plat/RURQ.mdx
TheGamingMousse: reason i do two things with `U` at once are:
- shortens the code a lot (because `range_add` and `range_set` would have 90% of the same code)
- it makes templating significantly easier (altho ig its not super applicable here)
idk, I can change it if you want but I think the way it currently is works fine

SansPapyrus683: i don't see any purpose in making it a tempalte argument when it really only takes a two-element indexable thing...maybe a custom struct at that point?

TheGamingMousse: > i don't see any purpose in making it a tempalte argument when it really only takes a two-element indexable thing...maybe a custom struct at that point?

fair enough, I can do that later. do you still want range updates to have setting and adding to be different methods?

SansPapyrus683: i mean say i set both values of the array to some nonzero number
what would happen?

TheGamingMousse: range set always takes precedence

the way i wrote the code is so that `lz[v]` would only ever have either `set` or `add` be nonzero though

SansPapyrus683: could you do something like this then

```cpp
enum QueryType {
    SET, ADD
};

struct Query {
    QueryType type;
    int val;
};
```

TheGamingMousse: ok it took me a while b/c i was being a bit of a clown, and I did have to change up some stuff, but I think it looks pretty nice now

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,127 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
-using ll = long long;
+#include <bits/stdc++.h>
 using namespace std;
+using ll = long long;
 
-const int maxN = 2e5 + 5;
-
-int N, Q;
-int a[maxN];
+enum QueryType { ADD, SET, NONE };
```
SansPapyrus683, line 159: ayo a "none" query type?

TheGamingMousse: technically its not needed but i think its good b/c it specifies when there's no lazy update to be performed

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,127 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
-#define rc (p << 1) + 1
+/**
+ * T = tree node type, which will be long long
+ * U = update type, which will be the Query struct
+ */
+template <class T, class U> class LazySegtree {
+  private:
+	const int sz;
+	vector<T> t;
+	vector<U> lz;
```
SansPapyrus683, line 174: i guess `t` stands for `tree`, could you rename it to that
also what does `lz` actually contain
comments, variable names, whatever you think helps

TheGamingMousse: bet

TheGamingMousse: ok I added some comments - mostly with the assumption that they know how to write segtree

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,135 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
+		if (x.type == ADD) {
+			// if lazy[v]'s type is NONE or ADD, then we add to the range
+			// otherwise, we add to our lazy set value
+			if (lazy[v].type != SET) {
+				lazy[v] = Query{ADD, lazy[v].val + x.val};
+			} else {
+				lazy[v] = Query{SET, lazy[v].val + x.val};
+			}
+			tree[v] += x.val * len;
+		} else {
```
SansPapyrus683, line 206:
```suggestion
		} else if (query.type == SET) {
```

TheGamingMousse: idk its implied alr but why not

SansPapyrus683: it probably isn't called with none but
smth smth explicit > implicit

content/5_Plat/RURQ.mdx
```diff
@@ -152,135 +152,135 @@ Given the mechanics behind the `pushdown` function, all you need is a regular
...
-		tree[rc].val = (r - mid) * tree[p].lzSet;
-		tree[lc].lzAdd = tree[rc].lzAdd = 0;
-		tree[p].lzSet = 0;
-	} else if (tree[p].lzAdd != 0) {  // lazy: range add
-		if (tree[lc].lzSet == 0) tree[lc].lzAdd += tree[p].lzAdd;
-		else {
-			tree[lc].lzSet += tree[p].lzAdd;
-			tree[lc].lzAdd = 0;
+	/** pushes down lazy update to children of v */
+	void pushdown(int v, int l, int r) {
```
SansPapyrus683, line 214: why `pushdown` but `range_update` and `range_sum`

TheGamingMousse: oh shiiii i forgot to update it!

content/5_Plat/RURQ.mdx
```diff
@@ -148,139 +148,139 @@ situations might take place after any update:
...
-void add(int p, int l, int r, int a, int b, ll val) {
-	if (a > r || b < l) return;
-	if (a <= l && r <= b) {
-		tree[p].val += (r - l + 1) * val;
-		if (tree[p].lzSet == 0) tree[p].lzAdd += val;
-		else tree[p].lzSet += val;
-		return;
+  public:
+	LazySegtree(const vector<T> &a)
+	    : sz((int)a.size()), tree(4 * sz), lazy(4 * sz) {
```
SansPapyrus683, line 247:
```suggestion
	    : sz(a.size()), tree(4 * sz), lazy(4 * sz) {
```

content/5_Plat/RURQ.mdx
```diff
@@ -148,139 +148,138 @@ situations might take place after any update:
...
+	ll val = 0;
+};
 
-#define lc p << 1
-#define rc (p << 1) + 1
+/**
+ * T = tree node type, which will be long long
+ * U = update type, which will be the Query struct
+ */
+template <class T, class U> class LazySegtree {
```
SansPapyrus683, line 174: is `U` still necessary?

TheGamingMousse: i could get rid of the template stuff if u want

SansPapyrus683: uh just `U`, you can keep `T`

TheGamingMousse: kk gimme a sec

content/5_Plat/RURQ.mdx
```diff
@@ -193,7 +193,7 @@ template <class T, class U> class LazySegtree {
 	}
 
 	/** applies lazy update to t[v], places update at lz[v] */
-	void apply(int v, int len, U x) {
+	void apply(int v, int len, Query x) {
```
SansPapyrus683, line 196: could these be constant references

TheGamingMousse: i was thinking about that, will do

SansPapyrus683 (approved):

ryanchou-dev (approved):
good!

PR #4819, opened 2024-10-04
Python implementation of Z-algorithm

modelflat:

- [x] I have tested my code.

AC proof: https://judge.yosupo.jp/submission/239873

### DIFF
--- a/content/6_Advanced/String_Search.mdx
+++ b/content/6_Advanced/String_Search.mdx
@@ -177,8 +177,10 @@ using namespace std;
 
 vector<int> z_function(const string& s) {
 	vector<int> z(s.size());
+	z[0] = s.size();
 
-	for (int i = 1, l = 0, r = 0; i < s.size(); i++) {
+	int l = 0, r = 0;
+	for (int i = 1; i < s.size(); i++) {
 		z[i] = max(0, min(z[i - l], r - i + 1));
 
 		while (i + z[i] < s.size() && s[z[i]] == s[i + z[i]]) {
@@ -194,9 +196,9 @@ vector<int> z_function(const string& s) {
 int main() {
 	string s;
 	cin >> s;
-	const auto z = z_function(s);
+	vector<int> z = z_function(s);
 
-	for (int i = 0; i < s.size(); i++) {
+	for (int i = 1; i < s.size(); i++) {
 		if (z[i] + i == s.size()) {
 			cout << i << " ";
 		}
@@ -227,16 +229,13 @@ def z_function(s: str) -> List[int]:
 	return z
 
 
-def main():
-	s = input()
-	z = z_function(s)
-	for i in range(1, len(z)):
-		if z[i] + i == len(s):
-			print(i, end=" ")
-	print(len(s))
+s = input()
+z = z_function(s)
+for i in range(1, len(z)):
+	if z[i] + i == len(s):
+		print(i, end=" ")
+print(len(s))
 
-
-main()
 ```
 
 </PySection>
--- a/content/6_Advanced/String_Search.mdx
+++ b/content/6_Advanced/String_Search.mdx
@@ -202,7 +202,7 @@ int main() {
 	cin >> s;
 	vector<int> z = z_function(s);
 	for (int i = 1; i < s.size(); i++) {
-		if (z[i] + i == s.size()) { cout << i << " "; }
+		if (i + z[i] == s.size()) { cout << i << " "; }
 	}
 	cout << s.size() << endl;
 }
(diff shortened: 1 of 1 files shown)


content/6_Advanced/String_Search.mdx
```diff
@@ -197,6 +197,27 @@ int main() {
 ```
 
 </CPPSection>
+<PySection>
+
+```py
+from typing import List
+
+
+def z_function(s: str) -> List[int]:
```
SansPapyrus683, line 206: this seems to do something different from the c++ code?

TheGamingMousse: honestly i think it's better to have the z function code separate from the code that handles input n stuff

but yeah this is true

modelflat: @SansPapyrus683, yep, it's different. I think my version is better cause it's isolated from any specific problem and I/O stuff. Makes it easier to understand and reuse. We should probably refactor the C++ version in a similar fashion. What do you think?

SansPapyrus683: i agree, though this impl is supposed to be used to solve that "finding periods" problem above

modelflat: I updated it to solve finding periods. Also extracted Z-algo in C++ implementation into a separate function. Let me know what you think.

AC C++ https://cses.fi/problemset/result/10765731/
AC Python https://cses.fi/problemset/result/10765787/

content/6_Advanced/String_Search.mdx
```diff
@@ -188,15 +186,58 @@ int main() {
...
+	}
 
-		if (z[i] + i == s.size()) { cout << i << ' '; }
+	return z;
+}
+
+int main() {
+	string s;
+	cin >> s;
+	const auto z = z_function(s);
```
SansPapyrus683, line 197: plz no auto

content/6_Advanced/String_Search.mdx
```diff
@@ -188,15 +186,58 @@ int main() {
...
+		zi = min(r - i, z[i - l]) if i < r else 0
+		while i + zi < n and s[i + zi] == s[zi]:
+			zi += 1
+		if i + zi > r:
+			l, r = i, i + zi
+		z.append(zi)
+	return z
+
+
+def main():
```
SansPapyrus683, line 228: is this needed

SansPapyrus683 (changes requested):

content/6_Advanced/String_Search.mdx
SansPapyrus683: like putting the code inside a function instead of the file directly

content/6_Advanced/String_Search.mdx
```diff
@@ -175,9 +175,7 @@ We would then have these two cases:
 #include <bits/stdc++.h>
 using namespace std;
 
-int main() {
-	string s;
-	cin >> s;
+vector<int> z_function(const string &s) {
```
TheGamingMousse, line 178: are we defining `z[0]` as being `n` or `0`? cuz the python one defines it as `n` (which I think makes more sense, IMO)

content/6_Advanced/String_Search.mdx
```diff
@@ -175,9 +175,7 @@ We would then have these two cases:
...
 	vector<int> z(s.size());
 
 	for (int i = 1, l = 0, r = 0; i < s.size(); i++) {
```
TheGamingMousse, line 181: i think it would look nicer if the `l = 0` and `r = 0` came before the loop, like how it is in the python version

content/6_Advanced/String_Search.mdx
modelflat: I prefer `main` to putting stuff in global scope because it has a local scope, thus avoiding clashes between variables. However I think for the brevity purposes we can get rid of `main`. Updated.

content/6_Advanced/String_Search.mdx
modelflat: Fixed!

modelflat:
Addressed comments, thanks @SansPapyrus683 @TheGamingMousse !

AC Python updated https://cses.fi/problemset/result/10772986/
AC C++ updated https://cses.fi/problemset/result/10773040/

SansPapyrus683:
yeah np
btw when you're done you can click the "resolve convo" button at the bottom

content/6_Advanced/String_Search.mdx
```diff
@@ -175,28 +175,67 @@ We would then have these two cases:
...
+def z_function(s: str) -> List[int]:
+	n = len(s)
+	z = [n]
+	l, r = 0, 0
+	for i in range(1, n):
+		zi = min(r - i, z[i - l]) if i < r else 0
+		while i + zi < n and s[i + zi] == s[zi]:
+			zi += 1
+		if i + zi > r:
+			l, r = i, i + zi
```
SansPapyrus683, line 228: noticing a small diff in impl between this while loop & the one in the c++ version

SansPapyrus683 (approved):

TheGamingMousse (approved):

PR #4847, opened 2024-10-17
py sol for Cycle Finding

freakin23:

- [x] I have tested my code.

AC: https://cses.fi/paste/2dc3861cb9075445a62428/

I used `INF` because using `float('inf')` results in WA for 3 TCs

### DIFF
--- a/content/6_Advanced/SPneg.mdx
+++ b/content/6_Advanced/SPneg.mdx
@@ -214,7 +214,7 @@ there is an negative cycle.
 using namespace std;
 using ll = long long;
 
-struct edge {
+struct Edge {
 	int from, to;
 	ll weight;
 };
@@ -223,16 +223,16 @@ const int MAXN = 2505;
 
 int n, m, parent[MAXN];
 ll dist[MAXN];
-vector<edge> graph;
+vector<Edge> graph;
 
-void BellmanFord(int source) {
+void bellman_ford(int source) {
 	fill(parent + 1, parent + n + 1, 0);
 	fill(dist + 1, dist + n + 1, 1e18);
 	dist[source] = 0;
 	int last_node_updated;
 	for (int i = 1; i <= n; i++) {
 		last_node_updated = -1;
-		for (edge &e : graph) {
+		for (Edge &e : graph) {
 			if (dist[e.from] + e.weight < dist[e.to]) {
 				dist[e.to] = dist[e.from] + e.weight;
 				parent[e.to] = e.from;
@@ -268,7 +268,7 @@ int main() {
 		cin >> a >> b >> c;
 		graph.push_back({a, b, c});
 	}
-	BellmanFord(1);
+	bellman_ford(1);
 }
 ```
 
--- a/content/6_Advanced/SPneg.mdx
+++ b/content/6_Advanced/SPneg.mdx
@@ -365,7 +365,7 @@ public class CyclesFinding {
 <PySection>
 
 ```py
-INF = 10**18
+INF = 10**18 # float('inf') will result in WA for some Test Cases
 
 
 def bellman_ford(source: int):
--- a/content/6_Advanced/SPneg.mdx
+++ b/content/6_Advanced/SPneg.mdx
@@ -367,36 +367,37 @@ public class CyclesFinding {
 ```py
 INF = 10**18
 
+
 def bellman_ford(soure: int):
-    dist[soure] = 0
-    last_node_update = 0
-    for i in range(1, n + 1):
-        last_node_update = -1
-        for (u, v, w) in graph:
-            if dist[u] + w < dist[v]:
-                dist[v] = dist[u] + w
-                parent[v] = u
-                last_node_update = v
+	dist[soure] = 0
+	last_node_update = 0
+	for i in range(1, n + 1):
+		last_node_update = -1
+		for (u, v, w) in graph:
+			if dist[u] + w < dist[v]:
+				dist[v] = dist[u] + w
+				parent[v] = u
+				last_node_update = v
 
-    if last_node_update == -1:
-        print("NO")
-        exit()
-    else:
-        print("YES")
-        cycle = []
-        for i in range(1, n - 1):
-            last_node_update = parent[last_node_update]
+	if last_node_update == -1:
+		print("NO")
+		exit()
+	else:
+		print("YES")
+		cycle = []
+		for i in range(1, n - 1):
+			last_node_update = parent[last_node_update]
 
-        x = last_node_update
-        while True:
-            cycle.append(x)
-            if x == last_node_update and len(cycle) > 1:
-                break
-            x = parent[x]
+		x = last_node_update
+		while True:
+			cycle.append(x)
+			if x == last_node_update and len(cycle) > 1:
+				break
+			x = parent[x]
 
-        cycle.reverse()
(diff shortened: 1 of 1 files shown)


freakin23:
Just wanna ask do you want me to change the function name "BellmanFord" to "bellman_ford" used in C++ solution.

TheGamingMousse:
> Just wanna ask do you want me to change the function name "BellmanFord" to "bellman_ford" used in C++ solution.

that would be nice

also maybe rename the `edge` struct to `Edge`

freakin23:
> > Just wanna ask do you want me to change the function name "BellmanFord" to "bellman_ford" used in C++ solution.
> 
> that would be nice
> 
> also maybe rename the `edge` struct to `Edge`

Okay

content/6_Advanced/SPneg.mdx
```diff
@@ -362,6 +362,57 @@ public class CyclesFinding {
 ```
 
 </JavaSection>
+<PySection>
+
+```py
+INF = 10**18
```
SansPapyrus683, line 368: you know

content/6_Advanced/SPneg.mdx
```diff
@@ -362,6 +362,57 @@ public class CyclesFinding {
...
+
+
+def bellman_ford(soure: int):
```
SansPapyrus683, line 371: > soure

typo?

content/6_Advanced/SPneg.mdx
freakin23: I already mentioned it above that using float('inf') will result in WA for 3 TCs.

content/6_Advanced/SPneg.mdx
freakin23: my bad.

content/6_Advanced/SPneg.mdx
SansPapyrus683: so put it in a comment, the readers don't read your pr descriptions

freakin23: done

content/6_Advanced/SPneg.mdx
```diff
@@ -362,6 +362,57 @@ public class CyclesFinding {
 ```
 
 </JavaSection>
+<PySection>
+
+```py
+INF = 10**18  # float('inf') will result in WA for some Test Cases
```
SansPapyrus683, line 368: wrong answer? you mean tle?

content/6_Advanced/SPneg.mdx
```diff
@@ -362,6 +362,57 @@ public class CyclesFinding {
...
+		cycle.reverse()
+		print(*cycle)
+
+
+n, m = map(int, input().split())
+
+graph = []
+parent = [0] * (n + 1)
+dist = [INF] * (n + 1)
+
```
SansPapyrus683, line 407:
```suggestion
```

SansPapyrus683 (changes requested):

content/6_Advanced/SPneg.mdx
freakin23: No, I mean Wrong Answer

freakin23: The answer should be Yes along with the nodes but it results in straight No.

SansPapyrus683: hm
that seems a bit unintuitive, do you know why?

freakin23: not really, I haven't put much thought into it.

freakin23: got it .... it's because of weight constraint [-1e9, 1e9]

freakin23: I think the comment shouldn't be appropriate.

freakin23: the thing is that float('inf') doesn't give any real value number. so in the `if` condition, it will not pass through it. 
for that, we need some real value.

freakin23: using `sys.maxsize` works. do you want me to use that.

ryanchou-dev: Might be worth just defining a large constant.

Also the reason why `float('inf')` doesn't work is that any value plus `inf` is just `inf`.

`float('inf') + 10 = float('inf')`

freakin23: yes this is the issue with some TCs.

SansPapyrus683: > Also the reason why `float('inf')` doesn't work is that any value plus `inf` is just `inf`.

i mean yeah, that's kinda just how infinity works

freakin23: @SansPapyrus683 is it better to use `sys.maxsize`

SansPapyrus683: no i feel we need to figure out why float inf doesn't work

SansPapyrus683: he just mentioned that float inf + any number is still inf
like if the code breaks if that happens (which is expected behavior math-wise), i feel like something's wrong

freakin23: even if we add this condition `(w < 0 and dist[u] == float('inf') and dist[v] == float('inf'))` alongside the other one.
we still cannot overcome the next statement, which is to update the dist value which will still be infinity and that's where it is going wrong.

TheGamingMousse: idk don't most people just define a really big constant for INF anyways lol

SansPapyrus683: > which is to update the `dist` value which will still be infinity and that's where it is going wrong.

ok but it _should_ be infinity, no? unless the algorithm itself relies on a big number instead of actual infinity

freakin23: No, the updated value should be less than infinity then only it will work.  This is exactly where it is messed up.

freakin23: Unless it is some real number it is not going to work.

ryanchou-dev: > > which is to update the `dist` value which will still be infinity and that's where it is going wrong.
> 
> ok but it _should_ be infinity, no? unless the algorithm itself relies on a big number instead of actual infinity

My understanding is that it should still be updated because $w$ can be negative. So, the condition $\text{dist}_u + w < \text{dist}_v$ wouldn't work as intended.

freakin23: we are all coming to the same conclusion: `float('inf')` cannot be used here bcz it will cause trouble in updating `dist` values

SansPapyrus683: image

freakin23: I already did that with this condition but it resulted in WA for some other test cases.
Screenshot from 2024-10-29 10-10-19
Screenshot from 2024-10-29 10-10-39

ryanchou-dev (approved):

TheGamingMousse (approved):

freakin23:
@SansPapyrus683 do you want to address some changes

SansPapyrus683 (approved):

PR #4859, opened 2024-10-24
update PIE.mdx

freakin23:

- [x] I have tested my code.

The previous value of `VALMAX`  was too high thus resulting in TLE, therefore I changed it accordingly.


freakin23:
I realized that there was a good chance that the 20th value would cause some trouble with the page width, so I changed it back. Just changing the number is enough.

ryanchou-dev (approved):
👍 thank you for catching these!

ryanchou-dev:
> I realized that there was a good chance that the 20th value would cause some trouble with the page width, so I changed it back. Just changing the number is enough.

I don't think it makes too much of a difference, the table & latex will overflow nonetheless.

freakin23:
> 👍 thank you for catching these!

Ur welcome 🤗

SansPapyrus683 (approved):
mihnea strikes again

PR #4869, opened 2024-10-27
correct C++ code / py sol for Maximum Number of Visible Points

freakin23:

- [x] I have tested my code.

The previous C++ was resulting in `RE` bcz we were excluding the case when `angles` array is empty and we are accessing the element which is not even there in for loop.

AC (Py): https://leetcode.com/problems/maximum-number-of-visible-points/submissions/1435032380/
AC (C++): https://leetcode.com/problems/maximum-number-of-visible-points/submissions/1435035886/

### DIFF
--- a/content/5_Plat/Geo_Pri.mdx
+++ b/content/5_Plat/Geo_Pri.mdx
@@ -514,9 +514,7 @@ class Solution {
 			angles.push_back(atan2(p[1] - location[1], p[0] - location[0]));
 		}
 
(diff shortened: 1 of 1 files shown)


freakin23:
On second thought we can also tackle the issue by the current for loop condition to `i < min((int)angle.size(), (int)points.size());` 
Let me know which one you prefer.

SansPapyrus683 (approved):

PR #4970, opened 2024-12-10
hotfix

Keys9873:

- [x] I have tested my code.

### DIFF
--- a/content/6_Advanced/DP_SOS.mdx
+++ b/content/6_Advanced/DP_SOS.mdx
@@ -66,7 +66,7 @@ for (int i = 0; i < (1 << n); i++) {
 We can speed this up if we iterate over only subsets of the current mask and add up all of the those values to get the sum over subsets for a particular mask.
 
 The difference comes from the fact that in the first example we iterate over every pair of subsets which takes $(2^n)^2$ time and the second we iterate directly over the subsets for each mask. This means each mask is only visited by $2^{n - k}$ other masks where k is the number of elements of the mask.
-This means that the total time complexity is O$\sum_0^n {n \choose k} \cdot 2^{n - k} = 3^n$).
+This means that the total time complexity is $\mathcal{O} (\sum_0^n {n \choose k} \cdot 2^{n - k} = 3^n)$.
 
 <LanguageSection>
 <CPPSection>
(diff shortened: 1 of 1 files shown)


content/6_Advanced/DP_SOS.mdx
```diff
@@ -77,7 +77,10 @@ vector<int> a(1 << n);
 vector<int> sos(1 << n);
 for (int i = 0; i < (1 << n); i++) {
 	// iterate over all subsets of i directly
-	for (int j = (i - 1) & i; j >= 0; j = (j - 1) & i) { sos[i] += a[j]; }
+	for (int j = (i - 1) & i;; j = (j - 1) & i) {
```
Sosuke23, line 80: What was wrong with previous `for loop` ?

TheGamingMousse, line 80: i might be tweaking but would `j > 0` as a loop condition do the same thing here

Sosuke23: I think the previous one was right it doesn't need any changes.

TheGamingMousse: envy said it was wrong so he’s prolly right

Keys9873: j > 0 would not work because we want the loop to run when j = 0

The previous condition is wrong because when j = 0, the value will be updated to be (0 - 1) & i which is equal to i. Then when the loop condition is being checked, it would pass since i >= 0.

TheGamingMousse: oh lolz

TheGamingMousse (approved):

content/6_Advanced/DP_SOS.mdx
Sosuke23: Why not add the condition `j != 0` in the loop only

TheGamingMousse: read his comment 💀

Sosuke23: my bad 😅

Sosuke23: great observation @Keys9873

Sosuke23 (approved):

PR #5108, opened 2025-02-08
alternate PST implementation

eysbutno:

- [x] I have tested my code.

wrote up an indice based PST implementation that is roughly 2x faster than the existing one; also just added a lil warning to be careful with pointers

AC:
image


TheGamingMousse:
image

content/6_Advanced/Persistent.mdx
```diff
@@ -269,10 +272,129 @@ int main() {
 }
 ```
 
+<Warning>
+
+This implementation is not garbage collected, so the tree nodes aren't deleted even if the instance of the segment tree is taken out of scope.
+
+</Warning>
```
SansPapyrus683, line 279: no pointers, does this warning still apply? or is this meant for the above one

SansPapyrus683 (approved):

content/6_Advanced/Persistent.mdx
TheGamingMousse: warning is only for pointers (since vectors are deleted when taken out of scope); should i put the warning above the impl?

TheGamingMousse: ok i just put the warning above the impl inside language section cuz i think that makes the most sense

PR #5210, opened 2025-03-16
Continue removing `twin.macro`

thecodingwizard:
Followup to #5200


SansPapyrus683:
is mit on spring break

PR #5231, opened 2025-03-23
Improve comments in sumdiv.mdx

kokosha:
The comment of a is coprime with 10^9+7 is not correct.

- [x] I have tested my code.

### DIFF
--- a/solutions/gold/sumdiv.mdx
+++ b/solutions/gold/sumdiv.mdx
@@ -117,15 +117,17 @@ int main() {
 
 		if (a > 1 && ((a - 1) % MOD) != 0) {
 			// if a is greater than 1 and a - 1 is not divisible by 10^9 + 7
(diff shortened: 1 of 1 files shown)


kokosha:
In the line before, we have (a - 1) % MOD) != 0.
A better comment would be:  if a - 1 is not coprime to 10^9+7 , this also means that a is a prime at the form of k * (10^9+7) + 1, so we can just multiply by (b+1).

SansPapyrus683:
wait isn't the if statement testing for divisibility here

kokosha:
yeah, i think is better to change for:  if a - 1 is divisible by 10^9+7 , this also means that a is a prime at the form of k * (10^9+7) + 1, so we can just multiply by (b+1).

SansPapyrus683:
made a change, is this what you meant?

SansPapyrus683 (approved):

PR #5235, opened 2025-03-24
Upgrade to Tailwind 4

thecodingwizard:


SansPapyrus683:
gatsby upgrade coming soon????

thecodingwizard:
Very unlikely :p Gatsby is barely maintained anymore.

I think we should consider migrating to Astro, Nextjs, React Router, or some other web framework at some point.

SansPapyrus683:
if next is on the table i would be willing to help
but like that would seem like such a gargantuan task

thecodingwizard:
Yeah, that is my concern, too :/

I think Next.js is a pretty reasonable choice. We will likely have to migrate off Gatsby at _some_ point -- it doesn't support React 19 -- but admittedly it isn't particularly urgent for us to do so.

PR #5243, opened 2025-03-25
Remove announcements; clean up dashboard

thecodingwizard:
I removed announcements since we haven't used it in a while, and I'm trying to simplify this codebase as much as possible (by removing unnecessary features).

I also updated the dashboard to only render in the client, after the page finishes loading.


SansPapyrus683:
ok so i know nothing about webdev
but shouldn't doing things server side be better?

thecodingwizard:
In this particular case, since the dashboard content can only be properly rendered after user data has loaded, it's better for it to be rendered in the client only. The Activity section might also have some hydration mismatch / rendering issues since it changes day-to-day but Gatsby server rendering is only done once at build time (which might have happened on a previous day)

PR #5275, opened 2025-04-07
update solutionMetadata

Sosuke23:

- [ ] I have tested my code.

### DIFF
--- a/content/2_Bronze/Intro_Graphs.problems.json
+++ b/content/2_Bronze/Intro_Graphs.problems.json
@@ -10,7 +10,8 @@
       "isStarred": false,
       "tags": ["Permutation"],
       "solutionMetadata": {
-        "kind": "internal"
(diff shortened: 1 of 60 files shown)


Sosuke23:
**Note:
If there are duplicate problems, and both are in different modules and if one of them is a focus problem and has kind: "in-module" while the other one in the different module is "internal"... it is better to keep the first one also "internal" instead of "in-module"  to avoid build error.** 

it took lot of time to fix the build issue.

SansPapyrus683:
hm in that case would it be better to remove the practice problem listed?

Sosuke23:
> hm in that case would it be better to remove the practice problem listed?

I thought of doing but since it is very less like around 10 ig that's why I kept on changing it

SansPapyrus683 (approved):

PR #5280, opened 2025-04-10
it's all about base edi

TheGamingMousse:

- [x] I have tested my code.

lmk if i should make the code reflect the non-2ptr approach

i think that might be a bit more suitable

### DIFF
--- a/content/2_Bronze/Conclusion.problems.json
+++ b/content/2_Bronze/Conclusion.problems.json
@@ -83,8 +83,8 @@
       "isStarred": false,
       "tags": [],
       "solutionMetadata": {
-        "kind": "USACO",
-        "usacoId": "509"
+        "kind": "internal",
+        "hasHints": true
       }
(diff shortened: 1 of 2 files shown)


Sosuke23 (approved):
orz 🔥

Sosuke23:
maybe include both approach...

but I think the current one is good. it's bronze btw and 2-ptr is not that hard for anyone to not understand

Sosuke23:
maybe include both approach...

but I think the current one is good. it's bronze btw and 2-ptr is not that hard for anyone to not understand

SansPapyrus683:
should we be linking to a silver module here

Sosuke23:
> > should we be linking to a silver module here
> 
> yeah that's my concern with only writing the 2ptr approach... i'll probably write the std::set one in a bit

maybe add the same problem in silver also and write a separate editorial about it

PR #5302, opened 2025-04-20
Matrix Exponentiation Revamp 1

HaccerKat:

- [x] I have tested my code.

I think there is still more that can be added, but this is a good start for now.


content/5_Plat/Matrix_Expo.mdx
```diff
@@ -170,6 +174,168 @@ else:
...
+## Implementation
+
+**Time complexity:** $\mathcal{O}(k^3 \log n)$
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+typedef long long ll;
+const int mod = 1e9;
+template <typename T> void matmul(vector<vector<T>> &a, vector<vector<T>> b) {
```
TheGamingMousse, line 235: why not `const vector<vector<T>> &b`

content/5_Plat/Matrix_Expo.mdx
```diff
@@ -170,6 +174,168 @@ else:
...
+		}
+	}
+
+	a = c;
+}
+
+template <typename T> struct Matrix {
+	vector<vector<T>> mat;
+	Matrix() {}
+	Matrix(vector<vector<T>> a) { mat = a; }
```
TheGamingMousse, line 253: why not const ref?

content/5_Plat/Matrix_Expo.mdx
```diff
@@ -170,6 +174,168 @@ else:
...
+	while (tt--) { solve(); }
+}
+```
+</CPPSection>
+
+</LanguageSection>
+
+## Takeaway
+
+The takeaway here is not the exact matrix used in linear recurrences, but the method in deriving it. The process of thinking about a vector before and after applying $M$, then deducing $M$ through logic, is a technique that generalizes far beyond standard linear recurrences. For example, see if you can find the equation needed and correct matrix (using $k = 5$ as an example) to solve this modified recurrence: $a_i = c_1a_{i-1} + c_2a_{i-2} + ... + c_ka_{i-k} + C$ for $i > k$, where $C$ is a given constant.
```
TheGamingMousse, line 317: maybe put the recurrence on a new line to improve readability

TheGamingMousse (commented):
peng orz

content/5_Plat/Matrix_Expo.mdx
HaccerKat: doing that gets compile error lol
c++ operator overloading just works like that

content/5_Plat/Matrix_Expo.mdx
HaccerKat: i made this template very general (allows modint for example, which i didnt use in this code)
it is not strictly necessary in this problem but i like the template

content/5_Plat/Matrix_Expo.mdx
TheGamingMousse: oh right, im dumb 💀

content/5_Plat/Matrix_Expo.mdx
TheGamingMousse: alr makes sense

TheGamingMousse (approved):
someone better at math than me should prolly take a look, but overall module looks great

content/5_Plat/Matrix_Expo.mdx
```diff
@@ -170,6 +174,177 @@ else:
...
+## Implementation
+
+**Time complexity:** $\mathcal{O}(K^3 \log N)$
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+typedef long long ll;
+
+const int mod = 1e9;
```
SansPapyrus683: rename to `MOD`

SansPapyrus683 (approved):
made some changes
can you check ac again

PR #5309, opened 2025-04-26
Added java solution for Codeforces problem 1846D

gsmnit:
D. Rudolph and Christmas Tree

- [x ] I have tested my code.
- [ x] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions).

### DIFF
--- a/solutions/bronze/cf-1846D.mdx
+++ b/solutions/bronze/cf-1846D.mdx
@@ -113,77 +113,73 @@ import java.io.*;
 import java.util.*;
 
 public class RudolphAndChristmasTree {
-  static BufferedReader br;
-  static PrintWriter pw;
-  static StringTokenizer st;
+	static BufferedReader br;
+	static PrintWriter pw;
+	static StringTokenizer st;
 
-  static int nextInt() throws IOException {
-    return Integer.parseInt(nextToken());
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
honestly this fits more as a user solution, as it isn't consistent w/ the c++ and python versions

TheGamingMousse:
> Yes this solution is slightly different from current internal solution. i can add explanation for this solution if required
> 
> ## Explanation 2
> total area that we have to paint is equal to area of all branches or triangles (totalAreaToColor = n d h 2 ). but every branch or triangle don't give d h 2 contribution in total area that we need to paint. some area of branch is converd by other branch for every branch / triangle we need to calculate and subtract this area from totalAreaToColor. this area too is isoscale triangle. if we have two branche A and B. if A coverd B. coverd area or triangle will have height (H) = height of base of B + d - height of base of A. this covered triangle is similar triangle with B. so we can calculate base of this triangle (B) = H d h . so the area that we need to calculate and subtract from totalAreaToColor for each branch or triangle = H B 2 . this approach is used in java solution
> 
> otherwise will add this as user solution.

i think this is more suitable as a user solution yeah

PR #5317, opened 2025-05-02
Fix solutions/platinum/ioi-11-race.mdx

0legmak:
Returning zero highways does not make sense for this problem. Both original and fixed solutions pass submission tests on oj.uz, probably because there is no test for k=1

- [x] I have tested my code.

### DIFF
--- a/solutions/platinum/ioi-11-race.mdx
+++ b/solutions/platinum/ioi-11-race.mdx
@@ -81,7 +81,6 @@ void small_to_large_merge(int u, int p) {
 }
 
 int best_path(int n, int k, int edges[][2], int weights[]) {
-	if (k == 1) { return 0; }
 	N = n;
 	K = k;
 	ret = INT_MAX;


Sosuke23:
I think there is nothing wrong in the previous solution. If you already know that for k = 1 the solution will be 0 then why not print it at the moment instead of doing all the process.

0legmak:
> I think there is nothing wrong in the previous solution. If you already know that for k = 1 the solution will be 0 then why not print it at the moment instead of doing all the process.

A path cannot consist of zero highways. Also consider this input:
N=2
K=1
H=0 1
L=1
The expected result is 1, not 0.

Sosuke23:
> > I think there is nothing wrong in the previous solution. If you already know that for k = 1 the solution will be 0 then why not print it at the moment instead of doing all the process.
> 
> A path cannot consist of zero highways. Also consider this input: N=2 K=1 H=0 1 L=1 The expected result is 1, not 0.

I get it. You are right about it. Thanks for pointing it out.

TheGamingMousse:
yeah having that if statement is a bit silly lol, thanks for pointing it out

PR #5335, opened 2025-05-23
toggle-appears-in

meher0704:

- [ x] I have tested my code.
- [ x] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).

### DIFF
--- a/src/components/ProblemsPage/ProblemHits.tsx
+++ b/src/components/ProblemsPage/ProblemHits.tsx
@@ -1,42 +1,47 @@
-import { Link } from 'gatsby';
-import { BaseHit, Hit } from 'instantsearch.js';
-import * as React from 'react';
-import { Highlight, useHits } from 'react-instantsearch';
-import { moduleIDToSectionMap } from '../../../content/ordering';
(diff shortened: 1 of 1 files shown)


Sosuke23:
Maybe a better name is needed, like "Show Tags"  and instead of clicking on the name add an "on/off" button like something beside it. if it is possible

Sosuke23:
the best way is to add a drop down named "Show Tags" with options "Yes" and "No"  just like how it is used for difficulty, modules, and all.

src/components/ProblemsPage/ProblemHits.tsx
```diff
@@ -68,24 +73,6 @@ function ProblemHit({ hit }: ProblemHitProps) {
           <ProblemStatusCheckbox problem={problem} size="large" />
         </ConfettiProvider>
       </div>
-      {/* <div>
-        <a
-          href={hit.url}
```
Sosuke23, line 73: dont remove any commented lines without confirming from anyone, it may be no use of you but if it is there then just keep it there only

meher0704 (commented):
so just don't remove commented code and use a drop down tag instead

Issue #5348, opened 2025-05-29
negative time


idksarah:
![Image](https://github.com/user-attachments/assets/994d8310-e07b-4316-88d0-37578bcdd944)

Sosuke23:
its working well in others

bqi343:
Let's keep this open for a bit longer in case the OP has additional detail to provide. We have had other reports about this (like #5169, #4499, #3734). 

Unfortunately, I don't think there is really anything we can do without additional info. I think the usual cause of issues like this is some issue with your browser / browser cache / browser extension or your firebase connection. Do you see any errors when you open the browser console?

@thecodingwizard do you have any other suggestions?

idksarah:
hi sorry i would check the console or try a different browser but since it only appears after breaking a streak, i can't for the next 48 hours.
extra info:
- im using chrome on windows
- error only appears when i don't have a streak
- error disappears after reloading
- occurs on both my laptop and pc

bqi343:
@Sosuke23 can you provide screenshots?

bqi343:
@idksarah is this if you leave the same window open for a long time (like a day)?

Sosuke23:
> [@Sosuke23](https://github.com/Sosuke23) can you provide screenshots?

![Image](https://github.com/user-attachments/assets/ca4606ed-f5d3-4f31-9768-6ce6894f079a)

bqi343:
hm ok I've never seen that

SansPapyrus683:
should we close this after the pr

bqi343:
sure. @Sosuke23 please open a separate issue for anything unrelated to the photo timer

PR #5350, opened 2025-06-01
Fix negative time in daily streak component by clamping delta time to 0

emhgit:

- [x] I have tested my code.

Linked issue: https://github.com/cpinitiative/usaco-guide/issues/5348

The negative time issue occurred because Date.now() in the useEffect hook could sometimes be greater than tomorrowMilliseconds, causing a negative delta and rendering negative time in the countdown. I fixed this by clamping the minimum value of the delta to zero, which prevents the countdown from displaying negative values.

### DIFF
--- a/src/components/Dashboard/DailyStreak.tsx
+++ b/src/components/Dashboard/DailyStreak.tsx
@@ -13,15 +13,16 @@ const ComeBackTimer = ({ tomorrowMilliseconds }) => {
 
   React.useEffect(() => {
     const interval = setInterval(() => {
-      setMilliseconds(tomorrowMilliseconds - Date.now());
+      setMilliseconds(Math.max(0, tomorrowMilliseconds - Date.now()));
     }, 1000);
     return () => clearInterval(interval);
   }, []);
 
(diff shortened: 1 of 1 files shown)


SansPapyrus683:
tysm for the fix! seems a bit like a hack, though? could just be me

emhgit:
> tysm for the fix! seems a bit like a hack, though? could just be me

it is sort of “hacky”, but it’s the best i can come up with since i’m new to the codebase. as i get more familiar with it, maybe i can produce a better solution

SansPapyrus683:
why would date.now() be greater than tmrwMilliseconds?

emhgit:
> why would date.now() be greater than tmrwMilliseconds?

there are two cases i can think of:

1. the unlock time for the photo is reached and the user stays on the page, causing the date.now() to exceed the tmrwMilliseconds
2. the data for the lastVisitDate is faulty, which causes a calculation error

both of these would cause the delta to be negative.

SansPapyrus683 (approved):
alright then

PR #5374, opened 2025-06-11
Cover it everything

pleth9:

- [x] I have tested my code.

### DIFF
--- a/solutions/silver/cf-1176E.mdx
+++ b/solutions/silver/cf-1176E.mdx
@@ -17,7 +17,7 @@ Since BFS visits every node at the same level before moving on, edges must conne
 
 The categories are disjoint and exhaustive, so one group must have less than or equal to $\left \lfloor \frac{n}{2} \right \rfloor$ nodes. We can output the smaller of the two sets to satisfy the problem's requirements.
 
-Another interpretation is that the BFS traversal constructs a bipartite spanning tree, whose level-based partition guarantees that one side has size at most $\left \lfloor \frac{n}{2} \right \rfloor$, yielding a valid set of nodes to select.
+Another interpretation is that the BFS traversal constructs a spanning tree, which we do a bipartite coloring on. Because doing a bipartite coloring on the tree satisfies the requirement, we can ignore all edges that aren't part of our spanning tree.
 
 ## Implementation
 
--- a/solutions/silver/cf-1176E.mdx
+++ b/solutions/silver/cf-1176E.mdx
@@ -153,7 +153,7 @@ import sys
 
 input = sys.stdin.readline
 
-for _ in range(int(input()):
+for _ in range(int(input())):
 	n, m = map(int, input().split())
 	adj = [[] for _ in range(n)]
 	for _ in range(m):
--- a/solutions/silver/cf-1176E.mdx
+++ b/solutions/silver/cf-1176E.mdx
@@ -10,7 +10,11 @@ author: David Zhou
 
 ## Explanation
 
-To solve this problem, we perform BFS starting from any node and calculate the distance from the root to each node. Every node will have either a parity of even or odd. Since each even node is connected to at least one odd node and vice-versa, we can separate them into two sections. The categories have no overlap, so one group must have less than or equal to $n/2$ nodes. We can output the smaller of the two sets to fulfill the problem's requirements.
+To solve this problem, we perform [BFS](https://usaco.guide/silver/graph-traversal?lang=cpp) starting from any node and calculate the distance from the root to each node. Every node will have a parity of even or odd. 
+
+Since each even node is connected to at least one odd node and vice-versa, we can separate them into two sections. The categories have no overlap, so one group must have less than or equal to $n/2$ nodes. 
+
+We can output the smaller of the two sets to fulfill the problem's requirements.
 
 ## Implementation
 **Time Complexity:** $\mathcal{O}(M+N)$
--- a/solutions/silver/cf-1176E.mdx
+++ b/solutions/silver/cf-1176E.mdx
@@ -15,7 +15,7 @@ Perform [BFS](https://usaco.guide/silver/graph-traversal?lang=cpp) starting from
 
 Since BFS visits every node at the same level before moving on, edges must connect nodes of opposite parities. This provides the bipartition that separates every node.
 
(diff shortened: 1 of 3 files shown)


solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+## Explanation
+
+To solve this problem, we need to find a bipartite coloring where every node is either selected or adjacent to one that is.
+
+Perform [BFS](https://usaco.guide/silver/graph-traversal?lang=cpp) starting from any node and calculate the distance (level) from the root to each node. Every node will have a parity of even or odd.
+
+Since BFS visits every node at the same level before moving on, edges must connect nodes of opposite parities. This provides the bipartition that separates every node.
+
+The categories are disjoint and exhaustive, so one group must have $\leq n/2$ nodes. We can output the smaller of the two sets to satisfy the problem's requirements.
```
eysbutno:
```suggestion
The categories are disjoint and exhaustive, so one group must have less than or equal to $\left \lfloor \frac{n}{2} \right \rfloor$ nodes. We can output the smaller of the two sets to satisfy the problem's requirements.
```

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+author: David Zhou
+---
+
+[Official Analysis (C++)](https://codeforces.com/blog/entry/67598)
+
+## Explanation
+
+To solve this problem, we need to find a bipartite coloring where every node is either selected or adjacent to one that is.
+
+Perform [BFS](https://usaco.guide/silver/graph-traversal?lang=cpp) starting from any node and calculate the distance (level) from the root to each node. Every node will have a parity of even or odd.
```
eysbutno:
```suggestion
Perform [BFS](/silver/graph-traversal) starting from any node and calculate the distance (level) from the root to each node. Every node will have a parity of even or odd.
```

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Since BFS visits every node at the same level before moving on, edges must connect nodes of opposite parities. This provides the bipartition that separates every node.
```
eysbutno, line 16: idk if it would be good to frame it this way, but at least to me the most intuitive way to understand this problem is that we just need to do a bipartite coloring of a spanning tree, and print the smaller set of nodes

(which obv works by the definition of the problem, since adding non-tree edges can only 'improve' our state)

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int main() {
+	int t;
+	cin >> t;
+	while (t--) {
```
eysbutno:
```suggestion
	int test_num;
	cin >> test_num;
	for (int t = 0; t < test_num; t++) {
```

just a bit of a thing we do

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+					dist[next] = dist[curr] + 1;
+					q.push(next);
+				}
+			}
+		}
+
+		// separates nodes by parity and outputs smaller group
+		vector<int> even, odd;
+		for (int i = 0; i < n; i++) {
+			dist[i] % 2 == 0 ? even.push_back(i + 1) : odd.push_back(i + 1);
```
eysbutno: if you write it with a ternary i would write it like

`(dist[i] % 2 == 0 ? even : odd).push_back(i + 1);`

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+import java.io.*;
+import java.util.*;
+
+public class Cover {
+	public static void main(String[] args) throws IOException {
+		BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+		PrintWriter pw = new PrintWriter(new OutputStreamWriter(System.out));
+
+		int t = Integer.parseInt(br.readLine());
+		while (t-- > 0) {
```
eysbutno: same thing with test_num instead of t with cpp

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,193 @@
...
+			String[] nm = br.readLine().split(" ");
+			int n = Integer.parseInt(nm[0]);
+			int m = Integer.parseInt(nm[1]);
+			List<List<Integer>> adj = new ArrayList<>();
```
eysbutno: `List<Integer>[]` is better imo

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+To solve this problem, we need to find a bipartite coloring where every node is either selected or adjacent to one that is.
+
+Perform [BFS](/silver/graph-traversal) starting from any node and calculate the distance (level) from the root to each node. Every node will have a parity of even or odd.
+
+Since BFS visits every node at the same level before moving on, edges must connect nodes of opposite parities. This provides the bipartition that separates every node.
+
+The categories are disjoint and exhaustive, so one group must have less than or equal to $\left \lfloor \frac{n}{2} \right \rfloor$ nodes. We can output the smaller of the two sets to satisfy the problem's requirements.
+
+Another interpretation is that the BFS traversal constructs a bipartite spanning tree, whose level-based partition guarantees that one side has size at most $\left \lfloor \frac{n}{2} \right \rfloor$, yielding a valid set of nodes to select.
```
eysbutno:
```suggestion
Another interpretation is that the BFS traversal constructs a spanning tree, which we do a bipartite coloring on. Because doing a bipartite coloring on the tree satisfies the requirement, we can ignore all edges that aren't part of our spanning tree.
```

pleth9:
changes are all AC

solutions/silver/cf-1176E.mdx
```diff
@@ -0,0 +1,192 @@
...
+```
+
+</JavaSection>
+<PySection>
+
+```py
+from collections import deque
+import sys
+
+input = sys.stdin.readline
```
SansPapyrus683: does using raw `input()` work

pleth9: raw input() TLEs for normal python but works in pypy

Issue #5300, opened 2025-04-20
Contact Form Submission - Suggestion (Solution: Kilonova - Sum of divisors)


maggieliu05:
Someone submitted the contact form!

**URL**: https://usaco.guide/problems/kilonova-sumdiv/solution
**Module**: Solution: Kilonova - Sum of divisors
**Topic**: Suggestion
**Message**: 
It would be good to give a solution that doesn't require modular inverse and works for arbitrary moduli.

bqi343:
Example (without template):

```cpp
pair<mi, mi> operator*(pair<mi, mi> l, pair<mi, mi> r) {
	return {l.f * r.f, l.s + l.f * r.s};
}

pair<mi, mi> geo_seq(mi p, ll e) {
	// returns: {p^e, sum_{i=0}^{e-1}p^i}
	if (e == 0) return {1, 0};
	if (e & 1) return geo_seq(p, e - 1) * pair<mi, mi>{p, 1};
	auto half = geo_seq(p, e / 2);
	return half * half;
}

int main() {
	setIO("sumdiv");
	def(ll, A, B);
	V<pair<ll, int>> v = factor(A);
	mi ans = 1;
	for (auto [p, e] : v) {
		auto [pe, sum_pe] = geo_seq(p, e);
		auto [_, sum_ofac] = geo_seq(pe, B);
		ans *= p * sum_pe * sum_ofac + 1;
	}
	ps(ans);
}
```

lamaqdahodwala:
I pasted the alternate implementation alongside the pre-existing solution in PR #5431

PR #5431, opened 2025-06-30
Add alternative implementation to Sum of Divisors problem

lamaqdahodwala:
Addresses #5300 

- [X] I have tested my code.

### DIFF
--- a/solutions/gold/sumdiv.mdx
+++ b/solutions/gold/sumdiv.mdx
@@ -138,3 +138,39 @@ int main() {
 
 </CPPSection>
 </LanguageSection>
+
+### Alternative Implementation
+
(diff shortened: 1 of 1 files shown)


bqi343:
Please add a solution that compiles on its own and doesn't use macros.

lamaqdahodwala:
I pasted in this solution from the comment on #5300. I don't know what the code is missing in order for it to compile. My best guess was that there was a template that I hadn't pasted in which caused the compilation to fail (since it was mentioned that the code snippet excluded a template), but the template code from the pre-existing solution didn't fix it.

bqi343:
Please fill in the missing parts, or write your own solution along the same lines.

Issue #5455, opened 2025-07-08
Contact Form Submission - Suggestion


maggieliu05:
Someone submitted the contact form!

**URL**: https://usaco.guide/problems/
**Module**: None
**Topic**: Suggestion
**Message**: 
it would be great if in the [Problems tab](https://usaco.guide/problems/) there was a button to select a random problem based on the filters youve selected

also an option to hide the tags in the left hand side would be great

bqi343:
> it would be great if in the [Problems tab](https://usaco.guide/problems/) there was a button to select a random problem based on the filters youve selected

Would returning the problems in random order suffice?

note: this is already mentioned in #86

> also an option to hide the tags in the left hand side would be great

probably the existing "Hide tags" button should additionally hide the unselected tags on the left hand side. cc @emhgit in case you'd be working on this

StathisKons:
> Would returning the problems in random order suffice?

yes, it would suffice for the reasons I made the suggestion, right now the order in which the problems appear ~~is~~ feels kind of predictable, which isnt ideal for the kind of practice I wanted

I had in mind [this blog](https://codeforces.com/blog/entry/98806) when I made the suggestion, I like to use usaco guide both for learning and practicing a specific topic, but also as an archive with a pool of random educational problems to solve

emhgit:
> > it would be great if in the [Problems tab](https://usaco.guide/problems/) there was a button to select a random problem based on the filters youve selected
> 
> Would returning the problems in random order suffice?
> 
> note: this is already mentioned in [#86](https://github.com/cpinitiative/usaco-guide/issues/86)
> 
> > also an option to hide the tags in the left hand side would be great
> 
> probably the existing "Hide tags" button should additionally hide the unselected tags on the left hand side. cc [@emhgit](https://github.com/emhgit) in case you'd be working on this

that’s a good idea. i’m currently away from my pc, but in about a day i’ll implement that

ghost:
[@emhgit](https://github.com/emhgit)
Please do consider level or section as filter option.(Multiselect)
Level: Easy Medium Hard Very Hard Insane
Setion: Bronze Silver Gold Platinium Advance
Or even better option to exclude a particular topic(s).

bqi343:
@cjspd-oly multiselect by level and section is already supported.

bqi343:
I think this is done (correct me if I'm wrong)

Issue #86, opened 2020-07-22
Searching Problems


bqi343:
## Incomplete Features

Note that this list is not necessarily ordered by importance.

- [ ] Dark Mode #4152 
- [ ] better mobile support?
- [ ] search by + display division (Bronze - Plat) of the module that contains the problem
- [ ] option to select a random (unsolved) problem from a division
- [ ] save searches + create problemsets
- [ ] add link to user solutions? ex. https://usaco.guide/problems/ys-montmort/user-solutions
- [ ] optimize algolia operations usage? https://github.com/cpinitiative/usaco-guide/issues/86#issuecomment-1873341366
- [ ] properly handle "search by status"? ie. separate search by status and search by name into two separate searching systems if we exceed 1000 problems / Algolia's limit on refinements: https://github.com/cpinitiative/usaco-guide/pull/4157#discussion_r1439093684

---

from https://github.com/cpinitiative/usaco-guide/issues/86#issuecomment-788536196 (note that this list may be a bit outdated):

okay so here was what I am thinking (am open to changes):
- tags should look a lot nicer. Perhaps we can do something like "category" on the e-commerce demo, or add a searchable list of tags, idk. in general tags UI needs to be improved
- there should be an option to increase the number of hits per page, as well as pagination
- difficulty / module division should be displayed
- should have option to filter by difficulty / module(s)
- should have option to show starred only
- show whether user solved the problem yet

we should also allow users to use problems page to create randomized problemsets to do. this would entail:
- all of the filtering functionality above
- have an option to randomize the problems & hide difficulty / tags / modules
- (far future) save searches / problemsets

---

## Completed Features

- [x] search by + display tags
- [x] search by + display problem name
- [x] search by + display starred
- [x] search by + display source
- [x] search by + display module
- [x] show whether user solved the problem yet
- [x] search by + display difficulty (relative to module) #3320
- [x] option to increase the number of hits per page, as well as pagination #3394
- [x] option to hide difficulty / tags / modules #3394
- [x] edit problem status from search page

bqi343:
Request for searching by source (ex. USACO Plat) + difficulty.

thecodingwizard:
Difficulty is questionable... (Wormhole sort is both easy and insane XD)

bqi343:
> Difficulty is questionable... (Wormhole sort is both easy and insane XD)

Taking min difficulty could be okay. Also,

> hide solution links on the problems page if the user has tags + solutions off

[...]

PR #5537, opened 2025-08-14
feat: add shuffle and random buttons to problem search

vracton:
Added shuffle and random buttons to problem search. Addresses part of #86 and #5455.

- [x] I have tested my code.

https://github.com/user-attachments/assets/5fa05c97-6fa0-4d24-b51a-faf29f72b7f6

### DIFF
--- a/src/components/ProblemsPage/ProblemHits.tsx
+++ b/src/components/ProblemsPage/ProblemHits.tsx
@@ -10,18 +10,18 @@ import {
   useHideModulesSetting,
   useShowTagsSetting,
 } from '../../context/UserDataContext/properties/simpleProperties';
+import { useUserProgressOnProblems } from '../../context/UserDataContext/properties/userProgress';
 import {
   AlgoliaProblemInfo,
   getProblemURL,
   isUsaco,
   ProblemInfo,
-  recentUsaco,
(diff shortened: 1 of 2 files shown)


SansPapyrus683:
could you add a screen recording of what this looks like

Sosuke23:
the buttons are working as expected... but i am bit skeptical about their placement.

vracton:
> the buttons are working as expected... but i am bit skeptical about their placement.

What do you think would be better?

Sosuke23:
> > the buttons are working as expected... but i am bit skeptical about their placement.
> 
> What do you think would be better?

instead of centre, to the right maybe

bqi343:
@StathisKons as the person who submitted #5455, what do you think?

StathisKons:
> @StathisKons as the person who submitted #5455, what do you think?

The button functionality looks awesome, only concern is that it seems like after pressing the `random` button the user will have to manually find the problem again to mark the problem (as `solving`/`solved` etc). 
other than that LGTM!

bqi343:
The random button seems sorta unnecessary? In the sense that you can just shuffle and choose the first problem. ig it's fine to keep though

PR #5586, opened 2025-08-27
FocusProblem.tsx Tailwind fix (one-char)

### LINKED ISSUE #5584: Focus Problem component styling issue - probably a typo in the Tailwind classes
When hovering over a Focus Problem, the bottom margin vanishes, this making it stick with the second Focus Problem. See it in action in [Paths on Grids](https://usaco.guide/gold/paths-grids)

From inspecting with DevTools and checking the source code, I have concluded that it tries to add the classes `-translate-y-1 transform shadow-lg`, but it fails to add the `-translate-y-1` class because of a concatenation with the `mb-4` class, thus making a `mb-4-translate-y-1` class, which does not exist.

https://github.com/cpinitiative/usaco-guide/blob/83af56f46d00f77299c16461ec36d888d6467ccb/src/components/markdown/FocusProblem.tsx#L34-L36

The solution is to add a space at the beginning of the string right after the `?` on Line 35. As this change is changing a single character, it's probably better to change it within another commit instead of making a pull request.

EntityPlantt:

- [x] I have tested my code.

Closes #5584

### DIFF
--- a/src/components/markdown/FocusProblem.tsx
+++ b/src/components/markdown/FocusProblem.tsx
@@ -32,7 +32,7 @@ export default function FocusProblem({
   return (
     <div
       className={`block shadow transition dark:bg-gray-900 mb-4${
-        isHovered ? ' -translate-y-1 transform shadow-lg' : ''
+        isHovered ? '-translate-y-1 transform shadow-lg' : ''
       }`}
       id={'problem-' + problem.uniqueId}
     >
--- a/src/components/markdown/FocusProblem.tsx
+++ b/src/components/markdown/FocusProblem.tsx
@@ -32,7 +32,7 @@ export default function FocusProblem({
   return (
     <div
       className={`block shadow transition dark:bg-gray-900 mb-4${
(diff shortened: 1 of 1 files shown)


bqi343:
huh it seems that the prettier formatter undid the change ... seems like an issue with the formatter? 🤔

bqi343 (changes requested):
@thecodingwizard can you investigate?

maybe related to https://prettier.io/docs/options.html#html-whitespace-sensitivity ?

EDIT: I don't think `--html-whitespace-sensitivity strict` helps here ...

ghost:
i got the issue fixed.
making a commit to this pr.

ghost:
i don't know how to link commit with this pr.

ghost:
We need to add a space before the interpolation. Change `mb-4${` to `mb-4 ${` — otherwise it causes a syntax issue in the generated class string.

bqi343:
Resolved by #5592

PR #5599, opened 2025-08-30
Wizard's Tour Java Solution and Formatting

guodavid0418:
- added Java sol to internal pages (same as C++ for both Tree and General case)
- also fixed formatting issues (originally showed C++ code even if you selected Java/Python as lang)
- note: the Tree code doesn't actually AC (neither does the C++) since its a case, so it should just be equivalent to the c++ code.

- [x] I have tested my code.

General Case code AC:
Screenshot 2025-08-29 at 11 04 06 PM

### DIFF
--- a/solutions/silver/cf-860D.mdx
+++ b/solutions/silver/cf-860D.mdx
@@ -2,10 +2,12 @@
 id: cf-860D
 source: CF
 title: Wizard's Tour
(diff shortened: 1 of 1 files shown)


eysbutno (approved):
yet another guorz pr

PR #5617, opened 2025-09-06
diluc and kaeya solution (1536C)

Streakwind:

- [x] I have tested my code.

### DIFF
--- a/solutions/gold/cf-1536C.mdx
+++ b/solutions/gold/cf-1536C.mdx
@@ -8,6 +8,7 @@ author: Kevin Sheng, Ruben Jing
 [Official Editorial (C++, Java)](https://codeforces.com/blog/entry/91520)
 
 ## Explanation
+
 Lets consider a string with a non-reduced ratio $2D:2K$ and a substring with
 ratio $D:K$. If we take out the substring with ratio $D:K$, we are left
 with another substring with ratio $D:K$ ($2D-D:2K-K$), therefore making


solutions/gold/cf-1536C.mdx
```diff
@@ -2,13 +2,38 @@
...
 title: Diluc and Kaeya
-author: Kevin Sheng
+author: Kevin Sheng, Ruben Jing
 ---
 
 [Official Editorial (C++, Java)](https://codeforces.com/blog/entry/91520)
 
+## Explanation
+The problem is essentially asking for the maximum amount of substrings you can
+split a string into such that the [reduced] ratio of  $D:K$ in the original
```
SansPapyrus683: why brackets

streakwind: good point

solutions/gold/cf-1536C.mdx
```diff
@@ -2,13 +2,38 @@
...
-author: Kevin Sheng
+author: Kevin Sheng, Ruben Jing
 ---
 
 [Official Editorial (C++, Java)](https://codeforces.com/blog/entry/91520)
 
+## Explanation
+The problem is essentially asking for the maximum amount of substrings you can
+split a string into such that the reduced ratio of $D:K$ in the original
+string is the same as the ratio of $D:K$  in all of its substrings.
```
SansPapyrus683: is there a need to restate the problem like this

streakwind: I couldn't understand the problem on the first few reads so for low intellect people like me probably

SansPapyrus683: ok but if they don't get the problem usually the first instinct isn't to go to the internal solution
at least i think that's the case(?)

streakwind: lmao that is true . maybe I can keep it as a hint?

streakwind: though the format is still being decided on iirc

SansPapyrus683: yeah sure it can be a hint, though we should prolly get a 2nd opinion

streakwind: can you find someone else to review I can’t request again on github

eysbutno: to me the problem was pretty clear on first read, so i don't think it's particularly necessary

however, i don't think it's always a bad thing to briefly summarize what the problem wants

solutions/gold/cf-1536C.mdx
```diff
@@ -2,13 +2,38 @@
...
+Keep track of the total number of 'D's and total number of 'K's. For each prefix,
+the ratio of $D:K$ in lowest terms is $\frac{D}{\gcd(D,K)}:\frac{K}{\gcd(D,K)}$.
+The solution to each prefix is how many times this reduced ratio has appeared so
+far plus one.  We add one because the amount of times this reduced ratio appeared
+is the amount of splits, and the amount of pieces resulting from the splits is the
+splits plus one.
+
 ## Implementation
 
+**Time Complexity:** $\mathcal{O}(NM)$
```
eysbutno: wait what's M here

this should be $\mathcal{O}(N \log N)$ no?

solutions/gold/cf-1536C.mdx
```diff
@@ -2,13 +2,38 @@
...
 ---
 
 [Official Editorial (C++, Java)](https://codeforces.com/blog/entry/91520)
 
+## Explanation
+The problem is essentially asking for the maximum amount of substrings you can
+split a string into such that the reduced ratio of $D:K$ in the original
+string is the same as the ratio of $D:K$  in all of its substrings.
+
+In order for a string to be split once or more, there must be a substring
```
eysbutno: i think this is a little redundant (the first half of this paragraph), since the fact at the end of this paragraph is what people would probably need

eysbutno (commented):
overall explanation is decent, I think you should perhaps put a bit more focus in explaining why the fraction stuff is correct (i.e. just note that to 'cut' a prefix of the string out of the current string, its ratio must be correct, and that we can always cut a prefix of the string out if the ratio is correct)

solutions/gold/cf-1536C.mdx
streakwind: test cases * string processing

eysbutno: yeah but sum of n is bounded, so this is a bit misleading

streakwind: I see, thanks. should it be $\mathcal{O}{(N)}$ though bc the implementation below uses an unordered map so access is $\mathcal{O}{(1)}$?

eysbutno: hmm, that's a good question

i think writing O(N) is fine, although it can probably be hacked 😭

SansPapyrus683:
oh i wrote the code for this LMFAAAOOO

SansPapyrus683 (approved):

eysbutno:
> oh i wrote the code for this LMFAAAOOO

`std::unordered_map<int, std::unordered_map<int, int>>` is diabolical 😭

PR #5688, opened 2025-10-01
Add language dropdown to internal user solutions page

MrinallU:

- [ ] I have tested my code.

### DIFF
--- a/src/components/ProblemSolutions.tsx
+++ b/src/components/ProblemSolutions.tsx
@@ -20,6 +20,17 @@ import useUserProblemSolutionActions from '../hooks/useUserProblemSolutionAction
 import useUserSolutionsForProblem from '../hooks/useUserSolutionsForProblem';
 import { ShortProblemInfo } from '../models/problem';
 import CodeBlock from './markdown/CodeBlock/CodeBlock';
+import {
+  Menu,
+  MenuButton,
+  MenuItem,
+  MenuItems,
+  Transition,
+} from '@headlessui/react';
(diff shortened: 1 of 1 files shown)


bqi343:
Can you take a screenshot of what changed?

Sosuke23:
i don't think this is necessary at all, and even if it is, it is not working as intended. 

If you select python in the dropdown, then why does it show other language solutions after python solutions? Shouldn't it just hide them at that moment, logically?

MrinallU:
> Can you take a screenshot of what changed?
Sure!
image

MrinallU:
> i don't think this is necessary at all, and even if it is, it is not working as intended.
> 
> If you select python in the dropdown, then why does it show other language solutions after python solutions? Shouldn't it just hide them at that moment, logically?

Just thought it would be somewhat useful to have since users don't have to go back a page just to change language.

bqi343 (approved):

Sosuke23:
> > i don't think this is necessary at all, and even if it is, it is not working as intended.
> > If you select python in the dropdown, then why does it show other language solutions after python solutions? Shouldn't it just hide them at that moment, logically?
> 
> Just thought it would be somewhat useful to have since users don't have to go back a page just to change language.

in user solutions you don't have to change language, every language solutions appears at once. 

Plus you haven't answer my second point

bqi343:
I agree with @MrinallU - since the user's preferred language affects the ordering of user solutions, including the language dropdown on that page is useful.

Issue #5686, opened 2025-10-01
Fix Editor Add Problem by URL


bqi343:
see https://github.com/cpinitiative/usaco-guide/issues/5682#issuecomment-3354402609

ref: #4251

ghost:
# FAILED FOR USACO AND CODEFORCES
## SyntaxError: Unexpected token 'A', "A server e"... is not valid JSON

# WORKS ONLY FOR CSES AND ATCODER
<img width="731" height="525" alt="Image" src="https://github.com/user-attachments/assets/2188a334-0b06-4638-9461-8fb296f95afb" />

<img width="733" height="566" alt="Image" src="https://github.com/user-attachments/assets/c4d56c4a-923f-46b9-ad11-b98913e6af1e" />

Sosuke23:
The current parser is correct, no doubt. see the example

<img width="1920" height="1017" alt="Image" src="https://github.com/user-attachments/assets/897fbe22-cc86-4f8b-b506-826e189832c2" />

ghost:
@Sosuke23 

<img width="940" height="450" alt="Image" src="https://github.com/user-attachments/assets/0efac9a2-1b22-44bd-bb47-a42c1559d3a3" />

<img width="739" height="330" alt="Image" src="https://github.com/user-attachments/assets/bef5eb06-759b-4b4f-8694-9f18f47ce1be" />

<img width="977" height="417" alt="Image" src="https://github.com/user-attachments/assets/06652c3d-fb97-4d99-bdea-0452adc6e8e2" />

Sosuke23:
i am saying that the issue maybe something else

ghost:
> i am saying that the issue maybe something else

Thank you so much.

Sosuke23:
plus it is giving 500 internal server error when you debug in network section,

emhgit:
i investigated the issue, here are the results:

1. using codeforces URLs causes errors because invalid JSON data is return. for example, the url `https://codeforces.com/problemset/problem/1660/B` this error:

`SyntaxError: Unexpected token 'E', "Error when"... is not valid JSON`

2.  i checked the backend error logs from my terminal to get more info:
error message shown in terminal: 

```bash
https://codeforces.com/problemset/problem/1660/B url

 ERROR 

Request failed with status code 403



  Error: Request failed with status code 403

  - createError.js?:16 createError
    [.]/[axios]/lib/core/createError.js?:16:15

  - settle.js?:17 settle
[...]

PR #5706, opened 2025-10-12
URL Parser Error Fixed

emhgit:

- [x] I have tested my code.

Relevant issue: https://github.com/cpinitiative/usaco-guide/issues/5686

The error in issue 5686 was occurring because the previous code was trying to directly scrape HTML from the problem webpage. CodeForces uses a cloudflare challenege page to block potential bot requests, which caused axios to return a 403 Forbidden error. 

I solved this by creating a function called `tryCodeforcesAPI` that directly requests data from the CF API. This function handles CF URLs in `https://codeforces.com/problemset/problem/{contestId}/{index}` and the `https://codeforces.com/contest/{contestId}/problem/{index}` formats. I also added a fallback that tries to directly scrape the CF problem webpage with browser headers (although this fallback is iffy). 

These changes result in proper JSON being returned, as shown in the example:

image

I also included proper browser headers for all other requests, which should hopefully bypass anti-bot measures of usaco.org and other websites.

**NOTE:**

When I start the development server on my localhost and use usaco.org problem URLs, the requests are successful. However, when the app is deployed on Vercel, the requests return a 403 Forbidden error. Even with the realistic browser headers I added, the usaco.org website is flagging the requests as bot traffic; this might be because of the cloud IP address Vercel uses, which would explain why it worked on my local residential IP address. Maybe a forward proxy would be used, but I would need to find a reliable and free one.

### DIFF
--- a/src/api/(parsers)/parse.ts
+++ b/src/api/(parsers)/parse.ts
@@ -26,10 +26,7 @@ const browserHeaders = {
   'Cache-Control': 'max-age=0',
 };
 
-async function fetchWithRetry(
-  url: string,
-  maxRetries: number = 3
-): Promise<string> {
+async function fetchWithRetry(url: string, maxRetries = 3): Promise<string> {
   for (let attempt = 0; attempt < maxRetries; attempt++) {
     try {
       const headers = browserHeaders;
@@ -92,8 +89,6 @@ async function tryCodeforcesAPI(url: string): Promise<string | null> {
     // Extract contest ID and problem index from URL
     // First try: https://codeforces.com/problemset/problem/{contestId}/{index}
     let match = url.match(/\/problemset\/problem\/(\d+)\/([A-Z])/);
-    let contestIdStr: string;
-    let problemIndex: string;
     let isProblemsetFormat = true;
 
     if (!match) {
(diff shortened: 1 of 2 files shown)


src/api/(parsers)/parse.ts
```diff
@@ -3,15 +3,204 @@ import parseAc from './ac';
...
+
+  throw new Error('All retry attempts failed');
+}
+
+// Fallback function to try Codeforces API if direct scraping fails
+async function tryCodeforcesAPI(url: string): Promise<string | null> {
+  try {
+    // Extract contest ID and problem index from URL
+    // URL format: https://codeforces.com/problemset/problem/{contestId}/{index}
+    const match = url.match(/\/problemset\/problem\/(\d+)\/([A-Z])/);
```
ryanchou-dev: hmm, most problems have two possible urls (problemset & contest)
https://codeforces.com/problemset/problem/2159/F
https://codeforces.com/contest/2159/problem/F

add an extra case to parse /contest/ urls? ^

emhgit: good catch, i will add this tomorrow

PR #5711, opened 2025-10-13
add codechef parser

Sosuke23:

- [ ] I have tested my code.


Sosuke23:
image

works fine now

src/api/(parsers)/cc.ts
```diff
@@ -0,0 +1,19 @@
+// example url: https://www.codechef.com/problems/GLADFIGHT
+export default function parseCC(url: string, html: string) {
+  const urlSplit = url.split('/');
+  const uniqueId = urlSplit.at(4);
```
ryanchou-dev: could you find the id by indexing from the back? will error if `https://` is omitted

Sosuke23: done and added a checker for the correct link too.

Sosuke23: it was not at all useful.... 
if we miss https:// it already give some kind of error -> `SyntaxError: Unexpected token 'A', "A server e"... is not valid JSON`

or if there is any issue with the link it gives the same error.

ryanchou-dev: received the same error, sorry about that 👍 
should be fine to leave as is though

PR #5734, opened 2025-10-25
Added Solution for Cf 755C

lems9520138:

- [x] I have tested my code.

### DIFF
--- a/solutions/silver/cf-755C.mdx
+++ b/solutions/silver/cf-755C.mdx
@@ -5,20 +5,19 @@ title: PolandBall & Forest
 author: Arnav Gokhale
 ---
 
-[Official Editorial](https://codeforces.com/blog/entry/146988)
+[Official Editorial](https://codeforces.com/blog/entry/49793)
 
 ## Explanation
 
 Let's focus on one tree and its diameter first, naming the two endpoints $E_1$ and $E_2$. For every vertex $v$ on this tree, $p[v]$ is either equal to $E_1$ or $E_2$, which we can show as proved [here](https://codeforces.com/blog/entry/101271).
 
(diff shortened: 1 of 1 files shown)


solutions/silver/cf-755C.mdx
```diff
@@ -0,0 +1,102 @@
+---
+id: cf-755C
+source: CF - 8VC Venture Cup 2017 - Elimination Round
+title: PolandBall & Forest
+author: Arnav Gokhale
+---
+
```
eysbutno, line 7: include official editorial :D

solutions/silver/cf-755C.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Explanation
+
+Let's focus on one tree and it's diameter first naming the two endpoints $E_1$ and $E_2$. For every vertex $v$ on this tree, $p[v]$ is either equal to $E_1$ or $E_2$.
```
eysbutno:
```suggestion
Let's focus on one tree and its diameter first, naming the two endpoints $E_1$ and $E_2$. For every vertex $v$ on this tree, $p[v]$ is either equal to $E_1$ or $E_2$.
```

also:
- do we want to address the case of multiple diameters existing? since it doesn't change the answer but could be a nuance worth talking about
- maybe link the explanation for the fact that furthest distance for any node is a diameter endpoint

solutions/silver/cf-755C.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+So, we can count the number of distinct vertices in $P$ and divide by two to find the number of trees.
+
+However, we must be careful in the case where $P[i]$ = $i$, where the tree is simply a isolated vertex.
```
eysbutno: maybe combine these two with the fact that any tree with size >1 will have both endpoints occur at least once and then mention the size=1 case, idk

PR #5736, opened 2025-10-28
Fix dead Kattis links

Matistjati:

- [X] I have tested my code.

In general, Kattis has moved away from being able to submit to finished contests. They are now available on open.kattis.com instead. When i grep:ed, I also found a link to https://cornell-hspc19.kattis.com/problems, which would then be replaced by https://open.kattis.com/problem-sources/Cornell%20University%20High%20School%20Programming%20Contest%202019. But it's commented out, so I decided to not change it.


Matistjati:
At least when grep:ing for "kattis\.com" in the root of the project, I don't find any more (except the Cornell contest, as mentioned). Would you expect any other format?

PR #5813, opened 2025-12-17
Best lineup silver

sachet-abey:

- [x] I have tested my code.

### DIFF
--- a/content/extraProblems.json
+++ b/content/extraProblems.json
@@ -122,7 +122,7 @@
       "source": "Silver",
       "difficulty": "Normal",
       "isStarred": false,
-      "tags": ["Connected Components"],
+      "tags": ["Connected Components", "Prefix Sums"],
       "solutionMetadata": {
         "kind": "USACO",
         "usacoId": "1040"


solutions/silver/usaco-1494.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+To understand this, consider the test case $[6, 4, 2, 5, 2, 3]$ where we consider moving 5 right after 6. The subarray $[4, 2]$ has maximum 4, and this is greater than the next element in $s$, 3, so this move is beneficial. However, if the 4 were a 1 or 2, it wouldn't have been.
+
+The only thing remaining is to find the optimal chosen element. To do this, we iterate $s$ from left to right, and whenever we find an element satisfying the condition, we choose this element to move and break. This is because future elements satisfying the condition aren't as optimal. Consider the array $[7, 5, 1, 6, 3, 4]$. We can either move 6 right after 7 or 4 right after 6, as they both satisfy the condition. However, the former leads to the answer $[7, 6, 5, 4]$ and the latter leads to the answer $[7, 6, 4, 3]$, which is lexicographically less.
```
eysbutno:
```suggestion
The only thing remaining is to find the optimal chosen element. To do this, we iterate $s$ from left to right, and whenever we find an element satisfying the condition, we choose this element to move and break. This is because future elements satisfying the condition aren't as optimal. Consider the array $[7, 5, 1, 6, 3, 4]$. We can either move $6$ right after $7$ or $4$ right after $6$, as they both satisfy the condition. However, the former leads to the answer $[7, 6, 5, 4]$ and the latter leads to the answer $[7, 6, 4, 3]$, which is lexicographically smaller.
```

in general put numbers in math mode

solutions/silver/usaco-1494.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+It can be shown that the optimal solution is achieved by performing the operation on one of the elements in $s$. Consider some chosen element in $s$ with index $x$ such that the previous and following elements have index $l$ and $r$, respectively. It is optimal to perform the operation on the index $x$ to move it right after $l$. This is because some element in the range $(l, x)$ can potentially be included in the lexicographically optimal sequence. The candidate for this is the maximum in this range, and this needs to be at least the value of $a_r$ to be included.
+
+To understand this, consider the test case $[6, 4, 2, 5, 2, 3]$ where we consider moving 5 right after 6. The subarray $[4, 2]$ has maximum 4, and this is greater than the next element in $s$, 3, so this move is beneficial. However, if the 4 were a 1 or 2, it wouldn't have been.
```
eysbutno:
```suggestion
To understand this, consider the test case $[6, 4, 2, 5, 2, 3]$, where we consider moving 5 right after 6. The subarray $[4, 2]$ has a maximum of $4$, which is greater than the next element in $s$, 3, so this move is beneficial. However, if the $4$ were a $1$ or $2$, moving it wouldn't have been beneficial.
```

PR #5837, opened 2025-12-27
cf 1291D explanation

Streakwind:

- [ ] I have tested my code.

### DIFF
--- a/solutions/platinum/baltic-10-candies.mdx
+++ b/solutions/platinum/baltic-10-candies.mdx
@@ -88,7 +88,7 @@ int main() {
 	}
 
 	cout << a[best.second] << ' ';
-	for (int i = 1; i < 70000; i++)
+	for (int i = 1; i < 700000; i++)
 		if (!dp[700000 + i]) return cout << i, 0;
 	return 0;
 }


solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+3. The string is of length 1
+
+The last two conditions are easy to check. However, to check the first condition, we
+must use prefix sums. For each of the 26 letters of the alphabet, we can hold a prefix
+sum counting how many times the letter has been seen. If on an interval, three letters
+have been seen at least once, then the string is an irreducible anagram.
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(n)$ per test case
```
pleth9: O(N) is the runtime complexity to process the length of the string.
O(1) is per query (or i gues "test case") as you just process your prefix sum

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+author: Ruben Jing
+---
+
+[Official Editorial (C++)](https://codeforces.com/blog/entry/73563)
+
+## Explanation
+
+Observe that a string is an irreducible anagram if any of these three conditions hold
+
+1. The string has more than three different characters
```
pleth9:
```suggestion
1. The string has three or more different characters
```

solutions/silver/cf-1291D.mdx
streakwind: you're right lemme update that

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
+---
+id: cf-1291D
+source: CF
+title: Irreducible Anagrams
+author: Ruben Jing
+---
+
+[Official Editorial (C++)](https://codeforces.com/blog/entry/73563)
+
+## Explanation
```
pleth9, line 10: add a little explanation on how we get to THREE as the number of characters needed in condition 1

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+Observe that a string is an irreducible anagram if any of these three conditions hold
```
eysbutno: Image

isn't this the condition for if a string has at least one irreducible anagram?

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+3. The string is of length 1
+
+The last two conditions are easy to check. However, to check the first condition, we
+must use prefix sums. For each of the 26 letters of the alphabet, we can hold a prefix
+sum counting how many times the letter has been seen. If on an interval, three letters
+have been seen at least once, then the string is an irreducible anagram.
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(N)$ preprocessing and $\mathcal{O}(1)$ per query
```
eysbutno: perhaps $\mathcal{O}(|s| + t)$ or something like that?

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+
+using namespace std;
+
+int main() {
+	string s;
+	int t;
```
eysbutno: move t down to after preprocessing

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+	cin >> s;
+	cin >> t;
+
+	vector<vector<int>> pref(26, vector<int>(s.length() + 1));
```
eysbutno: maybe make 26 a constant (e.g. `LETTERS` or smth?)

solutions/silver/cf-1291D.mdx
eysbutno: agree, for overall explanation maybe something like
- handle easy edge case, $l = r$
- for others, irreducible anagram $t$ must be where each proper prefix (i.e. prefix isn't the entire string) is not an anagram of $s[l:r]$
- best way to make it irreducible is to like, take "later" characters and shove them in the front, as that sort of forces everything to be wrong until the end (ik this isn't very formal but surely it maeks enough sense)
- then you just have to sort of mess around with certain cases, if $s_l \neq s_r$ then we can shove all occurrences of $s_r$ to the start and do whatever we want cuz the number of occurrences of $s_r$ is gonna be wrong in each prefix
  - from that we try the alternate case to figure out how to solve that, idk if there's a more intuitive construction for the $s_l = s_r$ case

u could (and prob should) put a spoiler for the proof on why the conditions are sufficient, tbh u can prob just yoink it from the edi

eysbutno: editorial's construction is to take last character distinct from all prev characters, spam that, spam $s_n$, and do anything with rest, which makes sense but isn't that intuitive to me at least

maybe like
- would be coolio if we could just shove all occurrences of $s_n$ at start, but something like `aaabca` exists and there's more like it
- if we padded the occurrences with some extra characters at the front it would probably work, if number of distinct = 2 then this doesn't work tho, but if we have an extra third distinct char it's coolio
- more distinct = more good trust

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+The case where three distinct characters exist is a bit more complicated. In order to assure that the first character of $s$ and
+an anagram are different, we can insert all occurrences of the last distinct character to appear to the front. Next, to assure that
+the last character of $s$ and an anagram are different, we can insert all occurrences of the last character after the last distinct
+character, if they are different (if they are the same, we don't need to do anything). Now, we can insert the rest of the string
+in any order. This anagram is irreducible. The last character in $s$ will be one of the first characters in $t$, forcing us to
+use the entire string if we wish to create an anagram.
+
+We can generalize these observations: a string has at least one irreducible
+anagram if any of these three conditions hold
```
eysbutno:
```suggestion
anagram if any of these three conditions hold:
```

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,134 @@
...
+using namespace std;
+
+const int LETTERS = 26;
+
+int main() {
+	string s;
+
+	cin >> s;
+
+	vector<vector<int>> pref(LETTERS, vector<int>(s.length() + 1));
```
eysbutno: maybe `std::vector<std::array<int, 26>> pref(s.length() + 1)`? IMO std::array is more appropriate here, and also cuz of the ordering of your loops

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+	for (int i = 1; i < s.length() + 1; i++) {
+		char let = s[i - 1];
+
+		for (int letter = 0; letter < LETTERS; letter++) {
```
eysbutno: pref[i] = pref[i - 1] works if you lay it out with std::array

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+		int num_diff = 0;
+
+		for (int i = 0; i < LETTERS; i++) {
+			if (pref[i][b] - pref[i][a - 1] > 0) num_diff++;
+		}
+
+		string ans = "";
+
+		if (b == a) {
```
eysbutno: i think it's easier to just combine the 3 conditions into one if statement and directly `std::cout << "Yes" << '\n';` or print no

solutions/silver/cf-1291D.mdx
```diff
@@ -0,0 +1,134 @@
...
+anagram because its only anagram is itself. Because its only anagram is itself, any segments that we choose
+will always be anagrams of each other, meaning it only has a reducible anagram.
+
+From the previous examples, it can be seen that having $t[l]=s[l]$ and/or $t[r]=s[r]$ does not help with
+creating an irreducible anagram. If this happens, we can split the string into two segments, one of
+length one at either $l$ or $r$, and another of everything else. For example, $s=\text{abcca}$,
+$\text{aabcc}$ is not an irreducible anagram because we can have a segment $[0, 1]$ and another $[2, 4]$.
+Thus it is optimal to have $t[l]\neq s[l]$ AND $t[r]\neq s[r]$.
+
+However, notice in the case of two distinct characters, an irreducible anagram cannot be formed. Even with the ends of
```
eysbutno: i think you can explain in a more general way

don't have the expl off the top of my head but i think official edi proves it

PR #5858, opened 2026-01-02
usaco 596

Streakwind:

- [ ] I have tested my code.

### DIFF
--- a/content/4_Gold/LIS.problems.json
+++ b/content/4_Gold/LIS.problems.json
@@ -70,15 +70,14 @@
     },
     {
       "uniqueId": "usaco-496",
-      "name": "Cowjog",
+      "name": "Cow Jog",
       "url": "http://www.usaco.org/index.php?page=viewproblem2&cpid=496",
       "source": "Old Gold",
       "difficulty": "Normal",
       "isStarred": false,
       "tags": ["DP"],
       "solutionMetadata": {
-        "kind": "USACO",
-        "usacoId": "496"
+        "kind": "internal"
       }
     },
     {
--- a/content/extraProblems.json
+++ b/content/extraProblems.json
@@ -1982,7 +1982,7 @@
       "isStarred": false,
       "tags": [],
       "solutionMetadata": {
-        "kind": "USACO",
+        "kind": "internal",
         "usacoId": "1397"
       }
     },
--- a/solutions/bronze/usaco-1397.mdx
+++ b/solutions/bronze/usaco-1397.mdx
@@ -1,0 +1,114 @@
+---
+id: usaco-1397
+source: USACO Bronze 2024 February
+title: Maximizing Productivity
+author: SeungJoon Park
+---
+
+[Official Analysis (C++, Python)](https://usaco.org/current/data/sol_prob3_bronze_feb24.html)
+
+## Explanation
+
+First, we rearrange $S+t_i \lt c_i$ to $S \lt c_i - t_i$, as sorting farms in descending order of $c_i - t_i$ makes comparisons easier.
+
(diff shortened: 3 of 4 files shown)


solutions/silver/usaco-596.mdx
```diff
@@ -2,26 +2,46 @@
...
 ---
 
 [Official Analysis (Java)](http://www.usaco.org/current/data/sol_gates_silver_jan16.html)
 
+## Explanation
+
+We can model the fences, and then floodfill to find the number of connected components. The number
+of regions closed by a fence is the number of connected components minus one, as one connected component
+is the rest of the farm that isn't enclosed by a fence. Each region only needs one gate to connect it
+to the rest of the farm, so the answer is the amount of connected components minus one.
```
eysbutno: maybe like
"We can model the fences, and floodfill to find the number of connected components, including the region not enclosed by a fence. Each gate connects two regions, so merging regions enclosed by a fence with the outside results in the number of connected components minus one gates being used." or something?

solutions/silver/usaco-596.mdx
```diff
@@ -2,26 +2,46 @@
...
+
+If a fence is seen, the floodfil is stopped.
```
eysbutno: how does this tie into the solution? (also floodfill is mispelled)

maybe categorize the rest of the explanation as figuring out how to find the number of CCs

solutions/silver/usaco-596.mdx
```diff
@@ -2,26 +2,46 @@
...
+
+To not go out of bounds, an array of size $4003 \times 4003$ should be used:
+$1$ unit for the origin position, $2000$ units in each direction from
+the origin and $1$ more unit in each direction to ensure that the
+farm includes area beyond the fence. The origin is at $(2001, 2001)$, and there is
+"buffer" space at row/column $0$ and $4002$, which is obtained from adding or
+subtracting $2000$ from the origin then correspondingly adding or subtracting $1$.
+
+If we floodfill through the entire grid without constraints,
+a stack overflow would occur. This is why we must use `maxx, minx, maxy, miny`.
```
eysbutno:
```suggestion
a stack overflow would occur. This is why we must use `maxx`, `minx`, `maxy`, and `miny`.
```

also does using these variables actually prevent stack overflow ? if we consider this part of the solution code:

```
// expand the farm to include area outside the fences
minx--;
maxx++;
miny--;
maxy++;
```

this means that we would visit the entire grid anyways (if im reading this correctly)

solutions/silver/usaco-596.mdx
```diff
@@ -2,26 +2,46 @@
...
 
 </Warning>
 
 ## Implementation
 
+**Time Complexity:** $\mathcal{O}((maxX - minX)\times(maxY-minY))$
```
eysbutno: i think $\mathcal{O}(N^2)$ is more accurate because:
- the size of the grid needed is proportional to N (the dimensions)
- we basically visit the entire grid anyways

solutions/silver/usaco-596.mdx
streakwind: yeah, they do prevent a stack overflow. I implemented it with the entire 4003 x 4003 grid which is like quite a large call stack for floodfill and I got memory limit exceeded.

the only reason the grid is so large is in case farmer john goes in one direction one thousand times but then the area won't be so big so we shouldn't actually floodfill the entire grid

solutions/silver/usaco-596.mdx
streakwind: $N$ represents the number of inputs so it might be a bit misleading

streakwind: wait actually on second thought the grid is scaled to like $4N$ so I think $N^2$ would work you're right

PR #5872, opened 2026-01-07
Fix C++ Solution for Acowdemia I

AsynchronousAI:

- [x] I have tested my code.


AsynchronousAI:
Previously the solution would ignore the l variable and fail test cases 
Screenshot 2026-01-07 at 12 53 02 PM
I have fixed it, and also wrapped repeating code in a lambda function
Screenshot 2026-01-07 at 12 52 29 PM

AsynchronousAI:
> do the python and java code also fail? if so it'd be nice to also fix those

Python seems to work normally since the typo was `l` being replaced with `1`, and there isn't a Java solution currently.

PR #5943, opened 2026-02-09
Added links to existing solution websites (earlier mentioned kind:"none")

ashoknarayan:
All these 3 questions have solution in Codeforces.  If there needs to be message like "Refer CF" then I will change it to that. For now I have just added the links of the editorial pages in CF


bqi343:
Should look like

```
      "solutionMetadata": {
        "site": "CF"
      }
```

for most CF problems.

ashoknarayan:
I changed it to `autogen-label-from-site`

bqi343 (approved):

PR #5946, opened 2026-02-11
Add Sliding Window Summation

AVS-Memer:

- [x] I have tested my code.

### DIFF
--- a/solutions/silver/usaco-1544.mdx
+++ b/solutions/silver/usaco-1544.mdx
@@ -8,13 +8,14 @@ author: Aryaveer Shishodia
 [Official Analysis (Python)](http://usaco.org/current/data/sol_prob3_silver_season26contest1.html)
 
 ## Explanation
-The "remainder when the number of ones in the length-K window of $b$ with leftmost index $i$ is divided by two" is analogous to the XOR of the window. The only difference between the windows of $r_i$ and $r_{i+1}$ is that $r_i$'s window contains $b_i$, while $r_{i+1}$'s contains $b_{i+k}$.
+The value of $r_i$ is equivalent to the XOR of the length-K window beginning at $i$. The only difference between the windows of $r_i$ and $r_{i+1}$ is that $r_i$'s window contains $b_i$, while $r_{i+1}$'s contains $b_{i+k}$.
 
-$$r_i\oplus r_{i+1} = b_i\oplus b_{i+k}$$
+$$(b_i\oplus b_{i+1} ...\oplus b_{i+k-1})\oplus(b_{i+1}...\oplus b_{i+k-1}\oplus b_{i+k}) = r_i\oplus r_{i+1} = b_i\oplus b_{i+k}$$
+$$r_i\oplus r_{i+1}\oplus b_i = b_{i+k}$$
 
-Since we are given $r$, choosing the value of $b_i$ determines $b_{i+k}$, which in turn determines $b_{i+2k}$, $b_{i+3k}$, and so on. This splits $b$ into $k$ binary chains, each determined by one of the first $k$ characters of $b$. For each chain, we can set the first character to `0` and `1`, and compare which one will minimize the amount of `1` bits.
+Since we are given $r$, choosing the value of $b_i$ determines $b_{i+k}$, which in turn determines $b_{i+2k}$, $b_{i+3k}$, and so on. This splits $b$ into $k$ binary chains, each dependent on one of the first $k$ characters of $b$. For each chain, we can set the first character to `0` or `1`, comparing which will minimize the amount of `1` bits.
 
-The only thing that restricts the first $k$ bits of $b$ is $r_1$, which will sometimes force us to flip one of our initial bits despite it increasing the number of `1` bits in the chain. We can check if our configuration satisfies $r_1$ by recording the bitwise XOR of the optimal first $k$ bits, and storing a running minimum $m$ the minimal cost of flipping a chain. If the bitwise XOR of the first k elements doesn't align with $r_1$, we can then flip that chain back and add $m$ to our count.
+The only thing that restricts the first $k$ bits of $b$ is $r_1$, which will sometimes force us to flip one of our initial bits despite it increasing the number of `1` bits in the string. We can check if our configuration satisfies $r_1$ by recording the bitwise XOR of the optimal first $k$ bits, and storing a running minimum $m$, representing the additional amount `1` bits to flip a chain. If the bitwise XOR of the optimal first $k$ bits doesn't align with $r_1$, we can then flip the selected chain back and add $m$ to our count.
 
 We can then apply the same idea to maximize the amount of `1` bits.
 


solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+Since we are given $r$, choosing the value of $b_i$ determines $b_{i+k}$, which in turn determines $b_{i+2k}$, $b_{i+3k}$, and so on. This splits $b$ into $k$ binary chains, each determined by one of the first $k$ characters of $b$. For each chain, we can set the first character to `0` and `1`, and compare which one will minimize the amount of `1` bits.
+
+The only thing that restricts the first $k$ bits of $b$ is $r_1$, which will sometimes force us to flip one of our initial bits despite it increasing the number of `1` bits in the chain. We can check if our configuration satisfies $r_1$ by recording the bitwise XOR of the optimal first $k$ bits, and storing a running minimum $m$ the minimal cost of flipping a chain. If the bitwise XOR of the first k elements doesn't align with $r_1$, we can then flip that chain back and add $m$ to our count.
+
+We can then apply the same idea to maximize the amount of `1` bits.
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(\Sigma{N})$
```
eysbutno: usually we just write $\mathcal{O}(N)$ here, since multitest doesn't really change complexity

solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+<LanguageSection>
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+int main() {
+	int t, n, k;
+	string x, y;
```
eysbutno: put n, k, x, and y in the loop for each test case

also, perhaps name the variables a bit differently? e.g.
- the statement uses $r$, not $x$
- what does $y$ mean? i assume it's the prefix xor? if so, should rename it and make a comment for it

eysbutno (approved):
explanation and code are good

not sure if bitwise is the best place for it though? to me it feels more prefix sum / greedy

solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+id: usaco-1544
+source: USACO Silver 2026 First Contest
+title: Sliding Window Summation
+author: Aryaveer Shishodia
+---
+
+[Official Analysis (Python)](http://usaco.org/current/data/sol_prob3_silver_season26contest1.html)
+
+## Explanation
+The "remainder when the number of ones in the length-K window of $b$ with leftmost index $i$ is divided by two" is analogous to the XOR of the window. The only difference between the windows of $r_i$ and $r_{i+1}$ is that $r_i$'s window contains $b_i$, while $r_{i+1}$'s contains $b_{i+k}$.
```
lems9520138: Instead of quoting that long passage, I would recommend just saying

The value of r_i is equivalent to the XOR of the length-k window beginning at i.

solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+$$r_i\oplus r_{i+1} = b_i\oplus b_{i+k}$$
```
lems9520138: Completely optional, but for some people something like this might be easier
Utilizing the the statement above we know that,
(b_i XOR b_(i + 1) .... XOR (b_(i + k - 1) XOR (b_(i + 1) XOR b_(i + 2) ... XOR b_(i + k) = r_i XOR r_i + 1, 

which is equal to 
b_i XOR b_i + k

solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+Since we are given $r$, choosing the value of $b_i$ determines $b_{i+k}$, which in turn determines $b_{i+2k}$, $b_{i+3k}$, and so on. This splits $b$ into $k$ binary chains, each determined by one of the first $k$ characters of $b$. For each chain, we can set the first character to `0` and `1`, and compare which one will minimize the amount of `1` bits.
```
lems9520138:
```suggestion
Since we are given $r$, choosing the value of $b_i$ determines $b_{i+k}$, which in turn determines $b_{i+2k}$, $b_{i+3k}$, and so on. This splits $b$ into $k$ binary chains, each dependent on one of the first $k$ characters of $b$. For each chain, we can set the first character to `0` or `1`, comparing which will minimize the amount of `1` bits.
```

lems9520138: Just a thought but a visualization/image for what those binary chains look like could be helpful, like say something like k = 3
R _ _ _ R _ _ _ R ...

solutions/silver/usaco-1544.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+The only thing that restricts the first $k$ bits of $b$ is $r_1$, which will sometimes force us to flip one of our initial bits despite it increasing the number of `1` bits in the chain. We can check if our configuration satisfies $r_1$ by recording the bitwise XOR of the optimal first $k$ bits, and storing a running minimum $m$ the minimal cost of flipping a chain. If the bitwise XOR of the first k elements doesn't align with $r_1$, we can then flip that chain back and add $m$ to our count.
```
lems9520138:
```suggestion
The only thing that restricts the first $k$ bits of $b$ is $r_1$, which will sometimes force us to flip one of our initial bits despite it increasing the number of `1` bits in the string. We can check if our configuration satisfies $r_1$ by recording the bitwise XOR of the optimal first $k$ bits, and storing a running minimum $m$, representing the additional amount of `1` bits to flip a chain. If the bitwise XOR of the optimal first $k$ bits doesn't align with $r_1$, we can then flip the selected chain back and add $m$ to our count.
```

lems9520138: I just wanted a second opinion on this, but the phrasing "bitwise XOR of the optimal first $k$ bits" just feels a bit awkward if anyone has any suggestions.

solutions/silver/usaco-1544.mdx
AVS-Memer: I am thinking of a diagram like this:
```
n=11, k=3
r:        1 0 0 1 1 0 0 0 1
Indices:  0 1 2 3 4 5 6 7 8 9 10
          |     |     |     |
Chain 0:  1 --> 0 --> 0 --> 0
Chain 1:    0 --> 0 --> 1 --> 0
Chain 2:      0 --> 1 --> 0
b:        1 0 0 0 0 1 0 1 0 0 0
```
We process each chain independently, counting the number of `1` bits if we start with `0`. For Chain 0, starting with `0` would produce more `1`s than starting with `1`, so we flip the first bit of Chain 0. This flips the entire chain to minimize the total number of `1` bits. Chains 1 and 2 remain unchanged since starting with `0` already gives the fewest `1`s.


I would rather get some feedback first before adding it

lems9520138: Yea to me this looks fine.

solutions/silver/usaco-1544.mdx
eysbutno: i think it's fine tbh

solutions/silver/usaco-1544.mdx
eysbutno: yep looks good!

PR #5949, opened 2026-02-12
Add link to solution metadata for Count_Min [Advanced]

ashoknarayan:
This [problem set](https://drive.google.com/file/d/10253azIZ_X-GTsEF1Ffy8W9W_QtySImZ/view) has available [solution](https://mbit.mbhs.edu/archive/2020/advanced_editorial.pdf).

Since the [permalink](https://usaco.guide/adv/count-min#problem-mbit-ZookeepersGathering) points to [this](https://codeforces.com/gym/102621/problem/L) empty looking page, there is slight confusion. 

Only the last problem is being referred here

I think we can update the question link too. Can you guide me on what to do ?

### DIFF
--- a/content/6_Advanced/Count_Min.problems.json
+++ b/content/6_Advanced/Count_Min.problems.json
@@ -39,8 +39,8 @@
       "isStarred": false,
       "tags": [],
       "solutionMetadata": {
-        "kind": "link",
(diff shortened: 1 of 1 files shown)


bqi343:
can we set the solution metadata to be the same as the other CF problems?

```
      "solutionMetadata": {
        "site": "CF"
      }
```

and keep the problem link unchanged

ashoknarayan:
Done

PR #5951, opened 2026-02-12
Update solution kind to autogen

ashoknarayan:
[this](https://usaco.guide/adv/multiplicative#problem-gfg-GCDSum) question has solution just below it only.


ashoknarayan:
I updated the problem link to the gfg practice page of the problem and put the previous link as the solution page since it was the solution

bqi343:
actually we should probably just remove this problem entirely and replace it with one having larger constraints

I'll give this module a makeover at some point, most of it wasn't written by me ...

PR #5979, opened 2026-02-25
Mooclear Generator Editorial

johnzhou721:

I have not really formatted this yet; this is an intial writeup, and no reviews needed until cleanups.

- [x] I have tested my code.

### DIFF
--- a/content/4_Gold/Intro_Sorted_Sets.mdx
+++ b/content/4_Gold/Intro_Sorted_Sets.mdx
@@ -2,7 +2,7 @@
 id: intro-sorted-sets
 title: 'More Operations on Sorted Sets'
 author: Darren Yao, Benjamin Qi, Andrew Wang
-contributors: Aadit Ambadkar, Jeffrey Hu, Jason Sun
+contributors: Aadit Ambadkar, Jeffrey Hu, Jason Sun, Raagav Bala
 prerequisites:
   - intro-sets
   - binary-search-sorted-array


solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,195 @@
+---
+id: usaco-1543
+source: USACO Silver 2026 First Contest
+title: Mooclear Reactor
+author: John Zhou
+---
+
+[Official Analysis](https://usaco.org/current/data/sol_prob2_silver_season26contest1.html)
```
eysbutno:
```suggestion
[Official Analysis (C++)](https://usaco.org/current/data/sol_prob2_silver_season26contest1.html)
```

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,195 @@
...
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+#include <iostream>
+#include <map>
+#include <vector>
+using namespace std;
+#define int long long
```
eysbutno: get rid of define int long long

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,195 @@
...
+int l[200005], r[200005], lab[200005], m[200005], b[200005], ff[200005], minn[200005],
```
eysbutno: to quote sans:
> "there's a lot of magic numbers here"

preferably define them

(my personal preference is to lambda spam, but im fine if you dont wanna do that)

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,195 @@
...
+	for (int i = 0; i < g[x].size(); i++) {
+		int y = g[x][i];
+		int z = w[x][i];
+		if (y == fa) continue;
+		int m;
+		int b;
+		m = propag(x, y, z).first;
+		b = propag(x, y, z).second;
+		if (::m[y] == 0) {
+			::m[y] = m;
```
eysbutno: can we not 😭

eysbutno (commented):
explanation is fine, though:
- kinda verbose
- can go a bit more into the actual details of implementing

code:
- lot of magic numbers here
- variables are kind of hard to understand here
- can we not have global scope resolution constantly
- needs a lot more comments; this problem is kind of an implementation-fest, so it's hard to read w/o comments

i also think the edi code for the problem is prob cleaner; can maybe borrow parts of that? dunno

johnzhou721:
@eysbutno I was submitting as an in progress draft, not ready to review, and am aware about the issue.

solutions/silver/usaco-1543.mdx
johnzhou721: Done!

solutions/silver/usaco-1543.mdx
johnzhou721: Done!

solutions/silver/usaco-1543.mdx
johnzhou721: Done!

johnzhou721:
(for eysbutno: I'm still working on this.)

johnzhou721:
@eysbutno This is ready again.

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+[Official Analysis (C++)](https://usaco.org/current/data/sol_prob2_silver_season26contest1.html)
+
+# Solution
+
+## Explanation
+
+First, notice that the constraints $(x, y, z)$ form an undirected weighted graph, where the constraints
+are the edges and the generators are the vertices.  Each connected component can be optimized independently,
+and below we shall consider one connected component.
```
eysbutno:
```suggestion
The constraints $(x, y, z)$ form an undirected weighted graph, where the constraints
are the edges and the generators are the vertices.  We optimize each component independently.
```

or something more brief

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+Let us first consider the cases when the component has no cycles. In this case, we choose an arbitary vertex,
```
eysbutno: "component has no cycles" IMO just say it's a tree, since that's more recognizable and it's the same thing

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,307 @@
...
+largest amount of generators that can function in this component.
+
+Specifically, for implementation, we choose a node $v$ to start from in each connected comopnent (and name
+this connected component Component $v$), and
+represent $v$'s power as $x$, so the power will be $1x+0$, which is $m[v]=1$
+and $b[v]=0$. Then, for each node we traverse to, if we're going from node $u$ to node $v$ through
+constraint $a_u + a_v = c$, then we can represent $a_v=c-a_u=c-(m[u]x+b[u])=-m[u]x+(c-b[u])$, aka.
+$m[v]=-m[u]$ and $b[v]=(c-b[u])$. We do this until we represent every node; we also set the
+expression equal to the minimum and maximum values of each node, and solve; the lower value we get
+is the minimum $x$, and the higher the maximum $x$ for the generator to function.
```
eysbutno: IMO this can (and definitely should) be condensed significantly

as an example:
> Starting from vertex $v$, we traverse the tree, defining each node's value by its parent node and edge values.  For example, if we start from node $v$ (which has a value of $x$) to node $u$ with an edge of weight $5$, we can rewrite $a_u$ as $5-a_v$, or $5-x$. After traversing the tree, we can rewrite each node's value as a function of $x$, either in the form $a-x$ or $x-a$, where $a$ is a positive constant.
> 
> For each node, after calculating its value in terms of $x$, we calculate the minimum and maximum value $x$ can take to satisfy the stable operating range of the reactor. This reframes our problem as finding the value of $x$ that is inside the most ranges, which can be solved using a difference array with a map.

also another note: no point in considering m, since m is either 1 or -1. this should make the explanation and code way simpler

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,307 @@
...
+never occured; if there is exactly one solution, then we fix $x$ for this connected component, and if
+it's ever fixed again through this process to a different value; declare as unsolvable. We only need
+to consider this for a node we're about to revisit, and not further down the chain, as the solving
+process connects those two chains of representations already. Concretely, this means that if we're
+about to revisit a node, we still compute its new representation as described in the previous paragraph;
+then, we set the two representations equal, to see if we need to fix $x$ for the connected component.
+
+At the end, if there is a fixed value for a connected component, then instead of using a procedure similar
+to a difference array, we just check how many ranges the single fixed value satisifies.  Doing the
+above procedure for each connected component yields the answer.
```
eysbutno: this is pretty vague

also, IMO it would make more sense to explain this after explaining how to solve the tree variant, since it'd be kind of confusing to not explain why we're doing all the $x$ stuff until the end

solutions/silver/usaco-1543.mdx
```diff
@@ -0,0 +1,307 @@
...
+Now, what happens if we have cycles? In our traversal for each component, cycles break the assumption that
+we will not revisit nodes. When we do revisit one, we will get *two* different representations of power
+mandated for a generator. We may equate the expressions, and solve for $x$; if unsolvable, this problem
+is unsolvable; if infinitely many solutions, then we don't need to do anything, same as if this revisit
+never occured; if there is exactly one solution, then we fix $x$ for this connected component, and if
+it's ever fixed again through this process to a different value; declare as unsolvable. We only need
+to consider this for a node we're about to revisit, and not further down the chain, as the solving
+process connects those two chains of representations already. Concretely, this means that if we're
+about to revisit a node, we still compute its new representation as described in the previous paragraph;
+then, we set the two representations equal, to see if we need to fix $x$ for the connected component.
```
eysbutno:
```suggestion
Now, what happens if we have cycles? In our traversal for each component, cycles break the assumption that
we will not revisit nodes. When we do revisit one, we will get *two* different representations of power
mandated for a generator. We equate the expressions and solve for $x$, resulting in three cases:
- If there are no **integer** solutions, fulfilling the constraints is impossible.
- If there are **infinite** solutions, we do nothing.
- If there is **one** solution, we set the value of $x$ for this component.
```

dunno how your code handles different possibilities of $x$ for the last case; imo it's easiest to just pick the first one we find and just validate later surely

eysbutno (commented):
not gonna review code yet, but general advice for that:
- appreciate the comments, but don't need to go that thorough (e.g. javadoc comments are definitely overkill)
- as mentioned earlier, do not need to consider m in linear equations. since it's always 1 or - variable naming is still a bit confusing (e.g. why is $n$ lowercase and $M$ uppercase?)

johnzhou721:
> * appreciate the comments, but don't need to go that thorough (e.g. javadoc comments are definitely overkill)

javadoc was suggested by @SansPapyrus683 -- I'll be less verbose on other parts, though.  @SansPapyrus683 Any advice on whether to docstring in this case?



> * as mentioned earlier, do not need to consider m in linear equations. since it's always 1 or -

You mean "or -1"?  But then you still need to store the sign in a vraible and 1 and -1 is best representaiton, and a more generalized approach to calculations is more readable IMO and doesn't take taht much ilnes.

solutions/silver/usaco-1543.mdx
johnzhou721: Yes it's "imo it's easiest"  appproach you've described"

eysbutno:
> > * appreciate the comments, but don't need to go that thorough (e.g. javadoc comments are definitely overkill)
> 
> javadoc was suggested by @SansPapyrus683 -- I'll be less verbose on other parts, though. @SansPapyrus683 Any advice on whether to docstring in this case?
> 
> > * as mentioned earlier, do not need to consider m in linear equations. since it's always 1 or -
> 
> You mean "or -1"? But then you still need to store the sign in a vraible and 1 and -1 is best representaiton, and a more generalized approach to calculations is more readable IMO and doesn't take taht much ilnes.

ok javadoc is fine but the current ones are kind of long

may have misread ur code, but generalizing expressions just makes it more complicated for the explanation

solutions/silver/usaco-1543.mdx
johnzhou721: > also another note: no point in considering m, since m is either 1 or -1. this should make the explanation and code way simpler

Conceptually perhaps simpler, but specialcasing m is 1 or -1 (or however we represent it) everywhere is kind of unintuitive IMHO.

I'll take your suggestion, though, and add the deriviation and emphasize the recursive nature a bit more

johnzhou721: Nevermind... the example already suggests enough about the derivation

SansPapyrus683:
forgot about tabs HAHAHAHAHA
it's been so long

johnzhou721:
@eysbutno tagging for another review

@SansPapyrus683 thanks, your changes lgtm, any tips for cleaner code?

eysbutno:
> @eysbutno tagging for another review
> 
> @SansPapyrus683 thanks, your changes lgtm, any tips for cleaner code?

not the goat himself, but as someone who (I think) writes somewhat clean code:
- no define int long long
- comments (as needed and not super long usually)
- variable names (should be reasonably inferrable; doesn't have to be super super descriptive since short code anyways)
- ~~no using namespace std~~
- if you can make an implementation more straightforward, you should usually opt to

SansPapyrus683:
honestly the namespace thing doesn’t even matter to me now

eysbutno:
i no longer use namespace std, but honestly it doesn't really matter for CP

PR #5985, opened 2026-02-28
Adding Solution to Grid Completion under PIE [Platinum]

ashoknarayan:
[Permalink](https://usaco.guide/plat/PIE#problem-cses-2429)

- [x] I have tested my code.

### DIFF
--- a/solutions/platinum/cses-2429.mdx
+++ b/solutions/platinum/cses-2429.mdx
@@ -36,13 +36,7 @@ Instead of counting only valid permutations, we:
 2. Subtract invalid configurations,
 3. Correct overcounting using the **Principle of Inclusion–Exclusion (PIE)**.
 
-Overcounting happens when:
-
-- A row violates its `A` constraint,
-- A row violates its `B` constraint,
-- Multiple violations occur simultaneously.
-
-To systematically handle this, we classify rows and columns into structural categories.
+Overcounting happens when there is a violation.
 
 ---
 ### What is a Violation?


bqi343 (changes requested):
imo it's not clear what a violation is. do you mean when an A and a B get assigned to the same location?

separately, good to say how to optimize the solution to O(N^2) -- note that if i is fixed, then the remaining portions of the formula are independent of each other

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,240 @@
...
+(C_5-i-k)!
+$$
+
+This formula counts all valid completions while correcting overcounting via the Inclusion–Exclusion Principle.
+
+---
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(N^3)$
```
bqi343: add space

```suggestion
**Time Complexity:** $\mathcal{O}(N^3)$

```

ashoknarayan:
> imo it's not clear what a violation is. do you mean when an A and a B get assigned to the same location?
> 
> separately, good to say how to optimize the solution to O(N^2) -- note that if i is fixed, then the remaining portions of the formula are independent of each other

Violation means to assign a row to a column different from it's forces`A` or `B` position. 
Also I have optimised the solution to O(n2). Had not thought of that. Thank you.

bqi343:
> Violation means to assign a row to a column different from it's forcesA or B position.

I don't think this makes sense? what if the initial grid is empty? then there are no forced As or Bs but you still need to make sure you don't assign A or B to the same position

ashoknarayan:
Yes, my explanation is not good. I will clear this soon.

ashoknarayan:
I realized I didn’t account for the case where `A` and `B` could end up in the same cell. I’ve updated it to cover both collisions and breaking forced positions. Does it look correct now?

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+---
+### What is a Violation?
+
+In our counting approach, each row eventually assigns one column for `A` and one column for `B`.
+
+A **violation** occurs when a row’s assignments are inconsistent with a valid completion. This can happen in two ways:
+
+- The assigned `A` and `B` columns coincide (both placed in the same cell), or
+- A forced `A` or `B` position in the original grid is not respected.
```
bqi343: I still don't understand why there's a second bullet point here. isn't the first bullet point the only violation you need to consider?

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,307 @@
...
+The Inclusion–Exclusion step counts configurations where certain rows are allowed to violate this condition, and then corrects the overcounting using alternating signs.
+
+---
+
+### Category Definitions
+
+Let:
+
+- `C0`: Rows with neither `A` nor `B`
+- `C1`: Rows with only `B`, and its column has no `A`
```
bqi343: C1: do you mean the number of Bs without any As in the same row or column? seems weird to define it as "the number of rows satisfying x condition" when it's invariant to the grid being transposed

Same for C2?

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+Now observe:
+
+- The `j` terms depend only on `j`.
+- The `k` terms depend only on `k`.
+- There is no interaction between `j` and `k`.
+
+So the double sum factorizes into:
+
+( sum over j ) * ( sum over k )
```
bqi343: can you write out the formula explicitly?

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+Each of these can be computed in $\mathcal{O}(N)$.
+
+Since `i` runs at most N times, the total complexity becomes:
```
bqi343:
```suggestion
Since `i` runs at most $N$ times, the total complexity becomes:
```

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+$\mathcal{O}(N^2)$
```
bqi343: there's no need for this to be on a line of its own.

solutions/platinum/cses-2429.mdx
ashoknarayan: Yes

solutions/platinum/cses-2429.mdx
ashoknarayan: I have totally confused it. Sorry for the trouble. I think now it will be fine.

solutions/platinum/cses-2429.mdx
ashoknarayan: I removed it

solutions/platinum/cses-2429.mdx
```diff
@@ -0,0 +1,312 @@
...
+- Multiple violations occur simultaneously.
+
+To systematically handle this, we classify rows and columns into structural categories.
+
+---
+### What is a Violation?
+
+In our construction, each row is assigned one column for `A` and one column for `B`.
+
+A **violation** occurs if these two assignments coincide — that is, both `A` and `B` end up in the same cell of a row. Such a configuration is invalid.
```
bqi343, line 46: this line looks good now, though now I'm not sure what lines 41-42 are referring to.

solutions/platinum/cses-2429.mdx
ashoknarayan: I defined everything in row perspective and did PIE of columns over rows.

The condition is symmetric, but since I wrote in Row-centric way, I tried to define in same manner

solutions/platinum/cses-2429.mdx
ashoknarayan: Just changed it now.

ashoknarayan: You are really fast.

bqi343 (approved):

PR #5996, opened 2026-03-03
Update Problem Difficulty

AksLolCoding:

- [x] I have tested my code.

Problem difficulty is correct now

### DIFF
--- a/content/extraProblems.json
+++ b/content/extraProblems.json
@@ -2700,7 +2700,7 @@
       "name": "Mooclear Reactor",
       "url": "https://usaco.org/index.php?page=viewproblem2&cpid=1543",
       "source": "Silver",
-      "difficulty": "N/A",
+      "difficulty": "Easy",
       "isStarred": false,
       "tags": [],
       "solutionMetadata": {


eysbutno:
easy 💀 gg

bqi343:
According to https://usaco.guide/general/using-this-guide#how-to-use-this-guide, difficulty is relative to other problems in the module.

> Difficulty represents how challenging a problem is expected to be to someone after they read through the module, and is not comparable across modules.

But this problem isn't in a module so perhaps let's just leave the difficulty blank?

@eysbutno thoughts on this?

(also, this problem isn't easy for Silver ...)

content/extraProblems.json
```diff
@@ -2700,7 +2700,7 @@
       "name": "Mooclear Reactor",
       "url": "https://usaco.org/index.php?page=viewproblem2&cpid=1543",
       "source": "Silver",
-      "difficulty": "N/A",
+      "difficulty": "Easy",
```
AsynchronousAI, line 2703:
```suggestion
      "difficulty": "Normal",
```
I would have personally placed it as "Hard", but honestly "Easy" is just way too generous.

eysbutno:
> According to https://usaco.guide/general/using-this-guide#how-to-use-this-guide, difficulty is relative to other problems in the module.
> 
> > Difficulty represents how challenging a problem is expected to be to someone after they read through the module, and is not comparable across modules.
> 
> But this problem isn't in a module so perhaps let's just leave the difficulty blank?
> 
> @eysbutno thoughts on this?
> 
> (also, this problem isn't easy for Silver ...)

I don't think you can see the labelled difficulty in the monthlies table anyways

I'll close the PR for now

bqi343:
The difficulty can be viewed in problem search.

lems9520138:
> The difficulty can be viewed in problem search.

I think for this maybe we should reopen and set difficulty to normal.

PR #6084, opened 2026-03-29
Patch 1

lems9520138:

- [x ] I have tested my code.

### DIFF
--- a/solutions/bronze/usaco-1445.mdx
+++ b/solutions/bronze/usaco-1445.mdx
@@ -15,9 +15,9 @@ For example, had the string been $\text{ABCDEF}$ and the $\text{C}$ was modified
 
 Using this, we first precompute a map representing the counts of each moo that was originally present in the string. Then, we can try every single possible edit to the string, with a total of $\mathcal{O}(26N)$ possible modifications. While trying every single possible edit to a specific position, due to only three substrings being affected, the global answer does not change by much. So, we can adjust the frequencies of moo's based on the $3$ substrings being affected after every modification, and check whether any moo's exceeded a frequency of $F$ due to the modifications. It is important to reverse the specific position back to its original character after the modifications, since only one position can be modified at a time.
 
-If any moo substring ever reaches a frequency of $f$ before or after modifications, it can be added to the set of answers.
+If any moo substring ever reaches a frequency of $F$ before or after modifications, it is added to the set of answers.
 
-While technically the time complexity is $\mathcal{O}(N)$, the factor of 26 in this case is significant because it represents the number of letters we bruteforce over.
+While technically the time complexity is $\mathcal{O}(N)$, the factor of 26 in this case is significant because it represents the number of letters we brute-force over.
 
 ## Implementation
 


content/extraProblems.json
```diff
@@ -508,6 +508,19 @@
...
+      "name": "It's Mooin' Time",
+      "url": "https://usaco.org/index.php?page=viewproblem2&cpid=1445",
+      "source": "Bronze",
+      "difficulty": "Easy",
+      "isStarred": false,
+      "tags": ["Complete Search"],
+      "solutionMetadata": {
+        "kind": "internal"
+      }
+    },
```
eysbutno: is this not already somewhere in guide? i would expect it to be at least

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+[Official Analysis (C++, Python)](https://usaco.org/current/data/sol_prob3_bronze_dec24.html)
+
+## Explanation
+
+Notice that modifying exactly one character can only affect at most 3 substrings.
+
+For example, had the string been $\text{ABCDEF}$ and the $\text{C}$ was modified to a $\text{Z}$, the affected substrings would now be, $\text{ABZ}$, $\text{BZD}$, and $\text{ZDE}$.
+
+Using this, we can first precompute a map representing the counts of each moo that was originally present in the string. Then, we can try every single possible edit to the string, with a total of $\mathcal{O}(26N)$ possible modifictaions. For each one of these changes, subtract the counts of the affected moos, and add the new counts.
```
eysbutno: > subtract the counts of the affected moos, and add the new counts.

this is a bit vague. maybe be a bit more specific? ie mentioning the exact steps/ideas:
- because only 3 substrings affected, global answer doesn't change much
- need to calculate the difference between states for each moo string affected

or something idk

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+If any string ever reaches a frequency of $F$ before or after modifictations, it can be added to the answers.
```
eysbutno:
```suggestion
If any string ever reaches a frequency of $F$ before or after modifications, it can be added to the set of answers.
```

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,114 @@
...
+	cin >> n >> f;
+	string c;
+	cin >> c;
+
+	set<string> answers;
+	map<string, int> counts;
+
+	// Function to check whether a given sequence starting at position i is a MOO
+	auto check = [&](int i) {
+		if (i < 0 or i + 2 >= n) return false;
```
eysbutno: 1. for lambdas, I'd prefer having a trailing return type (e.g. `[&](int i) -> bool { ... }`)
2. `is_moo` would be a better name
3. I think using `or` instead of `||` and same for `&&` is slightly unusual; it's not a huge deal but would prefer to swap it out

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+For example, had the string been $\text{ABCDEF}$ and the $\text{C}$ was modified to a $\text{Z}$, the affected substrings would now be, $\text{ABZ}$, $\text{BZD}$, and $\text{ZDE}$.
+
+Using this, we can first precompute a map representing the counts of each moo that was originally present in the string. Then, we can try every single possible edit to the string, with a total of $\mathcal{O}(26N)$ possible modifictaions. For each one of these changes, subtract the counts of the affected moos, and add the new counts.
+
+If any string ever reaches a frequency of $F$ before or after modifictations, it can be added to the answers.
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(26N)$
```
eysbutno, line 24: maybe use a constant $C$ instead of 26? cuz usually we don't include the pure numbers in time complexity, but it is fairly significant here

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,114 @@
...
+	string temp;
+
+	for (int i = 0; i < n; i++) {
+		char original = c[i];
+		bool done = false;
+
+		// Test each possible modifications plus the original character that was in the
+		// string This makes sure that the original counts of the string is restored
+		// afterwards
+		for (char mod = 'a'; mod <= 'z' + 1 && !done; mod++) {
```
eysbutno: i feel like this part of the code could be slightly more 'elegant' (e.g. why is done here? can't you just reset to original after the loop?)

in a similar vein, could loop over i - 2 to i, tho it doesn't matter too much

content/extraProblems.json
lems9520138: i thought it was but ig not, its in monthlies but i couldnt find metadata anywhere

solutions/bronze/usaco-1445.mdx
lems9520138: wym loop over i - 2 to i??

I can modify my comment to explain it better or recode it.

The main rational behind it was that it was shorter code... :| instead of having to manually test the regular posiiton, each modification basically does the modification so i could test every single modification + the original one which would modify it back to the original position.

If this is better, i could also rewrite the for loop so that it does every character once, (starts at letter + 1 in ascii, loops 26 times and do % 26 to get the current letter)

solutions/bronze/usaco-1445.mdx
lems9520138: where would I write C = 26, representing the numbers of letters.

Ig pure numbers is bad but i js thought since it alr kinda represents the # of letters.

solutions/bronze/usaco-1445.mdx
lems9520138: done

lems9520138:
Still need to fix the explanation a bit, not perfect but lmk if its a little bit clearer.

solutions/bronze/usaco-1445.mdx
guodavid0418: maybe mention in the explanation that while the time is technically O(N), the factor of 26 is significant

content/extraProblems.json
eysbutno: https://usaco.guide/bronze/intro-sets#problem-usaco-1445 :P 

always good to do a search of the repo first for this typa thing

lems9520138: oh 😭 i did not know you could do that mb.
Wait why is this hard??? Ig it needs maps knowledge but does that constitute it for a hard??

solutions/bronze/usaco-1445.mdx
eysbutno: ok not necessarily looped but a lot of code is kind of repeated

edi code is pretty clean in this regard: https://usaco.org/current/data/sol_prob3_bronze_dec24.html

content/extraProblems.json
```diff
@@ -508,6 +508,19 @@
         "kind": "internal"
       }
     },
+
+    {
```
eysbutno, line 512: I don't think you need to add this here? since it's already in intro sets, unless you wanna add it to complete search

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+[Official Analysis (C++, Python)](https://usaco.org/current/data/sol_prob3_bronze_dec24.html)
+
+## Explanation
+
+Notice that modifying exactly one character can only affect at most $3$ substrings.
+
+For example, had the string been $\text{ABCDEF}$ and the $\text{C}$ was modified to a $\text{Z}$, the affected substrings would now be, $\text{ABZ}$, $\text{BZD}$, and $\text{ZDE}$.
+
+Using this, we can first precompute a map representing the counts of each moo that was originally present in the string. Then, we can try every single possible edit to the string, with a total of $\mathcal{O}(26N)$ possible modifictaions. While trying every single possible edit to a specific position, due to only three substrings being affected, the global answer does not change by much. So, we can adjust the frequencies of moo's based on the $3$ substrings being affected after every modification, and check whether any moo's exceeded a frequency of $f$ due to the modifications. It is important to reverst the specific position back to its original character after the modifications, since only one position can be modified at a time.
```
eysbutno:
```suggestion
Using this, we first precompute a map representing the counts of each moo that was originally present in the string. Then, we can try every single possible edit to the string, with a total of $\mathcal{O}(26N)$ possible modifications. While trying every single possible edit to a specific position, due to only three substrings being affected, the global answer does not change by much. So, we can adjust the frequencies of moo's based on the $3$ substrings being affected after every modification, and check whether any moo's exceeded a frequency of $F$ due to the modifications. It is important to reverse the specific position back to its original character after the modifications, since only one position can be modified at a time.
```

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+If any moo substring ever reaches a frequency of $f$ before or after modifications, it can be added to the set of answers.
+
+While technically the time complexity is $\mathcal{O}(N)$, the factor of 26 in this case is significant because it represents the number of letters we bruteforce over.
```
eysbutno:
```suggestion
If any moo substring ever reaches a frequency of $F$ before or after modifications, it is added to the set of answers.

While technically the time complexity is $\mathcal{O}(N)$, the factor of 26 in this case is significant because it represents the number of letters we brute-force over.
```

or just use a constant C

content/extraProblems.json
lems9520138: Right I can remove it, complete search alr has enough questions

solutions/bronze/usaco-1445.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+<PySection>
+
+
+</PySection>
+
+<JavaSection>
+
+
+</JavaSection>
```
SansPapyrus683, line 103: why were these left in 🥀

eysbutno: gg 💀

@lems9520138

apex-kevin-s: with lems's permission i could make a pr that sansifies

lems9520138: yea if you want go ahead otherwise i can do it

lems9520138: > why were these left in 🥀

good catch lmao 💀

PR #6129, opened 2026-04-16
Edit internal solution for cf-762b (USB vs PS/2)

4nav:

- [x] I have tested my code.


solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,151 @@
...
-title: USB vs. PS/2
-author: Benjamin Qi
+id: cf-762b
+source: Codeforces Round 395 Div. 2
+title: USB vs PS/2
+author: Abhinav Singh
 ---
 
-[Official Editorial](https://codeforces.com/blog/entry/50010)
+[Problem Link](https://codeforces.com/contest/762/problem/B)
```
eysbutno: official editorial link is correct

problem link is automatically generated w/ the solution

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,151 @@
...
+- `ps2Ports`: PS/2-only ports
+- `universalPorts`: ports that accept either type
+
+Each mouse:
+- has a type (USB or PS/2)
+- has a cost
+
+Our goal is to **buy the maximum number of mice while minimizing total cost**.
+
+---
```
eysbutno: you don't need to restate the problem unless the problem statement is super convoluted or you're reframing it

also the extra separators `---` aren't necessary

apex-kevin-s:
This seems AI-generated.

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,151 @@
...
+
+### Fill USB and PS/2 ports first
```
eysbutno: this part of the explanation can (and should IMO) be condensed a lot

I think you can just sequence the explanation like:
- establish that universal ports are more flexible
  - greedy intuition is that, because they are more flexible, they can be handled later
  - this motivates us to like, try to fulfill all the non-universal mice and handle them later, esp since we need to maximize the number of mice obtained
- short expl on why this is optimal (i think exchange argument would be a better way to explain?)

eysbutno (commented):
hello! thanks for the contribution 

one note: the original editorial also had a 2P solution, where you instead try to like consider exchanging usb mice for ps2 mice instead -- I think you should leave that approach in, even if you don't plan on writing an edi for it

lems9520138:
Think you forgot to update metadata or smthing, when you click the solution on guide you get 404 not found error.

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,151 @@
 ---
-id: cf-762B
-source: CF
-title: USB vs. PS/2
-author: Benjamin Qi
+id: cf-762b
```
eysbutno: this needs to be capitalized, that's why 404

(cf-762B)

4nav:
> This seems AI-generated.
i used grammarly to make my sentences more concise but apart frm that not rlly, i js kept the tone as though i was explaining to a newbie

solutions/silver/cf-762B.mdx
4nav: done!

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,127 @@
 ---
 id: cf-762B
-source: CF
-title: USB vs. PS/2
-author: Benjamin Qi
+source: Codeforces Round 395 Div. 2
+title: USB vs PS/2
+author: Abhinav Singh
```
eysbutno:
```suggestion
source: CF
title: USB vs PS/2
author: Abhinav Singh, Benjamin Qi
```

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,127 @@
...
+- take the cheapest `usbPorts` USB mice
+- take the cheapest `ps2Ports` PS/2 mice
+
+At this stage, we’ve used the restricted ports as efficiently as possible.
+
+Now look at the remaining mice: universal ports don’t care about type anymore, so we just want the cheapest options overall.
+
+- put all unused mice into one array
+- sort it
+- take the cheapest `universalPorts` mice
```
eysbutno, line 32: IMO this should still be condensed

mainly, you don't really neeed to explain the implementation, just the idea. here, the idea can be succintly described something like
> Thus, it makes more sense to deal with restricted ports first. We sort all of the mice and greedily pair mice with computers. For each mouse, we first try to pair it with a computer restricted to that mouse's port, or pair it with a computer supporting both ports if all the restricted computers of that type have been paired.

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,127 @@
...
+
+### Why this works
+
+The key is delaying the use of flexible ports.
+
+If an optimal solution ever uses a universal port for a mouse while a restricted port is filled with a more expensive mouse of the same type, we can swap them. The cheaper mouse goes to the restricted port, and the more expensive one can be moved or discarded. This never reduces the number of mice used and never increases the cost.
+
+So there is always an optimal solution where we fill restricted ports first with the cheapest possible mice, and only then use universal ports.
```
eysbutno, line 40: could consider putting this in a spoiler

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,127 @@
...
+
+### Alternative approach
```
eysbutno, line 42: formatting wise would prefer something like

```
# Solution 1 (Greedy)

## Explanation

## Implementation

# Solution 2 (Two Pointers)

## Explanation

(can put this blurb in it)

## Implementation

(can just keep benq's code or demacro if you want)
```

4nav: sure, i'll demacro benq's code

solutions/silver/cf-762B.mdx
4nav: i mean if its an internal solution is spoilering necessary?

solutions/silver/cf-762B.mdx
4nav: sure

solutions/silver/cf-762B.mdx
eysbutno: sometimes for extra proofs/"why does this work" info, we put it in spoilers

but not particularly necessary

4nav:
is it good now?

eysbutno:
> is it good now?

I'll review later today

at a glance, code looks fine, for explanation I think it's generally better to explain intuition/reasoning before directly writing the algorithm out

4nav:
any changes needed?

lems9520138:
> any changes needed?

Hollup lemme review mb

4nav:
> > any changes needed?
> 
> Hollup lemme review mb

yeah anything wrong?

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,207 @@
...
-title: USB vs. PS/2
-author: Benjamin Qi
+title: USB vs PS/2
+author: Abhinav Singh, Benjamin Qi
 ---
 
 [Official Editorial](https://codeforces.com/blog/entry/50010)
 
-## Greedy
+## Explanation
```
eysbutno, line 10: I would personally organize the explanation differently

smth like:
- Computers with universal ports are more flexible, so if we use them first, we may screw over a restricted port later
- Any optimal solution where a universal port computer takes a cheaper mouse than a computer with the mouse's restricted type can be converted to one where universal takes the more expensive one (e.g. your third part)
- then follow up w/ a sentence or two explaining the actual algo

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,207 @@
...
 
-As described in editorial.
+Universal ports are more flexible than restricted ones—they can accept either type. So we should avoid using them unless necessary.
+
+If we assign a mouse that could go to a restricted port into a universal port, we might later be forced to use a more expensive mouse for that restricted port.
+
+
+### Greedy Strategy
```
guodavid0418: i'd just go straight into explaining the greedy algo w/o needing to list stuff like this, and then explain its correctness as you go or after it, but an extra section header "why this works" is unnecessary imo, just split into paragraphs

4nav: ight bet

solutions/silver/cf-762B.mdx
```diff
@@ -1,377 +1,207 @@
...
+
+- Sort all mice by cost
+- For each mouse:
+  - If a matching restricted port is available, use it
+  - Otherwise, use a universal port (if available)
+- After all restricted ports are filled, assign the cheapest remaining mice to universal ports
+
+
+### Why this works
```
lems9520138: Yeah again, this paragraph is kinda js unnecessary, you can insert it within the Greedy Strategy one.

4nav: once again, done

guodavid0418 (approved):

eysbutno (approved):

PR #6135, opened 2026-04-17
Add internal solution for cf-1244E

4nav:

- [x] I have tested my code.

### DIFF
--- a/solutions/silver/cf-1244E.mdx
+++ b/solutions/silver/cf-1244E.mdx
@@ -61,48 +61,48 @@ Thus, it is enough to check:
 using namespace std;
 using ll = long long;
 
-bool check(ll d, int n, const vector<ll>& a, const vector<ll>& pref, ll k) {
-    for (int i = 0; i < n; ++i) {
-        // Case 1: L = a[i]
-        {
-            ll L = a[i], R = L + d;
-            ll costLeft = (ll)i * L - (i > 0 ? pref[i - 1] : 0LL);
-            int r = upper_bound(a.begin(), a.end(), R) - a.begin();
-            ll costRight = r < n ? (pref[n - 1] - (r > 0 ? pref[r - 1] : 0LL)) - (ll)(n - r) * R : 0LL;
-            if (costLeft + costRight <= k) return true;
-        }
-        // Case 2: L = a[i] - d
-        {
-            ll R = a[i], L = R - d;
-            int p = lower_bound(a.begin(), a.end(), L) - a.begin();
-            ll costLeft = (ll)p * L - (p > 0 ? pref[p - 1] : 0LL);
-            ll costRight = (pref[n - 1] - pref[i]) - (ll)(n - i - 1) * R;
-            if (costLeft + costRight <= k) return true;
-        }
-    }
-    return false;
+bool check(ll d, int n, const vector<ll> &a, const vector<ll> &pref, ll k) {
+	for (int i = 0; i < n; ++i) {
+		// Case 1: L = a[i]
+		{
+			ll L = a[i], R = L + d;
+			ll costLeft = (ll)i * L - (i > 0 ? pref[i - 1] : 0LL);
+			int r = upper_bound(a.begin(), a.end(), R) - a.begin();
+			ll costRight =
+			    r < n ? (pref[n - 1] - (r > 0 ? pref[r - 1] : 0LL)) - (ll)(n - r) * R
+			          : 0LL;
+			if (costLeft + costRight <= k) return true;
+		}
+		// Case 2: L = a[i] - d
+		{
+			ll R = a[i], L = R - d;
+			int p = lower_bound(a.begin(), a.end(), L) - a.begin();
+			ll costLeft = (ll)p * L - (p > 0 ? pref[p - 1] : 0LL);
+			ll costRight = (pref[n - 1] - pref[i]) - (ll)(n - i - 1) * R;
+			if (costLeft + costRight <= k) return true;
+		}
+	}
+	return false;
 }
 
 int main() {
-    ios::sync_with_stdio(false);
-    cin.tie(nullptr);
-    ll n, k;
-    cin >> n >> k;
-    vector<ll> a(n);
-    for (auto& x : a) cin >> x;
-    sort(a.begin(), a.end());
-    vector<ll> pref(n);
-    pref[0] = a[0];
-    for (int i = 1; i < n; i++) {
-        pref[i] = pref[i - 1] + a[i];
-    }
-    ll lo = 0, hi = a[n - 1] - a[0];
-    while (lo < hi) {
-        ll mid = (lo + hi) / 2;
-        if (check(mid, n, a, pref, k)) hi = mid;
-        else lo = mid + 1;
-    }
-    cout << lo << "\n";
+	ios::sync_with_stdio(false);
+	cin.tie(nullptr);
+	ll n, k;
+	cin >> n >> k;
+	vector<ll> a(n);
+	for (auto &x : a) cin >> x;
+	sort(a.begin(), a.end());
+	vector<ll> pref(n);
+	pref[0] = a[0];
+	for (int i = 1; i < n; i++) { pref[i] = pref[i - 1] + a[i]; }
+	ll lo = 0, hi = a[n - 1] - a[0];
+	while (lo < hi) {
+		ll mid = (lo + hi) / 2;
+		if (check(mid, n, a, pref, k)) hi = mid;
+		else lo = mid + 1;
+	}
+	cout << lo << "\n";
 }
 ```
 
@@ -171,61 +171,59 @@ Thus, it is enough to check:
 using namespace std;
 using ll = long long;
 
-bool check(ll d, int n, const vector<ll>& a, const vector<ll>& pref, ll k) {
-    for (int i = 0; i < n; ++i) {
-        // Case 1: L = a[i]
-        {
-            ll L = a[i], R = L + d;
-            ll costLeft = (ll)i * L - (i > 0 ? pref[i - 1] : 0LL);
-            int r = upper_bound(a.begin(), a.end(), R) - a.begin();
-            ll costRight = r < n ? (pref[n - 1] - (r > 0 ? pref[r - 1] : 0LL)) - (ll)(n - r) * R : 0LL;
-            if (costLeft + costRight <= k) return true;
-        }
-        // Case 2: L = a[i] - d
-        {
-            ll R = a[i], L = R - d;
-            int p = lower_bound(a.begin(), a.end(), L) - a.begin();
-            ll costLeft = (ll)p * L - (p > 0 ? pref[p - 1] : 0LL);
-            ll costRight = (pref[n - 1] - pref[i]) - (ll)(n - i - 1) * R;
-            if (costLeft + costRight <= k) return true;
-        }
-    }
-    return false;
+bool check(ll d, int n, const vector<ll> &a, const vector<ll> &pref, ll k) {
+	for (int i = 0; i < n; ++i) {
+		// Case 1: L = a[i]
+		{
+			ll L = a[i], R = L + d;
+			ll costLeft = (ll)i * L - (i > 0 ? pref[i - 1] : 0LL);
+			int r = upper_bound(a.begin(), a.end(), R) - a.begin();
+			ll costRight =
+			    r < n ? (pref[n - 1] - (r > 0 ? pref[r - 1] : 0LL)) - (ll)(n - r) * R
+			          : 0LL;
+			if (costLeft + costRight <= k) return true;
+		}
+		// Case 2: L = a[i] - d
+		{
+			ll R = a[i], L = R - d;
+			int p = lower_bound(a.begin(), a.end(), L) - a.begin();
+			ll costLeft = (ll)p * L - (p > 0 ? pref[p - 1] : 0LL);
+			ll costRight = (pref[n - 1] - pref[i]) - (ll)(n - i - 1) * R;
+			if (costLeft + costRight <= k) return true;
+		}
+	}
+	return false;
 }
 
 int main() {
-    ios::sync_with_stdio(false);
-    cin.tie(nullptr);
+	ios::sync_with_stdio(false);
+	cin.tie(nullptr);
 
-    ll n, k;
-    cin >> n >> k;
+	ll n, k;
+	cin >> n >> k;
 
-    vector<ll> a(n);
-    for (auto &x : a) {
-        cin >> x;
-    }
+	vector<ll> a(n);
+	for (auto &x : a) { cin >> x; }
 
-    sort(a.begin(), a.end());
+	sort(a.begin(), a.end());
 
-    vector<ll> pref(n);
-    pref[0] = a[0];
-    for (int i = 1; i < n; i++) {
-        pref[i] = pref[i - 1] + a[i];
-    }
+	vector<ll> pref(n);
+	pref[0] = a[0];
+	for (int i = 1; i < n; i++) { pref[i] = pref[i - 1] + a[i]; }
 
-    ll lo = 0, hi = a[n - 1] - a[0];
+	ll lo = 0, hi = a[n - 1] - a[0];
 
-    while (lo < hi) {
-        ll mid = (lo + hi) / 2;
+	while (lo < hi) {
+		ll mid = (lo + hi) / 2;
 
-        if (check(mid, n, a, pref, k)) {
-            hi = mid;
-        } else {
-            lo = mid + 1;
-        }
-    }
+		if (check(mid, n, a, pref, k)) {
+			hi = mid;
+		} else {
+			lo = mid + 1;
+		}
+	}
 
-    cout << lo << "\n";
+	cout << lo << "\n";
 }
 ```
 
--- a/solutions/silver/cf-1244E.mdx
+++ b/solutions/silver/cf-1244E.mdx
@@ -1,11 +1,11 @@
 ---
 id: cf-1244E
-source: Codeforces
+source: CF
 title: Minimizing Difference
 author: Abhinav Singh
 ---
 
-[Official Editorial](https://codeforces.com/blog/entry/70553)
+[Official Analysis](https://codeforces.com/blog/entry/70553)
 
 ## Explanation
 
--- a/solutions/silver/cf-1244E.mdx
+++ b/solutions/silver/cf-1244E.mdx
@@ -15,7 +15,7 @@ We want to minimize the difference between the maximum and minimum values after 
 
 Instead of directly constructing the final array, fix a target difference `d` and check if it is achievable.
 
-If we can make all elements lie inside some interval of length `d`, then any larger `d` will also work.  
+If we can make all elements lie inside some interval of length `d`, then any larger `d` will also work.
 So the check is monotonic, and we can binary search on `d`.
 
 ### Checking a fixed d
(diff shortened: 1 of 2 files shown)


4nav:
there was no internal solution so i added one

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
+---
+id: cf-1244E
+source: Codeforces
+title: Minimizing Difference
+author: Abhinav Singh
+---
+
+[Official Editorial](https://codeforces.com/blog/entry/70553)
```
eysbutno:
```suggestion
---
id: cf-1244E
source: CF
title: Minimizing Difference
author: Abhinav Singh
---

[Official Analysis](https://codeforces.com/blog/entry/70553)
```

also need to link solution metadata (ie go to the module's problems.json and change solution type from external/autogenerated to internal)

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
...
+		if (check(mid, n, a, pref, k)) hi = mid;
+		else lo = mid + 1;
+	}
+	cout << lo << "\n";
+}
+```
+
+</CPPSection>
+
+</LanguageSection>---
```
eysbutno, line 109: this is repeated twice?

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+Thus, it is enough to check:
+- `L = a[i]`
+- `L = a[i] - d`
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(n \log n \log A)$, where $A$ is the initial range of values ($a[n-1] - a[0]$ after sorting).
+
+**Note:** Use `long long`. The total cost can reach about $10^{14}$.
```
eysbutno:
```suggestion
**Time Complexity:** $\mathcal{O}(N \log N \log A)$, where $A$ is the maximum value of $a_i$.
```

usually dont need to warn about long long unless it's subtle (not really here)

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+Fix a target difference `d` and check:
+
+> Can we make all elements lie inside some interval `[L, L + d]` using at most `k` operations?
+
+This works because the condition is monotonic:
+- If it is possible for some `d`, then it is also possible for any larger `d`
+- A larger window only makes it easier to fit all elements within `k` operations
+
+So we can binary search on `d`.
```
eysbutno, line 24: this can be condensed, since reader at this point should have some familiarity w/ binary search

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+As we slide `[L, L + d]`, the cost only changes when elements enter or leave the window:
+- when some `a[i]` enters from the left → `L = a[i]`
+- when some `a[i]` enters from the right → `L = a[i] - d`
+
+Between these points, which elements are inside the window does not change, so the cost is linear (constant slope).
+
+Thus, it is enough to check:
+- `L = a[i]`
+- `L = a[i] - d`
```
eysbutno, line 47: you should explain why it's safe to pick a value of $a_i$ to either be the start or end of our range, since that part is probably less intuitive than the actual calculating part

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,232 @@
...
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+using ll = long long;
+
+bool check(ll d, int n, const vector<ll> &a, const vector<ll> &pref, ll k) {
```
eysbutno, line 62: maybe use a lambda? would save a lot of the function arguments

solutions/silver/cf-1244E.mdx
4nav: hmm fair

solutions/silver/cf-1244E.mdx
4nav: oops

solutions/silver/cf-1244E.mdx
4nav: the way i understood was that cost was only changing when some a[i] was entering or exiting the window, which i TRIED to explain but idk how to make the explanation simpler tbf

solutions/silver/cf-1244E.mdx
4nav: sure

solutions/silver/cf-1244E.mdx
eysbutno: > the way i understood was that cost was only changing when some a[i] was entering or exiting the window

not necessarily; consider $a=[1, 1, 5, 10]$ and the target range $[6, 6]$; if we move this range to $[7, 7]$ or some similar case, cost increases

the reasoning is that any solution $[L, L + D]$ that doesn't have an endpoint on a value of $a_i$ would be at least as good (or better) if an endpoint was used, since if there are $A$ elements smaller than $L$ and $B$ elements bigger than $L+D$, we save $|A-B|$ operations by shifting it I think

solutions/silver/cf-1244E.mdx
eysbutno: solution is still unlinked btw

4nav: yeah i was kinda busy yesterday, doing rn

solutions/silver/cf-1244E.mdx
4nav: done

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,125 @@
...
+### Checking a fixed d
+
+Sort the array.
+
+We want every element inside `[L, L + d]`:
+- If `a[i] < L`, increase it to `L`
+- If `a[i] > L + d`, decrease it to `L + d`
+- Otherwise, leave it
+
+We compute the total cost using prefix sums.
```
eysbutno: please condense this, since you're just trying to get the point across that given $d$ and a range $[L, L+d]$ it's easy to validate

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+So unless the interval is already balanced, we can move it in one direction to improve (or not worsen) the cost.
+
+This means we can keep shifting the interval until one of its endpoints lands on some `a[i]`.
+
+Therefore, it is enough to check:
+- `L = a[i]`
+- `L = a[i] - d`
+
+for all `i`.
```
eysbutno: this part should be condensed, cuz you could prob just say like

> Thus, the net change of cost in shifting the interval is $|A - B|$, assuming the interval's endpoints do not fall on any values of $a_i$. Since it is always at least as optimal to move our interval endpoints to values of $a_i$, we can reduce the values of $L$ and $L+d$ to only values of $a_i$.

eysbutno (commented):
please use math formatting instead of code formatting btw (e.g. for some of the math stuff)

solutions/silver/cf-1244E.mdx
4nav: sure will do

4nav:
> please use math formatting instead of code formatting btw (e.g. for some of the math stuff)

i just learnt abt math mode on github, will use!

eysbutno (approved):
looks good now

I refactored the code to add comments and use 1 indexed prefix sums cuz there were a lot of `i > 0 ? pref[i - 1] : 0` ternary operators lol

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+source: CF
+title: Minimizing Difference
+author: Abhinav Singh
+---
+
+[Official Analysis](https://codeforces.com/blog/entry/70553)
+
+## Explanation
+
+Instead of constructing the final array, fix a target difference $d$ and check if it is achievable.
```
lems9520138:
```suggestion
Instead of trying to construct a final array, fix a target difference $d$ and check if it is achievable in under $k$ operations.
```

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+If all elements can be moved into some interval $[L, L + d]$, then any larger $d$ also works. Hence, the condition is monotonic and we can binary search on $d$.
```
lems9520138:
```suggestion
If all elements can be moved into some interval of length $d$, then it can be shown that all larger values of $d$ work. Hence, the condition is monotonic and we can binary search on $d$.
```

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+For a fixed $d$, after sorting, we try to fit all elements into $[L, L + d]$:
```
lems9520138:
```suggestion
For a fixed $d$, we try to fit all elements into $[L, L + d]$:
```

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+- increase elements below $L$
+- decrease elements above $L + d$
```
lems9520138: From here and on you use this variable $L$, but you never defined it. So either js say in your editorial that 
- you need to move the elements into some interval of length d
- You need to move the lements into some interval [L, L + d], where L is some arbitrary starting point of a interval or smthing.

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+The total cost can be computed using prefix sums.
```
lems9520138: I would be a bit more specific as to what total cost you are referring to here.

The total cost for performing the operations thus the max difference is $d$ can be computed via prefix sums ..

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+### Choosing $L$
```
lems9520138: This whole section is very overcomplicated IMO.

It suffices to just say something like L or L + d of some interval must belong to the array, and then show why, kinda like the official editorial does.

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+Let $A$ be the number of elements with $a_i < L$, and $B$ be the number of elements with $a_i > L + d$.
+
+Shifting $[L, L + d]$ (without crossing any $a_i$) changes the cost by $A - B$.
```
lems9520138: what you mean by "shifting" is a bit unclear

Specificy if you are shifting L to increase L or something.

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,114 @@
...
+Hence, we can move the interval to improve or preserve the cost until an endpoint hits some $a_i$.
+
+If the left endpoint hits $a_i$, then $L = a_i$.
+If the right endpoint hits $a_i$, then $L + d = a_i \Rightarrow L = a_i - d$.
+
+Thus, it suffices to consider
+$L = a_i \text{ or } L = a_i - d$.
+
```
lems9520138, line 63: Maybe I missed it or something

Can you add a section that explains now how with all the L positions you now got to consider for some L, you can use prefix sums to calculate the answer for all these and find the min answer for some L / if it is not possible.

lems9520138 (commented):
Nice job, just some stuff to fix :)

lems9520138 (changes requested):

solutions/silver/cf-1244E.mdx
4nav: we're basically shifting the window L,L+d

solutions/silver/cf-1244E.mdx
4nav: writing the math out? sure will do rn

solutions/silver/cf-1244E.mdx
```diff
@@ -13,24 +13,50 @@ Instead of trying to construct a final array, fix a target difference $d$ and ch
 
 If all elements can be moved into some interval of length $d$, then it can be shown that all larger values of $d$ work. Hence, the condition is monotonic and we can binary search on $d$.
 
-For a fixed $d$, we try to fit all elements into $[L, L + d]$:
+For a fixed $d$, we try to fit all elements into $[L, L + d]$, where $L$ is an arbitrary starting point for the interval:
 - increase elements below $L$
 - decrease elements above $L + d$
 
-The total cost can be computed using prefix sums.
+The total cost (number of unit increments/decrements needed to move all elements into [L,L+d]) can be computed using prefix sums.
```
lems9520138: The total cost, the number of increments/decrements needed to translate all elements into an interval [L,L+d] can be computed using prefix sums.

solutions/silver/cf-1244E.mdx
```diff
@@ -13,24 +13,50 @@ Instead of trying to construct a final array, fix a target difference $d$ and ch
...
 
-### Choosing $L$
+### Choosing L
 
-Let $A$ be the number of elements with $a_i < L$, and $B$ be the number of elements with $a_i > L + d$.
+We consider an interval [L, L + d] of fixed length d, where L is the left endpoint.
```
lems9520138: We consider an interval [L, L + d] of fixed length d, where L can be any number, and represents the left endpoint.

solutions/silver/cf-1244E.mdx
```diff
@@ -13,24 +13,50 @@ Instead of trying to construct a final array, fix a target difference $d$ and ch
...
 
-Shifting $[L, L + d]$ (without crossing any $a_i$) changes the cost by $A - B$.
-Hence, we can move the interval to improve or preserve the cost until an endpoint hits some $a_i$.
+If we move this interval slightly to the right without crossing any a[i], then:
+- elements left of L become more expensive to increase
+- elements right of L + d become cheaper to decrease
 
-If the left endpoint hits $a_i$, then $L = a_i$.
-If the right endpoint hits $a_i$, then $L + d = a_i \Rightarrow L = a_i - d$.
+So we can always keep moving in the direction that does not increase cost until we hit some a[i].
```
lems9520138: By "hit some a[i]", I assume you mean that the right endpoint touches some a[i], but could you maybe clarify that.

So we can keep shifting the until the right or left endpoint hits some a[i]?? (Idk exactly what u mean)

solutions/silver/cf-1244E.mdx
```diff
@@ -13,24 +13,50 @@ Instead of trying to construct a final array, fix a target difference $d$ and ch
...
 
-### Choosing $L$
+### Choosing L
 
-Let $A$ be the number of elements with $a_i < L$, and $B$ be the number of elements with $a_i > L + d$.
+We consider an interval [L, L + d] of fixed length d, where L is the left endpoint.
 
-Shifting $[L, L + d]$ (without crossing any $a_i$) changes the cost by $A - B$.
-Hence, we can move the interval to improve or preserve the cost until an endpoint hits some $a_i$.
+If we move this interval slightly to the right without crossing any a[i], then:
```
lems9520138: If we move this interval to the right without crossing any a[i], then:

solutions/silver/cf-1244E.mdx
```diff
@@ -13,24 +13,50 @@ Instead of trying to construct a final array, fix a target difference $d$ and ch
```
lems9520138, line 13: Overall, in some sections you forgot to use latex.

lems9520138 (changes requested):

solutions/silver/cf-1244E.mdx
4nav: yeah thats exactly what i meant

solutions/silver/cf-1244E.mdx
4nav: alr thanks!

4nav: done btw

4nav:
hello

solutions/silver/cf-1244E.mdx
```diff
@@ -11,26 +11,26 @@ author: Abhinav Singh
 
 Instead of trying to directly construct a final array, we fix the maximum difference $d$ and check if it is achievable in under $k$ operations. If all elements can be moved into some interval of length $d$, then it can be shown that all larger values of $d$ work. Hence, the condition is monotonic and we can binary search on $d$.
```
lems9520138: Instead of trying to directly construct a final array, we fix the maximum difference $d$ and check if it is achievable in under $k$ operations. If all elements can be moved into some interval of length $d$, then it is clear that all larger values of $d$ will work. Hence, the condition is monotonic and we can binary search on $d$.

solutions/silver/cf-1244E.mdx
```diff
@@ -11,26 +11,26 @@ author: Abhinav Singh
...
 
-For an array to have a maximum difference of $d$, all elements must be in the range $[L, L + d]$, where 
-$L$ is an arbitrary starting point. This means that our operations must move values of 
+For an array to have a maximum difference of $d$, all elements must be in the range $[L, L + d]$, where
+$L$ is an arbitrary starting point. This means that our operations must move values of
```
lems9520138: may be wrong but idk if arbitrary starting point is the correct word here I feel like
where $L$ is an arbitrary value representing the start of the interval.

maybe im tripping

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+<Spoiler title="Proof">
+
+The easiest way to show this is by using an exchange argument. Essentially,
+we show that any optimal solution can be transformed into the greedy construction
+without impacting the quality of the final solution.
+
+Suppose neither $L$ nor $R$ (which is equal to $L+d$) equals any value in $a$. Let $p$ be the number of elements below $L$ and $q$ the number above $R$. Shifting the window right by 1 changes the total cost by $p - q$ (elements below $L$ each need one more step, and elements above $R$ each need one fewer). Shifting left changes the cost by $q - p$ through similar reasoning.
+
+- If $p \neq q$: shifting in the cheaper direction strictly reduces total cost by $|p - q|$ per step, so the current placement is not optimal.
```
lems9520138:
```suggestion
- If $p \neq q$: shifting in the cheaper direction (the direction with more elements) strictly reduces total cost by $|p - q|$ per step, so the current placement is not optimal.
```

solutions/silver/cf-1244E.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+$$
+\sum_{a_i < L} (L - a_i) + \sum_{a_i > R} (a_i - R)
+$$
+
+In other words, we sum the costs of moving values of $a_i < L$ to equal $L$,
+and the costs for moving values of $a_i > R$ to equal $R$.
+
+If we sort our array $a_i$, then a prefix and suffix of values will need to be
+moved within range. We use binary search to find the prefix and suffix of values
```
lems9520138, line 60: correct me if Im wrong, but you wouldn't need binary search to find prefix values, just suffix values.

We use binary search to find the suffix of values that need to be adjusted, and prefix sums to sum up the relevant values of $a_i$.

eysbutno: no since a_i can also be equal to $L+d$

solutions/silver/cf-1244E.mdx
eysbutno: replaced by phrasing it as any integer since that's more accurate than 'arbitrary starting point'

solutions/silver/cf-1244E.mdx
lems9520138: oh yeah right mb

lems9520138 (approved):

PR #6172, opened 2026-05-12
Add 2 problems to additional practice for silver

sc3dev:

- [x] I have tested my code.
- [x] I have added [Request](https://github.com/sc3dev/usaco-guide/pull/new/patch1)my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).

Would just use the suggest a problem form but it seems to be broken for me
If you don't like one of the problems you can just delete it from the pull request

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4068,9 +4068,9 @@ __metadata:
   linkType: hard
 
 "@protobufjs/utf8@npm:^1.1.0":
-  version: 1.1.0
-  resolution: "@protobufjs/utf8@npm:1.1.0"
-  checksum: 10c0/a3fe31fe3fa29aa3349e2e04ee13dc170cc6af7c23d92ad49e3eeaf79b9766264544d3da824dba93b7855bd6a2982fb40032ef40693da98a136d835752beb487
+  version: 1.1.1
+  resolution: "@protobufjs/utf8@npm:1.1.1"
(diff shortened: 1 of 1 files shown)


bqi343:
> Would just use the suggest a problem form but it seems to be broken for me

Looking into this!

content/3_Silver/Conclusion.problems.json
```diff
@@ -98,6 +98,30 @@
         "kind": "internal"
       }
     },
+    {
+      "uniqueId": "cf-1526C2",
+      "name": "Potions (Hard Version)",
+      "url": "https://codeforces.com/contest/1526/problem/C2",
+      "source": "CF",
+      "difficulty": "Easy",
```
eysbutno: i think this should be normal because the greedy idea isn't so intuitive if you haven't seen it before

also it shares similarities to https://usaco.org/index.php?page=viewproblem2&cpid=1451 which is a gold problem, tho the gold variant is obv more complicated

eysbutno (commented):
thanks for the problem suggestions! I think both are solid

content/3_Silver/Conclusion.problems.json
sc3dev: I got it on my own pretty quickly and haven't seen anything like it but I'll change it

sc3dev:
crap did I accidentally close this? sorry I'm not good with github I'll open a new pull

PR #6174, opened 2026-05-14
[Editorial] Maximum Manhattan Distances

brebenelmihnea:

- [x] I have tested my code.


brebenelmihnea:
https://cses.fi/problemset/result/17163969/

image

bqi343:
Maybe https://usaco.guide/silver/silver-conclusion would be a better place for this? sort of odd to put it in geo primitives when you don't actually need to know anything from that module to solve this problem.

Same for the other PR.

solutions/platinum/cses-3410.mdx
```diff
@@ -0,0 +1,109 @@
...
+id: cses-3410
+source: CSES
+title: Maximum Manhattan Distances
+author: Mihnea Brebenel
+---
+
+## Solution
+
+Let $A(x_A,y_A)$ and $B(x_B,y_B)$ be two points, then the manhattan distance between the two of them is $d = |x_B-x_a|+|y_B - y_A|$. Usually, it's a good idea to group the variables that depend of one element, so we use the definition
+of the absolute value:
```
eysbutno:
```suggestion
For two points $A=(x_A, y_A)$ and $B=(x_B, y_B)$, the Manhattan distance between the two is $d = |x_B-x_a|+|y_B - y_A|$. It's often a good idea to eliminate the absolute values, so we try that using the definition of an absolute value:
```

not exactly sure what you meant by "good idea to group the varialbes that depend of one element", I assume the general meaning is to just casework out the absolute values

solutions/platinum/cses-3410.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+|y_B-y_A| = \left\{
+\begin{array}{ll}
+y_B-y_A & y_B - y_A \ge 0\\
+-(y_B-y_A) & y_B-yA <0
+\end{array}
+\right.
+$$
+
+Now, we distinguish four cases($2 *2$ from the absolute value explicitations) for the distance equation:
```
eysbutno:
```suggestion
Now, we distinguish four cases for the distance value, based on the cases for $x$ and $y$:
```

solutions/platinum/cses-3410.mdx
```diff
@@ -0,0 +1,109 @@
...
+x_B+y_B-(x_A+y_A) \\
+x_B-y_B-(x_A-y_A) \\
+-(x_B-y_B)+x_A-y_A \\
+-(x_B+y_B)+x_A+y_A
+\end{array}
+\right.
+$$
+
+As we can see, for each point the sum and the difference of the coordinates matters. We'll keep track of the maximum and minimum of the sum and difference.
+When a new point comes up, we updates the values. The maximum of the values: $\text{maxsum}-\text{minsum}$ and $\text{maxdiff}-\text{mindiff}$ is the answer of the four cases.
```
eysbutno: i think it would be good to expand on this section

e.g. something like

> As seen above, the value of $d$ in each case can be rewritten as the difference between the maximum and minimum values of either the sum or difference of each point's coordinates. Thus, to compute our answer, we store the maximum and minimum values of $x-y$ and $x+y$ across each prefix of points, and use these values to calculate our answer.

if you want to explain the correctness a bit more (not necessary but could be nice), just mention that if a pair of points $A$ and $B$'s distance $d$ is considered in the wrong case under our formulas, then evaluating it with the correct case would result in a more optimal answer.

solutions/platinum/cses-3410.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+int main() {
+
+	int n;
+	cin >> n;
+
+	int x, y;
+	cin >> x >> y;
+
+	long long minsum, maxsum;
```
eysbutno: `snake_case` would be more readable here imo

eysbutno (commented):
overall good work! I agree with benq in that silver conclusion is probably a better place for this, maybe as a normal problem?

solutions/platinum/cses-3410.mdx
brebenelmihnea: done

solutions/platinum/cses-3410.mdx
brebenelmihnea: done

solutions/platinum/cses-3410.mdx
brebenelmihnea: ok  i rephrased it

solutions/platinum/cses-3410.mdx
brebenelmihnea: ok

content/3_Silver/Conclusion.problems.json
```diff
@@ -98,6 +98,18 @@
...
       }
     },
+    {
+      "uniqueId": "cses-3410",
+      "name": "Maximum Manhattan Distances",
+      "url": "https://cses.fi/problemset/task/3410/",
+      "source": "CSES",
+      "difficulty": "Normal",
+      "isStarred": false,
+      "tags": [],
```
eysbutno: prob just tag it as geometry

eysbutno (approved):

content/3_Silver/Conclusion.problems.json
brebenelmihnea: ok

guodavid0418 (approved):
nice!

PR #6184, opened 2026-05-17
Revert "Bump uuid from 11.1.0 to 14.0.0"

ryanchou-dev:
Oops, reverting for breaking change on 14.0.0

Reverts cpinitiative/usaco-guide#6161


ryanchou-dev:
uuid already seems to be 14.0.0 on package.json, so not sure why this file is currently out of sync 👀

PR #6214, opened 2026-06-02
Hot reload

xyzqm:
DISCLAIMER: all code in this PR was generated by Claude, but I’ve tested it by authoring a solution with  images and updating .problems.json and it appears to work just fine.

This PR adds the option to run `yarn dev` with hot reload via the new command `yarn dev:watch`.


xyzqm:
For some reason `usaco-divisions.json` seems to be constantly getting auto-modified, not sure what's up with that...

emhgit:
> For some reason `usaco-divisions.json` seems to be constantly getting auto-modified, not sure what's up with that...

this just occurs whenever the DB is rebuilt, since the script that builds the DB autogenerates @usaco-divisions.json

bqi343:
> this just occurs whenever the DB is rebuilt, since the script that builds the DB autogenerates @usaco-divisions.json

Is it supposed to change though? Given that this PR didn't modify any of the problems.

emhgit:
> > this just occurs whenever the DB is rebuilt, since the script that builds the DB autogenerates @usaco-divisions.json
> 
> Is it supposed to change though? Given that this PR didn't modify any of the problems.

no, the file does not change. the script deletes the file and then regenerates it with the same data, but deleting the file makes git think it’s different. after adding it to the staging area, git will recognize that it’s the same (i.e. the file is not actually modified)

bqi343:
https://github.com/cpinitiative/usaco-guide/pull/6214/changes is showing a diff though

emhgit:
> https://github.com/cpinitiative/usaco-guide/pull/6214/changes is showing a diff though

i can’t access my PC rn so i cannot point out exactly why, but i'm guessing its because this PR branch and the main branch were out of sync

emhgit:
@bqi343 https://github.com/cpinitiative/usaco-guide/pull/6237

xyzqm:
I've restored the `usaco-divisions` file, how does everything else look?

emhgit (approved):

PR #6222, opened 2026-06-03
Fix incorrect KMP solution for Baltic OI 2019 Necklace

### LINKED ISSUE #6140: Incorrect code on https://usaco.guide/problems/baltic-oi-2019necklace/solution
Solution 2 using kmp is incorrect. It gets the wrong answer on the input "a a" even though the first code is correct. Here is correct code that uses kmp and passes:
```cpp
#include <bits/stdc++.h>
using namespace std;
#define a first
#define b second
vector<int> pi(const string& s) {
	int n = (int)s.size();
	vector<int> pi_s(n);
	for (int i = 1, j = 0; i < n; i++) {
		while (j > 0 && s[j] != s[i]) {
			j = pi_s[j - 1];
		}
		if (s[i] == s[j]) {
			j++;
		}
		pi_s[i] = j;
	}
	return pi_s;
}
vector<int> calc(string s, string t) {
	vector<int> ret(s.size());
	string cur = t + "#" + s;
	vector<int> p = pi(cur);
	for (int i = 0; i < s.size(); i++) ret[i] = p[i + t.size() + 1];
	return ret;
}
pair<int, pair<int, int>> solve(string s, string t) {
	t += ".";
	pair<int, pair<int, int>> ans = {0, {0, 0}};
	for (int i = 0; i < t.size() - 1; i++) {
		vector<int> left = calc(s, t);
		reverse(s.begin(), s.end());
		reverse(t.begin(), t.end());
		vector<int> right = calc(s, t);
		reverse(s.begin(), s.end());
		reverse(t.begin(), t.end());
		reverse(right.begin(), right.end());
		for (int j = 0; j <= s.size(); j++) {
			int l = j == 0 ? 0 : left[j - 1];
			int r = j == s.size() ? 0 : right[j];
			ans = max(ans,
					  {l + r,
					   {j - l, (2 * (t.size() - 1) - r + i) % (t.size() - 1)}});
		}
		rotate(t.begin(), t.begin() + 1, t.end());
	}
	return ans;
}
int main() {
	ios::sync_with_stdio(false);
	cin.tie(nullptr);
	string s, t;
	cin >> s >> t;
	pair<int, pair<int, int>> ans = solve(s, t);
	reverse(t.begin(), t.end());
	pair<int, pair<int, int>> cur = solve(s, t);
	cur.b.b = t.size() - cur.b.b - cur.a;
	ans = max(ans, cur);
	cout << ans.a << "\n";
	if (ans.a) cout << ans.b.a << " " << ans.b.b << "\n";
}
```

arnavkewalram:
## Summary

- Replaces the buggy Solution 2 (KMP) code which produced wrong answers on inputs like `"a a"` (reported in #6140).
- The previous implementation had incorrect index arithmetic in the KMP lookups; the new implementation iterates over all cut points of `T` (rotations) and uses the KMP prefix function via `calc(s, t) = pi(t + "#" + s)` to find the longest matching suffix/prefix at each split point of `S`.
- Updates the explanation in the solution to accurately describe the new approach.

## Test plan

- [ ] Verify `"a a"` gives output `1` (the bug case from the issue)
- [ ] Verify the sample input from the problem statement produces the correct output
- [ ] Check that Solution 1 (Magic DP) is untouched

Fixes #6140

### DIFF
--- a/solutions/advanced/baltic-19-necklace.mdx
+++ b/solutions/advanced/baltic-19-necklace.mdx
@@ -188,10 +188,10 @@ pair<int, pair<int, int>> solve(string s, string t) {
 		for (int j = 0; j <= (int)s.size(); j++) {
 			int l = j == 0 ? 0 : left[j - 1];
 			int r = j == (int)s.size() ? 0 : right[j];
-			ans = max(ans, {l + r,
-			                {j - l,
-			                 (int)(2 * (t.size() - 1) - r + i) %
-			                     (int)(t.size() - 1)}});
+			ans =
+			    max(ans,
+			        {l + r,
+			         {j - l, (int)(2 * (t.size() - 1) - r + i) % (int)(t.size() - 1)}});
 		}
 		rotate(t.begin(), t.begin() + 1, t.end());
 	}
--- a/solutions/advanced/baltic-19-necklace.mdx
+++ b/solutions/advanced/baltic-19-necklace.mdx
@@ -134,56 +134,82 @@ such that:
 - $B$ is a **prefix** of $S[i + 1 : N - 1]$ and a **suffix** of $T[0 : j]$.
 - $|A| + |B|$ is maximal.
 
-We can thus test each $i$ to find the best $j$ for that $i$, and then take the
-best overall result.
+We iterate over each cut point $j$ of $T$ (i.e., all $M$ rotations of $T$). For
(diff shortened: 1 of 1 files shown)


arnavkewalram:
Closing — submitted prematurely, will reopen when ready.

PR #6237, opened 2026-06-05
Modify Non-Deterministic Output of @usaco-divisions.json File

emhgit:

- [x] I have tested my code.

Apparently, the `readdir(..., { recursive: true })` function of the `fs/promises` library does not consistently return files in the same order for different machines. So, the `problems` table of the final database was dependent on the ordering of the files processed. 

Sorted the file array returned by the function to guarantee alphabetical ordering of files processed; added conditions to prioritize insertion of data of problems found in modules over problem data found in @extraProblems.json


bqi343 (approved):
is it possible for a problem to appear in both extraproblems and a module? anyway, seems fine

emhgit:
> is it possible for a problem to appear in both extraproblems and a module? anyway, seems fine

the current data schema would not allow that; however, there is a `moduleProblemsList` (or something similar) table in the db that allows queries for all the modules a problem is in

PR #6244, opened 2026-06-08
Patch 2

lems9520138:

- [x ] I have tested my code.


solutions/silver/cf-2044G2.mdx
```diff
@@ -0,0 +1,98 @@
+---
+id: cf-2044G2
+source: CF
+title: Medium Demon Problem (hard version)
+author: Arnav Gokhale
+---
+
+[Official Analysis](https://codeforces.com/blog/entry/137306)
```
eysbutno: oh right i forgot to mention this for the other PR, but mention sol is C++

solutions/silver/cf-2044G2.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Unlike the easy version, spiders can now hold multiple plushies. Model each connected component as a cycle with rooted trees attached to its nodes, where each tree is rooted at a node on the cycle. So, how long will it take for some spider $i$ in a tree to give away all of it's plushies? This is equal to the subtree size of $i$, since spider $i$ will give away exactly one plushie every year, and all the plushies in the subtree of $i$ will must pass through $i$.
+
+By constructing a seperate adjacency list, we can run a DFS on each of these trees, allowing us to compute all the subtree sizes, ignoring the root (which is part of the cycle). The final answer will be the maximum subtree size among all such nodes.
```
eysbutno:
```suggestion
First, note the structure of the graph. Every node has a singular outgoing edge, making it a functional graph. As a result, the graph consists of several connected components, each containing exactly one cycle, with all the nodes either in the cycle or feeding into it.

Notice that if a node is not in the cycle, the process will not become stable until that node's plushie reaches the cycle. Consider the nodes along a path leading into a cycle. On each day, the spiders will pass one of their plushies (if they have one) to $r_i$. However, some of the spiders that send a plushie will not receive one in return, causing the system to have changed numbers. Thus, like the easy version, once every plushie reaches a cycle, each spider will both receive and give a plushie, keeping the system stable.

Unlike the easy version, spiders can now hold multiple plushies. Model each connected component as a cycle with rooted trees attached to its nodes, where each tree is rooted at a node on the cycle. So, how long will it take for some spider $i$ in a tree to give away all of its plushies? This is equal to the subtree size of $i$, since spider $i$ will give away exactly one plushie every year, and all the plushies in the subtree of $i$ must pass through $i$.

By constructing a separate adjacency list, we can run a DFS on each of these trees, allowing us to compute all the subtree sizes, ignoring the root (which is part of the cycle). The final answer will be the maximum subtree size among all such nodes.
```

while grammarly sometimes pmo, it's good for finding simple mistakes

eysbutno: also one thing that isn't super necessary but would be good to mention: we can think of it as two bounding factors: the time it'd take for all the plushies to 'flow' through a cycle 'entrance' $i$, and the time it takes to reach. since subtree is >= max depth, we just take subtree size

solutions/silver/cf-2044G2.mdx
```diff
@@ -0,0 +1,98 @@
...
+	vector<int> a(n);
+	vector<vector<int>> c(n);
+	for (int i = 0; i < n; i++) {
+		cin >> a[i];
+		a[i]--;
+		c[a[i]].push_back(i);
+	}
+
+	vector<int> b(n);
+	vector<bool> v(n);
```
eysbutno: same thoughts apply from prev review; please rename the vars to something more understandable/symbolic

solutions/silver/cf-2044G2.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Unlike the easy version, spiders can now hold multiple plushies. Model each connected component as a cycle with rooted trees attached to its nodes, where each tree is rooted at a node on the cycle. So, how long will it take for some spider $i$ in a tree to give away all of its plushies? This is equal to the subtree size of $i$, since spider $i$ will give away exactly one plushie every year, and all the plushies in the subtree of $i$ must pass through $i$.
+
+By constructing a separate adjacency list, we can run a DFS on each of these trees, allowing us to compute all the subtree sizes, ignoring the root (which is part of the cycle). The final answer will be the maximum subtree size among all such nodes.
```
johnzhou721: ... plus 2; explain why.

solutions/silver/cf-2044G2.mdx
```diff
@@ -0,0 +1,98 @@
...
+author: Arnav Gokhale
+---
+
+[Official Analysis](https://codeforces.com/blog/entry/137306)
+
+## Explanation
+
+First, note the structure of the graph. Every node has a singular outgoing edge, making it a functional graph. As a result, the graph consists of several connected components, each containing exactly one cycle, with all the nodes either in the cycle or feeding into it.
+
+Notice that if a node is not in the cycle, the process will not become stable until that node's plushie reaches the cycle. Consider the nodes along a path leading into a cycle. On each day, the spiders will pass one of their plushies (if they have one) to $r_i$. However, some of the spiders that send a plushie will not receive one in return, causing the system to have changed numbers. Thus, like the easy version, once every plushie reaches a cycle, each spider will both receive and give a plushie, keeping the system stable.
```
johnzhou721, line 14: Same comment from easy version on this paragraph

solutions/silver/cf-2044G2.mdx
johnzhou721, line 1: Perhaps link to easy version and deduplicate some stuff?  Or you can keep the current context and just link to easy version since it's already used in discussion.

johnzhou721 (changes requested):
Overall just a couple nits

solutions/silver/cf-2044G2.mdx
johnzhou721: Also a critical issue: clarify which subtrees.  They aren't the immediate trees rooted by the point on the cycle -- it's actually the *sub*tree that's one level deeper than that; it matters because you're taking maximum for each of those *sub*trees, not adding them together for the subnode count of a single tree centered on the loop and then maxing that.

Image

johnzhou721 (changes requested):
One more functional issue

solutions/silver/cf-2044G2.mdx
lems9520138: This is covered in the last paragraph.

johnzhou721 (approved):

eysbutno (approved):

PR #6267, opened 2026-06-12
Add note for Sabotage

johnzhou721:
Yes, another alternate solution.

- [x] I have tested my code.


solutions/silver/usaco-419.mdx
```diff
@@ -107,3 +109,78 @@ print(f"{l:.3f}", file=open("sabotage.out", "w"))
 
 </PySection>
 </LanguageSection>
+
+## Solution 2 (Binary Search, Maximum Sum Subsequence)
```
xyzqm: Do you mean "Maximum Sum Subarray"?

solutions/silver/usaco-419.mdx
```diff
@@ -107,3 +109,78 @@ print(f"{l:.3f}", file=open("sabotage.out", "w"))
...
+Note that we now must find if a specific average is possible in $\mathcal O(n)$ time.
+However, the average is dependent on the number of stations that are not unplugged, which makes this difficult in $\mathcal O(n)$.
+But since we already know the target average, we can temporarily subtract each number in the list by the target average;
+now, the sign of the sum of a set of stations will tell us whether their average is equal to or below the target ($<=0$) or above the target ($>0$).
+
+To have the best chance at achieving the average, the sum of the remaining elements in the temporary array after the sabotage should be at a minimum;
+this means the sum of the sabotaged elements, a continuous range in $[2, n-1]$ should be at a maximum.
+
+The maximum continuous range in an list could be found using a standard DP algorithm:
+Let $K_{2,3,\dots,N-1}$ be the maximum sum of a range ending at index $i$.
```
xyzqm: I think you should just write $K_i$ here rather than $K_{2,3,\dots,N-1}$

xyzqm (dismissed):
I'm not sure that this is substantially different from the first solution, but I suppose the perspective shift from prefix sums to a DP could be educational. Regardless, I do like how clear your writing is!

johnzhou721:
@xyzqm I made the requested changes; would appreciate another review.

I introduced this trick because I feel like dealing with average probelms by binary search + subtracting is creative and also because this is the official solution.

johnzhou721:
@eysbutno @guodavid0418 @lems9520138 Please review when you've got time — thanks!

eysbutno:
IMO, this may be better suited as a note in the original solution? the reason is:
- the code is essentially identical to the first solution
- mathematically they're equivalent, it's just a nice reframing of it in a slightly more intuitive way

johnzhou721:
@eysbutno Since DP isn't that important in silver anyways, I added a tiny note about the alternative instead.  Take a look!

johnzhou721:
@xyzqm

solutions/silver/usaco-419.mdx
```diff
@@ -21,6 +21,12 @@ $$
 $$
 For a fixed $j$, we should try to minimize the left side expression. So we can iterate $j$ from $2$ to $N - 1$ and keep a running minimum of the left side of the inequality. If the condition is satisfied, we can achieve an average $\le X$.
 
+<Info title="Note">
+
+If we pre-subtract each element in the inequality by the average $X$ we look for from the milk production array and compute its prefix sum, then we can reformulate the final inequality as a maximum subarray sum problem and use DP instead.
+
+</Info>
```
eysbutno: 1. you can leave title blank unless you have a more specific note you want to attach
2. this is kind of vague? below is how i'd personally write it, obv may not be the best way but just my thoughts

> An alternative approach initially subtracts the average $X$ we are computing from each value in the array. In this new array, if the sum of elements is less than or equal to zero, then that means that Farmer Nhoj can achieve an average $\leq X$. Thus, we aim to find the subarray with the largest sum to achieve the lowest total sum of elements, which can be done using greedy methods.

johnzhou721: Sorry, got too caught up in the explicit math while I was doing the refactor, and then just thought that none of it was worth explaining.  But I like your intuition balance.  Let me rewrite this myself tomorrow.

johnzhou721:
```suggestion
<Info>

Alternatively, one may subtract the average $X$ being verified from each value in $M$, resulting in an array where the average of a set in $M$ is smaller than or equal to $X$ if and only if the sum of the corresponding set in the modified array is less than or equal to $0$. This means that the maximum subarray sum of the modified array except the first and last elements is desired to minimize the average of the modified sum of the remaining elements, obtained using a greedy method.

To see how this is equivalent to the main approach, note that that the sum of $[l, r]$ in the modified array is the sum in the original array minus $X\cdot (r-l+1)$.  Using this to rewrite the prefix sum array and thus the last equation in the main solution will result the same optimization goal.

</Info>
```

johnzhou721:
@eysbutno Ready for a rereview.

I added a note on how the 2 solutions connect mathematically since it is not immediately obvious to a silver level participant (me).

eysbutno (approved):

guodavid0418 (approved):

PR #6271, opened 2026-06-12
Add "Tree Queries" as a focus problem for the HLD module

xyzqm:


content/5_Plat/HLD.mdx
```diff
@@ -347,6 +347,149 @@ int main() {
 </CPPSection>
 </LanguageSection>
 
+## Tree Queries
+
+<FocusProblem problem="tree-queries" />
```
bqi343, line 352: why a focus problem, wouldn't it be better in the problems section?

if you want an example of HLD without segment tree what about LCA

xyzqm: Hm, I just wanted to capture the idea of solve it on a line -> then you can solve it on a tree with HLD. Do you know any other problems that might be suitable for illustrating this?

bqi343: isn't this most HLD problems

xyzqm: Sure, but just browsing through the problems section it seems like most of the problems only use HLD for the more specific purpose of "vertex/path updates and path compose", so I just wanted to showcase a problem outside of this subcategory.

bqi343: One of the resources mentions path update and subtree queries.

This problem is path update plus entire tree queries so it's related ig.

PR #6272, opened 2026-06-12
Move extra problem in Flow_LB to the correct table

### LINKED ISSUE #6255: Contact Form Submission - Mistake (Advanced - Flow with Lower Bounds)

**URL**: https://usaco.guide/adv/flow-lb
**Module**: Advanced - Flow with Lower Bounds
**Topic**: Mistake
**Message**: 
In Flow with Lower Bounds module, there is a problem bar at the start of the page is this supposed to be there like that, or should it be included in the problems section below?

xyzqm:
Fixes #6255

### DIFF
--- a/content/6_Advanced/Flow_LB.mdx
+++ b/content/6_Advanced/Flow_LB.mdx
@@ -8,8 +8,6 @@ description: 'Adding lower bounds to flow edge constraints.'
 frequency: 0
 ---
 
-<Problems problems="hungry" />
-
 ## Tutorial
 
 <Resources>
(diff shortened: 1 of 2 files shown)


bqi343:
actually the positioning was intentional (first give an example of a problem needing flow with lower bounds, then introduce the topic)

xyzqm:
Should it be a focus problem then, maybe with an in-module solution too?

bqi343:
At least one problem from the module should be a focus problem with an in-module solution.

PR #6279, opened 2026-06-14
fix: reclassify Bovine Acrobatics from Easy to Normal in Silver Greedy Sorting

xneolxnoel:
## Summary
- Reclassify "Bovine Acrobatics" (usaco-1350) from Easy to Normal
- Move it after Rest Stops to sit as the first Normal problem

## Rationale
The solution not only requires greedy thinking, but also requires the
implementation of grouping / queue (deque-based approach), making Normal
a more appropriate difficulty.

## Test plan
- [x] Verified Bovine Acrobatics appears under Normal difficulty on local dev server
- [x] Verified it's positioned after Rest Stops and before Berry Picking


eysbutno (approved):
sure

lems9520138 (approved):
👍

johnzhou721 (approved):
Agreed.  This took me a LONG LONG time back then and I already forgot how to solve it.

USACO power surge :sob: in 5 more years this will be considered Easy again

Issue #2949, opened 2022-07-20
Mixed Problem Sets


thecodingwizard:
Will mixed problem sets be helpful / is it within the scope of the USACO Guide?

## Motivation

Currently, the USACO Guide has a reasonable number of problems per topic, but no mixed problem sets. Therefore, people might say "I went through the USACO Guide, did all the problems, and still can't promote." In this situation, will providing very large mixed problem sets for practice help?

The goal would be to make the USACO Guide an actual one-stop-shop. People go through the USACO Guide to learn all the topics, and then they do practice problems until they promote. Currently I think the Guide does a good job of the former, but after you've gone through the Guide, if you're still unable to promote you basically have to figure out how to get problems yourself (especially once you've exhausted problems from the Guide, which honestly aren't that many?).

## Getting Mixed Problemsets

Basically for every problem we'll need (not all are necessary):

- Basic problem info (name, URL, etc)
- Division + difficulty relative to division
- Tags
- Solutions
- Quality (is this actually a valid metric?)

I don't particularly know how to get this information, especially for a lot of problems. I also don't know how to ensure quality control of the problems (do we even need to, or do we just assume that all problems are worth doing unless proven otherwise?).

We could either crowdsource from users (quality control questionable), or find problems ourselves (idk how long this would take).

Really old USACO problems might be a good source? We have the problem statements, test data, and solutions, and we can relatively easily create an online judge to make all those problems submittable. I don't know how good the quality is though.

## Using Mixed Problemsets

Some random ideas:

- After every few modules, list a bunch of problems from the preceding modules
- At the end of every division ("additional practice"), put a bunch of problems (we already have some, but like, a lot more)
- On the problems page, make a better "practice problemset generator" (filter by topic, difficulty, remove problems you've already solved, etc)
- Make a "mock practice contest generator"

SansPapyrus683:
i can probably make a CLI for that
integrating it into the website? yeah that's beyond me
all for the idea though

caoash:
I think crowdsourcing is best, w/ maybe some rating system on problem quality. If difficulty doesn't matter, we can also just scrape them from somewhere

> Really old USACO problems might be a good source
maybe training pages as well?

bqi343:
Difficulty rating is helpful, but I think a quality rating is not. Doing the problems in increasing order of difficulty makes sense, but doing the problems in increasing or decreasing order of quality doesn't. If we think a problem is of questionable quality then we shouldn't include it.

[...]

PR #6282, opened 2026-06-15
Implement mixed problem sets for bronze

500EJ:
This is a basic implementation of mixed problem sets for bronze following my proposal in issue #2949. It is enabled with "Progression Mode" in settings (disabled by default, also maybe someone can come up with a better name). Essentially, it incorporates spaced repetition of previous modules by moving unstarred problems to be reviewed in a later module. If this looks good, it might also make sense to add some info about this feature to the General section.

Edit: Also someone with write access might want to link it to the issue in the sidebar, I don't want to put "closes" for now.

- [x] I have tested my code.

### DIFF
--- a/content/2_Bronze/Ad_Hoc.mdx
+++ b/content/2_Bronze/Ad_Hoc.mdx
@@ -127,7 +127,7 @@ print(res)
 
 ## Problems
 
-<Problems problems="general" />
+<Problems problems="general" progressionSet />
 
 ## Quiz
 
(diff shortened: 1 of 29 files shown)


eysbutno:
I'm not sure if directly taking unstarred problems and putting them in review mode is the best way? my main concerns are:
- this strongly biases module problems towards normal/hard problems, which obv makes gradually working thru the guide harder
- imo it's often worthwhile to solve non-starred problems when learning a topic, and the current implementation takes away all of those options

I think the idea of progression/review sets is good, but would likely be better if (a) we picked out newer usaco problems for this or (b) chose which module problems should be put in review by hand

PR #6326, opened 2026-06-27
docs: correct Sleeping in Class complexity

### LINKED ISSUE #6324: Contact Form Submission - Mistake (Solution: USACO Bronze 2022 February - Sleeping in Class)

**URL**: https://usaco.guide/problems/usaco-1203-sleeping-in-class/solution
**Module**: Solution: USACO Bronze 2022 February - Sleeping in Class
**Topic**: Mistake
**Message**: 
In the implementation you are claiming that the time complexity is O(N*sum) which is not only wrong, but also misleading since a pure O(N*sum) would NOT pass, or still, SHOULD NOT, pass for the problem costraints which are N <= 1e5 and sum 1e6, which is 1e11 ( completely out ), why it passes then ? well, it passes because as you said, you are taking only divisors of the total sum, so the complexity it's NOT N*sum but N * sqrt(sum) which is a completely different complexity, thank you for your time, have a nice day!

Saithej2k:
Summary
- Correct the Sleeping in Class time complexity explanation to account for only divisor candidates doing a full scan.

Testing
- `git diff --check`

Closes #6324

### DIFF
--- a/solutions/bronze/usaco-1203.mdx
+++ b/solutions/bronze/usaco-1203.mdx
@@ -20,12 +20,12 @@ $\texttt{sum}/\texttt{num\_hours}$. Since every modification reduces the length
 of the log by 1, the number of modifications Elsie makes will be
 $N-\texttt{sum}/\texttt{num\_hours}$.
 
-Since $N$ and $\texttt{sum}$ are constant, to maximize the above expression
+Since $N$ and $\texttt{sum}$ are constant, to minimize the above expression
 our only option is to minimize $\texttt{num\_hours}$.
 
-In order to do that, we iterate over possible values of $\texttt{num\_hours}$
-from smallest to largest and check if it's possible to reduce the log to an
-array with all elements equal to it.
+In order to do that, we iterate over divisor values of
+$\texttt{num\_hours}$ from smallest to largest and check if it's possible to
+reduce the log to an array with all elements equal to it.
 As soon as we hit a valid element, we print out the number of modifications needed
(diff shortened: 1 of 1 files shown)


johnzhou721 (changes requested):
The code does not use the optimized sqrt approach so that would have to be changed as well.

Plus I think you can get away with just listing the final time complexity instead of defining d explicitly, as finding the prime factors themselves already require sqrt operations exactly.

Very Good catch though!!!  This would be a good performance improvement that is educational to show.

4nav:
is it just me or is the text just not rendering

johnzhou721:
It's rendering for me

johnzhou721:
There's many more changes being made, most of which I trust is correct; but I'll review this a bit later.  Sorry.

johnzhou721 (approved):
Make sure you test the codes through usaco.org.  But the code looks great (and are all relevant to my original requests).  I really liked the valid() abstraction, although I'd like another pair of eyes on if it's needed in this case.

@lems9520138?

Saithej2k:
Thanks for the review! Tested the C++, Java, and Python snippets against the official released USACO test data and a small randomized check; all passed.

lems9520138:
> Thanks for the review! Tested the C++, Java, and Python snippets against the official released USACO test data and a small randomized check; all passed.

K then lgtm

lems9520138 (approved):

Sosuke23 (approved):

Issue #6069, opened 2026-03-24
Algolia Usage Spike


bqi343:
@emhgit @ryanchou-dev I'm getting alerts about 30% of algolia search quota being used within two days -- possibly related to the latest PR that touched this file?

https://github.com/cpinitiative/usaco-guide/commits/master/src/pages/problems/index.tsx

@thecodingwizard Can you check exactly when usage started spiking? I don't have permissions.

emhgit:
i believe it’s because the algolia indexing is occurring every time a preview deployment occurs on vercel because the indexing script is chained to the build script in package.json. i don’t have access to my PC right now, but i will fix this in around 4 hours and find an alternative for automating the indexing. if someone is available, i’d remove the indexing script from the package.json for now

ryanchou-dev:
Removed from package.json for now.

emhgit:
actually a solution might be to configure the algolia env variables (on vercel dashboard) to only be available on the production branch (master/main) of the repo

thecodingwizard:
last time this happened it was because every page view triggered an empty algolia search query. the solution was to lazy load algolia and only issue a search if the search modal opened.

haven't investigated what's going on this time though

bqi343:
@thecodingwizard @ryanchou-dev it looks like all deployments are failing since this commit: https://github.com/cpinitiative/usaco-guide/commit/18c297f39eb3fca26f4d1bee8d999a0149bc57a9

is it due to this commit or some other reason?

emhgit:
this is because the `yarn tsx scripts/index-content.ts` command was removed from the `prebuild` script in package.json. although the command contains the word “index,” the `yarn tsx scripts/index-content.ts` command is not involved with algolia and needs to be added back for builds to function properly.

bqi343:
> haven't investigated what's going on this time though

@thecodingwizard Can you check whether Ryan's change had any effect on algolia usage?

https://github.com/cpinitiative/usaco-guide/commit/18c297f39eb3fca26f4d1bee8d999a0149bc57a9

emhgit:
this [403 forbidden error](https://support.algolia.com/hc/en-us/articles/4406981959441-Why-is-my-app-blocked#:~:text=403) appeared when I tried using the search feature on my local development server (`yarn dev`):

<img width="1095" height="402" alt="Image" src="https://github.com/user-attachments/assets/c625bb36-b014-4ca5-bc03-9948603cf15d" />

[...]

PR #6333, opened 2026-06-29
Only reindex changed Algolia records to prevent full reindex on every build

dycoke:

- [x] I have tested my code.

Addresses https://github.com/cpinitiative/usaco-guide/issues/6069


src/utils/algolia-queries.ts
aunncodes: field is never used

aunncodes: match fields can probably be deleted but if it isn't this tags is duplicate

aunncodes: Small nit but promise.all can probably be moved to the return since returning promises instead of records is a bit confusing

aunncodes (changes requested):
Small nits, mostly good though

emhgit (approved):

PR #6342, opened 2026-06-30
Editorial for 412D

johnzhou721:

- [x] I have tested my code.

### DIFF
--- a/content/3_Silver/Conclusion.problems.json
+++ b/content/3_Silver/Conclusion.problems.json
@@ -452,10 +452,9 @@
       "source": "CF",
       "difficulty": "Hard",
       "isStarred": false,
-      "tags": ["DFS"],
+      "tags": ["DFS", "TopoSort"],
       "solutionMetadata": {
-        "kind": "autogen-label-from-site",
-        "site": "CF"
+        "kind": "internal"
       }
     },
     {
--- a/content/3_Silver/Conclusion.problems.json
+++ b/content/3_Silver/Conclusion.problems.json
@@ -450,7 +450,7 @@
       "name": "Giving Awards",
       "url": "https://codeforces.com/contest/412/problem/D",
       "source": "CF",
-      "difficulty": "Hard",
+      "difficulty": "Normal",
       "isStarred": false,
       "tags": ["DFS", "TopoSort"],
       "solutionMetadata": {
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -17,7 +17,7 @@ If there is, we can neglect an edge in a cycle, which still gives a viable confi
 
 We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
 This ensures that all neighbors of a vertex come before itself.
-In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as breaking an edge is permissible.
+In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
 
 We start DFS’ing from each node of it is not already visited.
 Note that it is possible for a non-visited node to be a neighbor of some other vertex; however, we still DFS it and mark it as visited, as the other vertex would be visited later in another DFS, where the current vertex will be skipped.
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -34,9 +34,9 @@ We use a post-order DFS to traverse all the nodes; at each node, we DFS all its 
 This ensures that all neighbors of a vertex come before itself.
 In case of encountering a cycle while traversing, we would revisit an already-visited vertex, which we can safely skip, as neglecting an edge is permissible.
 
-Note that by DFS'ing each unvisited vertex in order, we may start visiting a neighbor of some other vertex
-we have not yet traversed; however, that other vertex would be traversed in a later DFS, so the order
-is still correct.  The later DFS will also skip the current neighbor, as it has already been traversed.
+Note that by DFS'ing each unvisited vertex in order, we may visit a neighbor of some other vertex we have not yet visited.
+However, this is not an issue, as that other vertex would be traversed in a later DFS, so the order is still correct.
+Its DFS will simply skip the already visited neighbor.
 
 ## Implementation
 
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -47,7 +47,6 @@ Note that we may start visiting a neighbor of some other vertex we have not yet 
 
 ```cpp
 // Based on solution by JuanMata on CodeForces
-#include <cstring>
 #include <iostream>
 #include <vector>
 
@@ -76,8 +75,6 @@ int main() {
 		cin >> u >> v;
 		graph[u].push_back(v);
 	}
-
-	memset(visited, 0, sizeof(visited));
 
 	for (int i = 1; i <= n; i++) {
 		if (!visited[i]) { dfs(i); }
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -19,7 +19,9 @@ The correctness of solution 2 is mostly easily proven using information about DF
 
 ## Explanation
 
-Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
+Having each $a_i$ goes after its $b_i$ in the list results in a working configuration, as $b_i$ cannot directly follow $a_i$ when it is before $a_i$ in the first place.
+This can always happen, except in situations where the graph formed by connecting each $a_i$ to $b_i$ has cycles.
+We call each $b_i$ the neighbor of $a_i$.
 
 Construct a directed graph with edges $a_i\to b_i$, and call $b_i$ the neighbor of $a_i$.
 It is possible to make each node follow all its neighbors if there are no cycles.
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -19,7 +19,7 @@ We use a flood fill DFS to traverse all the nodes; at each node, we DFS all its 
 This ensures that all neighbors of a vertex come before itself.
 In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as breaking an edge is permissible.
 
-We start DFS’ing from each node of it is not already visited. 
+We start DFS’ing from each node of it is not already visited.
 Note that it is possible for a non-visited node to be a neighbor of some other vertex; however, we still DFS it and mark it as visited, as the other vertex would be visited later in another DFS, where the current vertex will be skipped.
 That is still correct, as we’ve already done the DFS on that vertex now.
 
@@ -32,52 +32,44 @@ That is still correct, as we’ve already done the DFS on that vertex now.
 <CPPSection>
 
 ```cpp
+#include <cstring>
 #include <iostream>
 #include <vector>
-#include <cstring>
 
 using namespace std;
 
 vector<int> graph[30005];
 bool visited[30005];
 
-void dfs(int current)
-{
-    // skip if visited
-    if (visited[current]) {
-        return;
-    }
-    visited[current] = true;
-    
-    // visit neighbors
-    for (int i = 0; i < graph[current].size(); i++) {
-        dfs(graph[current][i]);
-    }
-    
-    cout << current << ' ';
+void dfs(int current) {
+	// skip if visited
+	if (visited[current]) { return; }
+	visited[current] = true;
+
+	// visit neighbors
+	for (int i = 0; i < graph[current].size(); i++) { dfs(graph[current][i]); }
+
+	cout << current << ' ';
 }
 
-int main()
-{
-    int n, m;
-    cin >> n >> m;
-    
-    for (int i = 0; i < m; i++) {
-        int u, v;
-        cin >> u >> v;
-        graph[u].push_back(v);
-    }
-    
-    memset(visited, 0, sizeof(visited));
-    
-    for (int i = 1; i <= n; i++) {
-        if (!visited[i]) {
-            dfs(i);
-        }
-    }
-    
-    cout << endl;
-    return 0;
+int main() {
+	int n, m;
+	cin >> n >> m;
+
+	for (int i = 0; i < m; i++) {
+		int u, v;
+		cin >> u >> v;
+		graph[u].push_back(v);
+	}
+
+	memset(visited, 0, sizeof(visited));
+
+	for (int i = 1; i <= n; i++) {
+		if (!visited[i]) { dfs(i); }
+	}
+
+	cout << endl;
+	return 0;
 }
 ```
 
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -15,7 +15,7 @@ Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neig
 It is possible to make all nodes come after all its neighbors if there is no cycles in this graph.
 If there is, we can neglect an edge in a cycle, which still gives a viable configuration, as the cycles are at least of size $3$, meaning the two vertices associated with the edge we break will not be adjacent in the first place, so their direction does not matter.
 
-We use a flood fill DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
 This ensures that all neighbors of a vertex come before itself.
 In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as breaking an edge is permissible.
 
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -32,7 +32,7 @@ To see why, we use an example: we require vertex A to placed before B, B to be b
 
 We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
 This ensures that all neighbors of a vertex come before itself.
-In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
+In case of a encountering a cycle while traversing, we would revisit an already-visited vertex, which we can safely skip as neglecting an edge is permissible.
 
 We start DFS’ing from each node of it is not already visited.
 Note that we may start visiting a neighbor of some other vertex we have not yet traversed; however, that other vertex would be traversed in a later DFS, so the order is still correct.  The later DFS will also skip the current neighbor, as it has already been traversed.
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -32,6 +32,7 @@ That is still correct, as we’ve already done the DFS on that vertex now.
 <CPPSection>
 
 ```cpp
+// Based on solution by JuanMata on CodeForces
 #include <cstring>
 #include <iostream>
 #include <vector>
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -84,7 +84,6 @@ int main() {
 	}
 
 	cout << endl;
-	return 0;
 }
 ```
 
--- a/solutions/silver/cf-412D.mdx
+++ b/solutions/silver/cf-412D.mdx
@@ -12,7 +12,8 @@ author: John Zhou
 Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
 
 Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$.
-It is possible to make all nodes come after all its neighbors if there is no cycles in this graph.
+It is possible to make each node follow all its neighbors if there are no cycles.
+This will lead to a valid configuration, as an element cannot follow another directly if it is before the latter element in the first place.
 If there is, we can neglect an edge in a cycle, which still gives a viable configuration, as the cycles are at least of size $3$, meaning the two vertices associated with the edge we break will not be adjacent in the first place, so their direction does not matter.
 
 We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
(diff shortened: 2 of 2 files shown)


solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
+
+Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$ for each $i$.
+It is possible to make all nodes come after all its neighbors if there is no cycles in this graph.
+If there is, we can neglect an edge in a cycle, which still gives a viable configuration, as the cycles are at least of size $3$, meaning the two vertices associated with the edge we break will not be adjacent in the first place, so their direction does not matter.
+
+We use a flood fill DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
```
4nav: pretty sure its postorder dfs not flood fill dfs

content/3_Silver/Conclusion.problems.json
```diff
@@ -452,10 +452,9 @@
       "source": "CF",
       "difficulty": "Hard",
```
4nav: not ur fault but hard?

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+[Official Editorial](https://codeforces.com/blog/entry/11756)
+
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
+
+Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$ for each $i$.
+It is possible to make all nodes come after all its neighbors if there is no cycles in this graph.
+If there is, we can neglect an edge in a cycle, which still gives a viable configuration, as the cycles are at least of size $3$, meaning the two vertices associated with the edge we break will not be adjacent in the first place, so their direction does not matter.
```
4nav: i get what you mean but imo the wording in this paragraph is rlly dense and not very easy to digest - also saying breaking an edge implies we'd modify the graph but i think you meant ignore

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+We use a flood fill DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+This ensures that all neighbors of a vertex come before itself.
+In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as breaking an edge is permissible.
```
4nav: same wording problem here - "breaking" shd be ignoring

4nav (changes requested):

content/3_Silver/Conclusion.problems.json
lems9520138: Since for silver, I would definetly put it as Medium.

johnzhou721: I think it's hard because using post order may not be immediately intuitive because Silver does not have experience with toposort yet

@bqi343?

solutions/silver/cf-412D.mdx
johnzhou721:
```suggestion
We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
```

solutions/silver/cf-412D.mdx
johnzhou721:
```suggestion
In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,79 @@
...
+author: John Zhou
+---
+
+[Official Editorial](https://codeforces.com/blog/entry/11756)
+
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
+
+Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$ for each $i$.
```
johnzhou721:
```suggestion
Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,79 @@
...
+It is possible to make all nodes come after all its neighbors if there is no cycles in this graph.
```
johnzhou721:
```suggestion
It is possible to make each node follow all its neighbors if there are no cycles.
This will lead to a valid configuration, as an element cannot follow another directly if it is before the latter element in the first place.
```

solutions/silver/cf-412D.mdx
johnzhou721:
```suggestion
If there is a cycle, we can neglect an edge in a cycle. This is still valid, as the cycles are at least $3$ vertices.

To see why, we use an example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
+
+Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$.
+It is possible to make each node follow all its neighbors if there are no cycles.
+This will lead to a valid configuration, as an element cannot follow another directly if it is before the latter element in the first place.
+If there is a cycle, we can neglect an edge in a cycle. This is still valid, as the cycles are at least $3$ vertices.
+
+To see why, we use an example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
```
bqi343: this explanation doesn't convince me that the solution is correct. what if there are more cycles in the graph, which then require A->B or B->C to be removed?

content/3_Silver/Conclusion.problems.json
bqi343: the solution given in the official editorial is fairly simple which could justify lowering the difficulty

this internal sol seems harder though, and probably not in scope for USACO Silver. it appears to rely on DFS properties that aren't introduced in the Silver graph modules

solutions/silver/cf-412D.mdx
johnzhou721: orz orz orz

This would not happen in an actual DFS—we DFS A->B and B->C first before breaking C->A and by that point AB and BC is already guaranteed.  A different cycle back to A will break a different edge when A is marked as visited as AB is already traversed

what do you think?? Is the explanation above reasonable logically?  If so I’ll polish it and add it in an Info component as this soln is already mostly guessforces anyways

bqi343: idk, it's still not really convincing that this argument covers all possible cases

imo the most convincing way to show correctness is to rely on properties of the DFS tree introduced in this (not Silver) module: https://usaco.guide/adv/BCC-2CC?lang=cpp#resource-https%3A%2F%2Fcodeforces.com%2Fblog%2Fentry%2F68138

content/3_Silver/Conclusion.problems.json
johnzhou721: Well me proposing the different soln was mainly a cultural difference because in China you need to know postorder etc for CSP-J and NOIP’s lower division which is way easier than Silver in terms of problem-solving.

Anyways is it worth still appropriate to keep thisn solution? I can add it to toposort in gold as well—I just feel this soln is more intuitive overall.

```suggestion
      "difficulty": "Normal",
```

solutions/silver/cf-412D.mdx
johnzhou721: IMHO intuition behind this trick is a tad more useful than the formal proof

would you be against quoting the reference and explaining how to use DFS tree to prove in a spoiler?

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,82 @@
...
+source: CF
+title: Giving Awards
+author: John Zhou
+---
+
+[Official Editorial](https://codeforces.com/blog/entry/11756)
+
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
```
lems9520138: what do you mean follow indirectly? like just after it in the array?

solutions/silver/cf-412D.mdx
bqi343: That should be okay. Please include at top of the internal sol a note about the official editorial solution being very different, and that your solution is not necessarily in scope for USACO Silver.

johnzhou721: Thanks.  Tag me tomorrow if I don’t do this by then.

Also another TODO is to open an issue with writing the simplified solution (which I think is way too much guessforces)

lems9520138: I'll actually add solution 1 cuz I dont really like solution 2

johnzhou721: Looks like you've already done so -- I've been planning to do these changes today.  In that case, tysm!

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+This will lead to a valid configuration, as an element cannot follow another directly if it is before the latter element in the first place.
+If there is a cycle, we can neglect an edge in a cycle. This is still valid, as the cycles are at least $3$ vertices.
+
+To see why, we use an example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
+
+We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+This ensures that all neighbors of a vertex come before itself.
+In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
+
+We start DFS’ing from each node of it is not already visited.
```
bqi343: ^ don't understand what this is saying

same for the next sentence as well

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+<Warning>
+
+The correctness of solution 2 is mostly easily proven using information about DFS trees from [this resource](/adv/BCC-2CC?lang=cpp#resource-https%3A%2F%2Fcodeforces.com%2Fblog%2Fentry%2F68138). However, this is outside the scope of USACO Silver.
+
+</Warning>
+
+## Explanation
+
+Since there are no restrictions of both $u,v$ and $v,u$, it is always possible for $a_i$ to follow $b_i$ either directly or indirectly in each restriction; if we make this happen all the time (except in cyclic situations as described below), then we’d get a working configuration.
```
lems9520138: are you just trying to say it is always possible for a_i to be after b_i or trying to say that a_i and b_i can always touch?

Also, you have used u, v in the first part of the sentence and then transition to a_i, b_i  when it seems like both of these are just refering to a restriction, so I would try to keep it consistent.

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+Construct a directed graph with edges $a_i\to b_i$, where we call $b_i$ the neighbor of $a_i$.
```
lems9520138:
```suggestion
Construct a directed graph with edges $a_i\to b_i$, and call $b_i$ the neighbor of $a_i$.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+It is possible to make each node follow all its neighbors if there are no cycles.
```
lems9520138: so just double checking, but here by "follow" you meant reside after all it's neighbors in the array?

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+This will lead to a valid configuration, as an element cannot follow another directly if it is before the latter element in the first place.
```
lems9520138: Since this kinda feels a bit like weird sounding (lines 25-26) one way I would propose to make it clear is

If there are no cycles, by making each node follow all of it's neighbors, we are left with a valid configuration. For two elements in a cycle, ..... (something like it can't be after some neighbor will being required to be before it due to the cycle meaning, so a cycle is not possible ...).

My feedback is: (feel free to look at it however you want)
1. The transition from line 25-26 seems a bit weird to me like you are saying this will lead to a configuration implying line 25 was some sort of if we do ..., but line 25 was an assertion saying we have a valid configuration if this is true.
 
2. The part of line 26 when you explain "if it is before the latter element in the first place" seems a little unclear, because you don't really reference the cycle is why it is technically supposed to be before the latter element in the first place.

Really sorry for being hypercritical 😭

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+If there is a cycle, we can neglect an edge in a cycle. This is still valid, as the cycles are at least $3$ vertices.
+
+To see why, we use an example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
+
+We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+This ensures that all neighbors of a vertex come before itself.
+In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
```
lems9520138: In the case of encountering a cycle while traversing, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is allowable.
Is this what you meant? Also to clarify for myself, is ignoring an edge permissible according to that logic you had above?

solutions/silver/cf-412D.mdx
lems9520138: Likewise for 35, 36-37 can prolly be simplified cuz I kinda understand what you're saying??

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+	int n, m;
+	cin >> n >> m;
+
+	for (int i = 0; i < m; i++) {
+		int u, v;
+		cin >> u >> v;
+		graph[u].push_back(v);
+	}
+
+	memset(visited, 0, sizeof(visited));
```
lems9520138: Is there a reason for memset? I mean i dont care although I was js curious.

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+	for (int i = 1; i <= n; i++) {
+		if (!visited[i]) { dfs(i); }
+	}
+
+	cout << endl;
+	return 0;
```
lems9520138: Also yeah Idk if you need this

lems9520138 (changes requested):
Overall, it's getting there almost good to go, just a couple of english/structural changes!

Good job!

solutions/silver/cf-412D.mdx
johnzhou721: Nah first clause is also kinda redundant since that case is covered later.
```suggestion
Having each $a_i$ goes after its $b_i$ in the list results in a working configuration, as $b_i$ cannot directly follow $a_i$ when it is before $a_i$ in the first place.
This can always happen, except in situations where the graph formed by connecting each $a_i$ to $b_i$ has cycles.
We call each $b_i$ the neighbor of $a_i$.
```

solutions/silver/cf-412D.mdx
johnzhou721: Line 26 is about the no-cycles case.  I'm gonna cut this whole paragraph and tighten

solutions/silver/cf-412D.mdx
johnzhou721: Yes and yes
```suggestion
In case of a encountering a cycle while traversing, we would revisit an already-visited vertex, which we can safely skip as neglecting an edge is permissible.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+If there is a cycle, we can neglect an edge in a cycle. This is still valid, as the cycles are at least $3$ vertices.
+
+To see why, we use an example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
+
+We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+This ensures that all neighbors of a vertex come before itself.
+In case of a cycle, we would revisit an already-visited vertex, which we can safely skip as ignoring an edge is permissible.
+
+We start DFS’ing from each node of it is not already visited.
+Note that it is possible for a non-visited node to be a neighbor of some other vertex; however, we still DFS it and mark it as visited, as the other vertex would be visited later in another DFS, where the current vertex will be skipped.
```
johnzhou721:
```suggestion
Note that we may start visiting a neighbor of some other vertex we have not yet traversed; however, that other vertex would be traversed in a later DFS, so the order is still correct.  The later DFS will also skip the current neighbor, as it has already been traversed.
```

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,92 @@
...
+That is still correct, as we’ve already done the DFS on that vertex now.
```
johnzhou721:
```suggestion
```

johnzhou721 (commented):
@lems9520138 I'm tightening this up significantly.

solutions/silver/cf-412D.mdx
johnzhou721: The clause is cut and I clarified wording

solutions/silver/cf-412D.mdx
johnzhou721: Yep.  But I'm gonna cut this paragraph

solutions/silver/cf-412D.mdx
johnzhou721: Simplified a lot here.

solutions/silver/cf-412D.mdx
johnzhou721: Being chinese they teach you this method... they also encourage memset(0x7f) to do infinity.

solutions/silver/cf-412D.mdx
johnzhou721:
```suggestion
```

this was for nostalgia

solutions/silver/cf-412D.mdx
johnzhou721: but sincey ou brought it up imma edit it

johnzhou721: Hold your horses the real reason is because I copied this code from a public CodeForces submission someone shared and then heavily edited it; this was leftovers.

johnzhou721:
@lems9520138 @bqi343 This is ready for a rereview--thanks

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,84 @@
...
+This can always happen, except when the graph has cycles.
+If there is a cycle in the graph, however, we can neglect an edge in a cycle. This is still valid, as the cycles have at least $3$ vertices.
+
+To see why, consider this example: we require vertex A to placed before B, B to be before C, and C to be before A; from the first two requirements, A must be before C with at least $1$ element (B) in between. Since A and C are not adjacent, their order does not matter, so we may neglect the third edge.
+
+We use a post order DFS to traverse all the nodes; at each node, we DFS all its neighbors before outputting the current node number.
+This ensures that all neighbors of a vertex come before itself.
+In case of a encountering a cycle while traversing, we would revisit an already-visited vertex, which we can safely skip as neglecting an edge is permissible.
+
+Note that by DFS'ing each unvisited vertex in order, we may start visiting a neighbor of some other vertex we have not yet traversed; however, that other vertex would be traversed in a later DFS, so the order is still correct.  The later DFS will also skip the current neighbor, as it has already been traversed.
```
4nav:
```suggestion
Note that by DFS'ing each unvisited vertex in order, we may visit a neighbor of some other vertex we have not yet visited. However, this is not an issue, as that other vertex would be traversed in a later DFS, so the order is still correct. Its DFS will simply skip the already visited neighbor
```

4nav (approved):

solutions/silver/cf-412D.mdx
```diff
@@ -0,0 +1,84 @@
...
+## Implementation
+
+**Time complexity:** $\mathcal O(n+m)$
+
+<LanguageSection>
+
+<CPPSection>
+
+```cpp
+// Based on solution by JuanMata on CodeForces
```
Sosuke23: is this necessary?

johnzhou721: Let’s give credit cause I copied soln off of codeforces

johnzhou721: (I copied the code and changed it

johnzhou721:
@Sosuke23 can we merge now?

solutions/silver/cf-412D.mdx
Sosuke23: > Let’s give credit cause I copied soln off of codeforces

What about editorial then? Do you wrote it yourself or used LLM?

johnzhou721: I wrote editorial text myself but used the implementation from a codeforces submission

here
```suggestion
// Based on code by JuanMata on CodeForces
```

Sosuke23 (approved):

Sosuke23:
> @Sosuke23 can we merge now?

i need an approval from either @bqi343 or @lems9520138

bqi343:
I'll leave it to @lems9520138

lems9520138:
Yes, give me a day or ping if I forget I was going to write solution 1, and add it to this pr since Im not the biggest fan of this current solution (esp for a silver question)

johnzhou721:
@lems9520138 Bump

lems9520138 (approved):
I actually don't have time so I will js merge.

PR #6353, opened 2026-07-05
Update solutions/gold/usaco-899.mdx Fixed incorrect code (would have gotten TLE)

tinyfold88:

- [x] I have tested my code.


Sosuke23:
Can you provide some justification for your solution? 
for which case the current solution fails.

tinyfold88:
The reason is very similar to why you need that continue statment in normal dijkstra. Using a special contruction that maximises the number of relaxations will make the original code tle.

Sosuke23 (approved):

PR #6354, opened 2026-07-06
Cf 1690-G

4nav:

- [x] I have tested my code.

If anyone has a solution where we don't need to do lazy prop pls drop it down I'll credit you as co - author

### DIFF
--- a/content/3_Silver/Conclusion.problems.json
+++ b/content/3_Silver/Conclusion.problems.json
@@ -302,7 +302,7 @@
       "isStarred": false,
       "tags": ["Sorted Set"],
       "solutionMetadata": {
-        "kind": "Internal",
+        "kind": "Internal"
       }
     },
     {
--- a/content/3_Silver/Conclusion.problems.json
+++ b/content/3_Silver/Conclusion.problems.json
@@ -302,8 +302,7 @@
       "isStarred": false,
       "tags": ["Sorted Set"],
       "solutionMetadata": {
-        "kind": "autogen-label-from-site",
-        "site": "CF"
+        "kind": "Internal",
       }
     },
     {
--- a/content/3_Silver/Conclusion.problems.json
+++ b/content/3_Silver/Conclusion.problems.json
@@ -302,7 +302,7 @@
       "isStarred": false,
       "tags": ["Sorted Set"],
       "solutionMetadata": {
-        "kind": "Internal"
+        "kind": "internal"
       }
     },
     {
--- a/solutions/silver/cf-1690G.mdx
+++ b/solutions/silver/cf-1690G.mdx
@@ -37,61 +37,57 @@ To track distinct values, keep a **set of distinct speeds**, synced after each u
 ```cpp
 #include <bits/stdc++.h>
 using namespace std;
- 
+
 int main() {
 	ios::sync_with_stdio(false);
 	cin.tie(nullptr);
- 
+
 	int t;
 	cin >> t;
- 
+
 	while (t--) {
 		int n, q;
 		cin >> n >> q;
- 
+
 		vector<long long> a(n);
 		for (auto &x : a) cin >> x;
- 
+
 		// Build the initial set of train-start indices
 		vector<long long> pref_min(n);
 		pref_min[0] = a[0];
 		set<int> starts = {0};
- 
+
 		for (int i = 1; i < n; i++) {
 			pref_min[i] = min(pref_min[i - 1], a[i]);
-			if (pref_min[i] < pref_min[i - 1]){
-                starts.insert(i);
-            }
+			if (pref_min[i] < pref_min[i - 1]) { starts.insert(i); }
 		}
- 
+
 		while (q--) {
 			int k;
 			long long d;
 			cin >> k >> d;
 			k--;
- 
+
 			// Largest start index <= k governs carriage k's current true speed
 			int j = *prev(starts.upper_bound(k));
 			long long cur_speed = a[j];
- 
+
 			a[k] -= d;
- 
+
 			if (a[k] >= cur_speed) {
 				cout << starts.size() << " ";
 				continue;
 			}
- 
+
 			starts.insert(k);
- 
+
 			// Erase every later start index dominated by the new, lower speed
 			auto it = next(starts.find(k));
-			while (it != starts.end() && a[*it] >= a[k]) {
-				it = starts.erase(it);
-			}
- 
+			while (it != starts.end() && a[*it] >= a[k]) { it = starts.erase(it); }
+
 			cout << starts.size() << " ";
 		}
- 
+
 		cout << '\n';
 	}
 }
--- a/solutions/silver/cf-1690G.mdx
+++ b/solutions/silver/cf-1690G.mdx
@@ -1,0 +1,185 @@
+---
+id: cf-1690G
+source: CF
+title: Count the Trains
+author: Abhinav Singh
+---
+
+[Official Analysis (C++)](https://codeforces.com/blog/entry/103658)
+
+## Explanation
+
+Since carriage $i$ can never outrun anything in front of it, its true speed is the prefix minimum of speeds from the start. Since this array of speeds is non-increasing, it looks like a staircase, and each "step", or each change in value is a boundary between trains. So the answer is just the number of **distinct values** in $speed[]$.
+
+Each message lowers one engine's cap, which forces us to recompute how far that drop propagates: any carriage to the right that was already $\geq$ the new cap gets dragged down to it. If the new cap doesn't go below the carriage's current real speed, nothing changes: it was already bottlenecked by something ahead of it, so we skip the update entirely. 
+
+Otherwise, we need to apply the drop as a range assignment and update the distinct-value count without rescanning the array. Both problems are solved by maintaining the speeds in a **[segment tree with lazy propagation](https://usaco.guide/plat/RURQ?lang=cpp)** supporting range assign and point query. Since the speed array stays monotonic at all times, we can binary search on the tree to find the rightmost affected carriage.
+
+For the count itself, we keep a **set of distinct speeds** live. After the segment tree update reflects the new prefix minimums, we must sync the set. We can do this by taking the prefix min at the position right before ours ($prev$). In the set, use upper_bound to find the next larger value than the new cap -  if that value is less than prev, it's now stale (no carriage still has that speed) so we erase it, and keep walking forward erasing stale values until we hit one that's $\geq$ $prev.$
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}((N + M) \log^2 N)$
+
+<LanguageSection>
+<CPPSection>
+
+```cpp
+#include <bits/stdc++.h>
+using namespace std;
+
+vector<int> seg;
+vector<bool> lazy;
+
+void push(int v) {
+    if (!lazy[v]) return;
+
+    seg[2 * v] = seg[v];
+    seg[2 * v + 1] = seg[v];
+
+    lazy[2 * v] = true;
+    lazy[2 * v + 1] = true;
+
+    lazy[v] = false;
+}
+
+// Build the initial prefix minimum array.
+void build(int v, int tl, int tr, const vector<int>& pref_min) {
+    if (tl == tr) {
+        seg[v] = pref_min[tl];
+        return;
+    }
+
+    int tm = (tl + tr) / 2;
+    build(2 * v, tl, tm, pref_min);
+    build(2 * v + 1, tm + 1, tr, pref_min);
+}
+
+// Assign every value in [l, r] to val.
+void update(int v, int tl, int tr, int l, int r, int val) {
+    if (l > r) return;
+
+    if (tl == l && tr == r) {
+        seg[v] = val;
+        lazy[v] = true;
+        return;
+    }
+
+    push(v);
+
+    int tm = (tl + tr) / 2;
+
+    update(2 * v, tl, tm, l, min(r, tm), val);
+    update(2 * v + 1, tm + 1, tr, max(l, tm + 1), r, val);
+}
+
+// Point query.
+int query(int v, int tl, int tr, int pos) {
+    if (tl == tr){
+        return seg[v];
+    }
+
+    push(v);
+
+    int tm = (tl + tr) / 2;
+
+    if (pos <= tm){
+        return query(2 * v, tl, tm, pos);
+    }
+
+    return query(2 * v + 1, tm + 1, tr, pos);
+}
+
+int main() {
+    ios::sync_with_stdio(false);
+    cin.tie(nullptr);
+
+    int t;
+    cin >> t;
+
+    while (t--) {
+        int n, q;
+        cin >> n >> q;
+
+        vector<int> a(n);
+        for (int &x : a) cin >> x;
+
+        // Compute the initial prefix minima.
+        vector<int> pref_min(n);
+        pref_min[0] = a[0];
+
+        for (int i = 1; i < n; i++){
+            pref_min[i] = min(pref_min[i - 1], a[i]);
+        }
+
+        seg.assign(4 * n + 5, 0);
+        lazy.assign(4 * n + 5, false);
+
+        build(1, 0, n - 1, pref_min);
+
+        // Store the distinct prefix minima.
+        set<int> distinct_speeds;
+        for (int i = 0; i < n; i++){
+            distinct_speeds.insert(query(1, 0, n - 1, i));
+        }
+
+        while (q--) {
+            int k, d;
+            cin >> k >> d;
+            k--;
+
+            a[k] -= d;
+
+            // Current real speed of carriage k.
+            int cur = query(1, 0, n - 1, k);
+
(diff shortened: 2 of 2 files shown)


content/3_Silver/Conclusion.problems.json
```diff
@@ -302,8 +302,7 @@
       "isStarred": false,
       "tags": ["Sorted Set"],
       "solutionMetadata": {
-        "kind": "autogen-label-from-site",
-        "site": "CF"
+        "kind": "Internal"
```
johnzhou721:
```suggestion
        "kind": "internal"
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
+---
+id: cf-1690G
+source: CF
+title: Count the Trains
+author: Abhinav Singh
+---
+
```
johnzhou721, line 7: Segment trees in Silver is kinda admitting orz--but after this PR merges,
please file an issue to add a non segment tree solution.

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
...
+[Official Analysis (C++)](https://codeforces.com/blog/entry/103658)
+
+## Explanation
+
+Since carriage $i$ can never outrun anything in front of it, its true speed is the prefix minimum of speeds from the start. Since this array of speeds is non-increasing, it looks like a staircase, and each "step", or each change in value is a boundary between trains. So the answer is just the number of **distinct values** in $speed[]$.
```
johnzhou721:
```suggestion
Since carriage $i$ can never outrun anything in front of it, its true speed is the prefix minimum of speeds from the start. Since this array is non-increasing, it looks like a staircase, and each change in value is a boundary between trains. So the answer is just the number of distinct values in this prefix minimum array, $\mathtt{speed}$.
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+Each message lowers one engine's cap, which forces us to recompute how far that drop propagates: any carriage to the right that was already $\geq$ the new cap gets dragged down to it. If the new cap doesn't go below the carriage's current real speed, nothing changes: it was already bottlenecked by something ahead of it, so we skip the update entirely.
```
johnzhou721:
```suggestion
Each message lowers one engine's cap, and causes any carriage to the right that has a higher speed than the new cap to slow down to it. If the new cap doesn't go below the carriage's current real speed, then we skip the update entirely, as it was already bottlenecked by engines in front of it.
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+Otherwise, we need to apply the drop as a range assignment and update the distinct-value count without rescanning the array. Both problems are solved by maintaining the speeds in a **[segment tree with lazy propagation](https://usaco.guide/plat/RURQ?lang=cpp)** supporting range assign and point query. Since the speed array stays monotonic at all times, we can binary search on the tree to find the rightmost affected carriage.
```
johnzhou721:
```suggestion
Otherwise, we need to apply the drop as a range assignment and update the distinct-value count without rescanning the array, which is solved by maintaining the speeds in a **[segment tree with lazy propagation](https://usaco.guide/plat/RURQ?lang=cpp)** to support range assign and point query. Since the speed array stays monotonic, we binary search on it to find the rightmost affected carriage to assign the range to.
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+For the count itself, we keep a **set of distinct speeds** live. After the segment tree update reflects the new prefix minimums, we must sync the set. We can do this by taking the prefix min at the position right before ours ($prev$). In the set, use upper_bound to find the next larger value than the new cap -  if that value is less than prev, it's now stale (no carriage still has that speed) so we erase it, and keep walking forward erasing stale values until we hit one that's $\geq$ $prev.$
```
johnzhou721:
```suggestion
For the count, we maintain a **set of distinct speeds**. After the segment tree update, we sync the set by taking $\mathtt{prev}$, the effective speed at the position right before the change, and use `upper_bound` to find the next larger value than the new cap. If that value is less than $\mathtt{prev}$, it's now stale (as the section where the speed gradually decreases has effectively been flattened to a single value) so we erase it.  We repeat this process to remove all stale speeds until the value is larger or equal to $\textt{prev}$, and then insert the new speed.
```

Now justify the time complexity of this

johnzhou721 (changes requested):
Some clarity improvements; on top of that, tighten the wording anc clean up where you split your paragraphs.  Good work

solutions/silver/cf-1690G.mdx
4nav: gotchu

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+For the count, we maintain a **set of distinct speeds**. After the segment tree update, we sync the set by taking $\mathtt{prev}$, the effective speed at the position right before the change, and use `upper_bound` to find the next larger value than the new cap. If that value is less than $\mathtt{prev}$, it's now stale (as the section where the speed gradually decreases has effectively been flattened to a single value) so we erase it.  We repeat this process to remove all stale speeds until the value is larger or equal to $\textt{prev}$, and then insert the new speed.
+
+## Implementation
+
+**Time Complexity:** Each query performs binary search over the speed array, with each step requiring a segment tree query, for a total of $\mathcal{O}(log^2 N)$ time. The remaining segment tree updates and set operations each take $\mathcal{O}(log N)$, while set erasures are amortized since each distinct speed is removed at most once, giving us the total time complexity of $\mathcal{O}((N + M) \log^2 N)$
```
johnzhou721: Can probably tighten

johnzhou721 (approved):

eysbutno (changes requested):
lazy segtree on a silver problem..?

surely there's a way to solve it without lazy seg? like i'm fine with mentioning it as an aside, but having it as the main sol is not ideal

johnzhou721:
agree if we review + merge this we should at least add the official editorial on a separate soln heading and then file a follow up issue to write it up

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+id: cf-1690G
+source: CF
+title: Count the Trains
+author: Abhinav Singh
+---
+
+[Official Analysis (C++)](https://codeforces.com/blog/entry/103658)
+
+## Explanation
+Let $a_i$ be the raw (possibly reduced) cap on carriage $i$, and $v_i$ its true speed. Since carriage $i$ can never outrun anything in front of it, $v_i$ is the prefix minimum $v_i = \min(a_1, \ldots, a_i)$. Since this array is non-increasing, it looks like a staircase, and each change in value is a boundary between trains. So the answer is just the number of distinct values in $v_i$.
```
johnzhou721:
```suggestion
Let $a_i$ be the current cap on carriage $i$, and $v_i$ the true speed it is running at. Since carriage $i$ can never outrun anything in front of it, $v_i$ is the prefix minimum $v_i = \min(a_1, \ldots, a_i)$. Since this array is non-increasing, it looks like a staircase, and each change in value is a boundary between trains. So the answer is just the number of distinct values in $v_i$.
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+Instead of storing the whole prefix minimum array, maintain a set of just the **indices** where a new train starts: : index $0$, plus every index $i$ where $v_i < v_{i-1}$. A "start index" $j$ is where the prefix min drops, meaning $v_j$ itself was the new minimum, or equivalently $a_j = v_j$ . To get the true speed at any position, we can just find its governing start index $j$ and just read off of it's raw value, i.e $a_j$.
```
johnzhou721:
```suggestion
Instead of storing the whole prefix minimum array, maintain a set of **start indices**, or the indices where a new train starts, which are index $0$ and every index $i$ where $v_i < v_{i-1}$. Note that $a_j = v_j$ at any start index $j$, so the true speed of a position $k$ is just $a_j$ with $j$ being its the largest start index such that $j\le k$.
```

We can cut a lot of duplication here.

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+When a query lowers $a_k$, find the largest start index $j \le k$ in the set - $a_j$ is $k$'s current true speed. If the new $a_k$ is still $\ge a_j$, nothing changes, since carriage $k$ was already bottlenecked by something ahead of it. Otherwise, insert $k$ into the set as a new start, then walk forward erasing every subsequent start index $j'$ with $a_{j'} \ge a_k$, since those trains now have been swallowed a lower speed limit. The size of the set after each query is the answer.
```
johnzhou721: What if $k$ is already a starting index?

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+
+
+
+<Info title="Alternative: Segment Tree Approach">
+
+Each message lowers one engine's cap, and causes any carriage to the right that has a higher speed than the new cap to slow down to it. If the new cap doesn't go below the carriage's current real speed, then we skip the update entirely, as it was already bottlenecked by engines in front of it. Otherwise, apply the drop as a range assignment and update the distinct-value count without rescanning the array, which we can do by maintaining the speeds in a **[segment tree with lazy propagation](https://usaco.guide/plat/RURQ?lang=cpp)** to support range assign and point query. Since the speed array stays monotonic, we binary search on it to find the rightmost affected carriage to assign the range to.
+
+To track distinct values, keep a **set of distinct speeds**, synced after each update: take $\mathtt{prev}$, the speed just before the change, then use `upper_bound` to find the next larger value than the new speed and erase it if it's less than $\mathtt{prev}$ (i.e. stale, since that section just got flattened to one value). Repeat until the next value is $\ge \mathtt{prev}$, then insert the new speed.
```
johnzhou721: Just get rid of this entirely since a lot of this is the same as the above; just say we can also use a lazy segtree with range assign and point query to maintain and query the speeds as well, rather than binary-searching from a set to get our desired speed.

but tbh i think this is better removed it's just a bit of extra complexity imho that doesn't have any advantage

@eysbutno i want another opinion on this

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+	int t;
+	cin >> t;
+
+	while (t--) {
+		int n, q;
+		cin >> n >> q;
+
+		vector<long long> a(n);
+		for (auto &x : a) cin >> x;
```
johnzhou721: Braces

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,96 @@
...
+using namespace std;
+
+int main() {
+	ios::sync_with_stdio(false);
+	cin.tie(nullptr);
+
+	int t;
+	cin >> t;
+
+	while (t--) {
```
johnzhou721: Don't use the name t and t--, use `test_num` and a for-loop with t=0...<test_num

johnzhou721 (changes requested):
Infobox is a bit too long and not really connected with the solution; besides that there's a paragarph or two that you can slime down a bit, and a minor functional isssue.  greatwork

solutions/silver/cf-1690G.mdx
4nav: I dont mind removing it tbh

johnzhou721: yeah but i might've messed up when evaluating the instructional value

johnzhou721:
@4nav bump

4nav:
k im back

solutions/silver/cf-1690G.mdx
4nav: gotcha

solutions/silver/cf-1690G.mdx
4nav: lol ima keep it as of now but @eysbutno its up to you

solutions/silver/cf-1690G.mdx
johnzhou721: Nvm… the subsequrn explanation is compatible regardless

solutions/silver/cf-1690G.mdx
johnzhou721: @4nav I don’t mind keeping but it has a lot in common with main solution so let’s trim this section down

johnzhou721 (approved):
Trim the info box down and we’re good so approving

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,94 @@
...
+<Info title="Alternative: Segment Tree Approach">
+
+Instead of storing train boundaries in a set, we can maintain the speeds in a **[segment tree with lazy propagation](https://usaco.guide/plat/RURQ?lang=cpp)** to support range assign and point query. Since the speed array is always non-increasing, we can binary search for the last carriage whose speed exceeds the new cap, assign that suffix to the new speed, and update the number of distinct speeds accordingly.
+
+</Info>
+
+
+## Implementation
+
+**Time Complexity:** Each index is inserted into the set at most once and erased at most once over the whole run, so the total erase work across all queries is $\mathcal{O}(n \log n)$, on top of one $\mathcal{O}(\log n)$ lookup per query.
```
eysbutno: explain the amortization in explanation and just keep the time complexity itself in the implementation header

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,94 @@
...
+title: Count the Trains
+author: Abhinav Singh
+---
+
+[Official Analysis (C++)](https://codeforces.com/blog/entry/103658)
+
+## Explanation
+Let $a_i$ be the current cap on carriage $i$, and $v_i$ the true speed it is running at. Since carriage $i$ can never outrun anything in front of it, $v_i$ is the prefix minimum $v_i = \min(a_1, \ldots, a_i)$. Since this array is non-increasing, it looks like a staircase, and each change in value is a boundary between trains. So the answer is just the number of distinct values in $v$.
+
+Instead of storing the whole prefix minimum array, maintain a set of **start indices**, or the indices where a new train starts, which are index $0$ and every index $i$ where $v_i < v_{i-1}$. Note that $a_j = v_j$ at any start index $j$, so the true speed of a position $k$ is just $a_j$ with $j$ being its the largest start index such that $j\le k$.
```
eysbutno:
```suggestion
Let $a_i$ be the maximum speed of carriage $i$, and $v_i$ the current speed it is running at. Since carriage $i$ can never outrun carriages in front of it, $v_i$ is the prefix minimum $v_i = \min(a_1, \ldots, a_i)$. The array $v_i$ is non-increasing, with each change in value being a boundary between trains. Thus, at any given moment, the answer is the number of distinct values in $v$.

Instead of storing the whole prefix minimum array, maintain a set of *start indices*, or the indices where a new train starts, which are index $0$ and every index $i$ where $v_i < v_{i-1}$. Note that $a_j = v_j$ at any start index $j$, so the true speed of a position $k$ is $a_j$, where $j$ is the closest starting index to the left of $k$.
```

solutions/silver/cf-1690G.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+When a query lowers $a_k$, find the largest start index $j \le k$ in the set - $a_j$ is $k$'s current true speed. If the new $a_k$ is still $\ge a_j$, nothing changes, since carriage $k$ was already bottlenecked by something ahead of it. Otherwise, insert $k$ into the set as a new start, then walk forward erasing every subsequent start index $j'$ with $a_{j'} \ge a_k$, since those trains now have been swallowed a lower speed limit. The size of the set after each query is the answer.
```
eysbutno:
```suggestion
When a query decreases $a_k$, one of the following happens:
- The true speed $v_k$, as defined by a starting index, is equal to or slower than $a_k$. Thus, the set of starting indices (i.e., the number of trains) remains the same.
- The previous true speed $v_k$ is greater than $a_k$, resulting in $k$ becoming a starting index. Note that $k$ becoming a starting index may result in some starting indices becoming obsolete. To remove these obsolete indices, we iteratively check the nearest starting index to the right of $k$, and see if its value is greater than $a_k$. If so, we erase and continue checking indices.

At each point, the number of trains is equal to the number of starting indices. We have at most $\mathcal{O}(N+Q)$ starting index insertions, and thus the number of index removals is also bounded by $\mathcal{O}(N+Q)$. Thus, the overall time complexity amortizes to $\mathcal{O}((N+Q) \log N)$.
```

or something like this? i think explaining the motivations a bit more would help

eysbutno (commented):
overall looks good

solutions/silver/cf-1690G.mdx
4nav: fairs

eysbutno (approved):

Issue #6235, opened 2026-06-05
Contact Form Submission - Mistake (General - Olympiads)


maggieliu05:
Someone submitted the contact form!

**URL**: https://usaco.guide/general/olympiads
**Module**: General - Olympiads
**Topic**: Mistake
**Message**: 
Is there any replacement for the acmicpc.net link? It's down indefinitely.

Same question for Grass Planting from Plat HLD module

redbaron1234:
Should we take down this problem from the guide? Since I tried searching for replacements for both, but couldn't find any.

bqi343:
For Grass Planting https://www.spoj.com/problems/GRASSPLA/ might not work

(there is a warning that it is broken but I'm not sure whether that is true or not)

redbaron1234:
I tried testing the solution code into the online judge, but it came out as a runtime error, so I'm not sure if I should replace with this link. I can take down the problem if there are no more online judges.

redbaron1234:
Actually, if we update the USACO guide IDE online judge with the test data, then maybe we could replace both links with our own online judge, but that might take a while, so for now I'll submit a PR for removing the problem from the Plat Page, then later I can try and add the test data to the IDE online judge.

bqi343:
> I tried testing the solution code into the online judge, but it came out as a runtime error

Really? I tried the model solution (minus the file I/O) and it ACed

(https://usaco.org/current/data/sol_grassplant.html)

redbaron1234:
Oh! I tried an external solution, since I couldn't find the model sol, so I guess it was wrong. I'll submit a new PR just replacing the link, again sorry about that.

PR #6357, opened 2026-07-07
Delete outdated problem

redbaron1234:

I temporarily deleted this problem, since there wasn't a replacement for the acmicpc link, or the official USACO problem for testing the solution. We can add it back once the guide IDE online judge has the test data for this specific problem (Grass Planting).

Closes issue #6235

- [x] I have tested my code.


lems9520138:
Is there a reason for the deletion other than just outdated. Imo I don't know many instances where we delete a question because it's old, but I could be wrong

redbaron1234:
The main reason for the deletion was because there was no way to check the solution, but now there was a link that will be replaced.

PR #6381, opened 2026-07-16
Editorial - Alchemy

Sosuke23:

- [ ] I have tested my code.


solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,75 @@
...
+## Explanation
+
+Bessie the cow wants to maximise her units of metal $N$, for which
+she can use the recipes provided to convert other constituent
+metals into metal $N$.
+
+We can recursively check the $N$-th metal constituents, and if it
+has enough availability we will consume one unit of it and check
+for other metal constituents. We will repeat this until we run out
+of supply for constituent metals.
```
eysbutno, line 19: should address why this greedy works:
- since there's only one recipe for each metal, it's a binary choice between constructing a unit of the metal, and using up a unit that already exists
- we can always delay constructing a new unit of metal, so we should prioritize using what we already have and constructing what's needed later if needed
- recursion is easiest way to impl

solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+## Implementation
+
+**Time Complexity:** $$
```
eysbutno: add time comp?

ik editorial states code is $\mathcal{O}(N^2 A)$ but recursive code is probably slower cuz ur revisiting nodes a lot more (eg consider a complete graph)

solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+int main() {
+	int n;
+	std::cin >> n;
+	std::vector<int> metals(n);
+	for (int &x : metals) { std::cin >> x; }
+
+	int k;
+	std::cin >> k;
+	std::unordered_map<int, std::vector<int>> recipes;
```
eysbutno: why not `std::vector<std::vector<int>> recipes(n)` instead

cuz $l$ is the index for a metal which is bounded by $n$

Sosuke23: just preference, not reason in particular.

should i change map -> vector

solutions/bronze/usaco-1229.mdx
Sosuke23: is it somethinh like $O(2^{\text{max depth of recipe chain}})$, if i am not tripping

solutions/bronze/usaco-1229.mdx
Sosuke23: okay, thanks for the points 👍

solutions/bronze/usaco-1229.mdx
eysbutno: generally yes if it's convenient for the problem

solutions/bronze/usaco-1229.mdx
eysbutno: probably? i'm not 100% sure to be honest

you can also mention the iterative approach when noting the time complexity

solutions/bronze/usaco-1229.mdx
Sosuke23: okay

solutions/bronze/usaco-1229.mdx
Sosuke23: > you can also mention the iterative approach when noting the time complexity

I don't get it

eysbutno: since I think the time complexities of the iterative and recursive approaches are different, you can note that the iterative approach is faster / has a different time complexity

(correct me if im wrong tho)

Sosuke23: yup, u r right

<Info title = "Note">

The intuitive approach of this problem is faster than this 
recursive approach. Can you figure out what it might be?

</Info>

will this suffice?

solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,99 @@
...
+1. Since there is only one recipe for each metal, constructing a
+new unit of any metal is a binary choice: either we consume from
+our existing stock, or we construct a new unit from its recipe.
+
+2. We can always delay constructing a new unit of a metal, so we
+should prioritize using our existing stock first and build new
+units of metal only if necessary.
+
+3. The ordering of recipes forms a tree-like structure, so a
+recursive approach is easier to implement than other approaches.
```
eysbutno, line 38:
```suggestion
Bessie the cow wants to maximise her units of metal $N$, for which
she can use the recipes provided to convert other constituent
metals into metal $N$.

We can recursively check the $N$-th metal constituents, and if it
has enough availability, we will consume one unit of it and check
for other metal constituents. We will repeat this until we run out
of supply for constituent metals.

We recursively check if it is possible to create a unit of metal $N$.
For the current metal, we either consume
an existing unit of it, or we recursively construct the necessary metals in its recipe. 
We construct units of metal $N$ recursively until it is no longer possible.

This recursive greedy approach works perfectly for this problem for
the following reasons:

- Since there is only one recipe for each metal, constructing a
new unit of any metal is a binary choice: either we consume from
our existing stock, or we construct a new unit from its recipe.

- We can always delay constructing a new unit of metal, so we
should prioritize using our existing stock first and build new
units of metal only if necessary.

The ordering of recipes forms a tree-like structure, so a
recursive approach is easier to implement than other approaches.
```

solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(2^{max\_depth\_of\_recipe\_chain})$
+
+<Info title = "Note">
+
+The intuitive approach of this problem is faster than this
+recursive approach. Can you figure out what it might be?
```
eysbutno:
```suggestion
The iterative implementation of this algorithm is more efficient. As an exercise,
try implementing an iterative approach that runs in $\mathcal{O}(N^2 \cdot \max(a_i), using the fact
that recipes utilize strictly less valuable metals.
```

solutions/bronze/usaco-1229.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+</Info>
```
eysbutno:
```suggestion
**Time Complexity:** $\mathcal{O}(N \cdot \sum a_i)$

<Info>

At first glance, the recursive code appears to be exponential. However, this is not the case due to amortization.

Consider an individual unit of a metal $i$. If possible, it will eventually be used to create a unit of metal $N$, meaning that it will contribute to at most $N$ calls. Since our source of metal can only deplete, we can analyze the individual contributions of each unit of metal to analyze the complexity of the algorithm.

</Info>
```

i believe the reasoning is correct ? since during the time it's possible to create a unit of metal $N$, the logic holds, and because we terminate early if it's not possible, idt there'd be any time complexity blow-ups

eysbutno (approved):

lems9520138 (approved):

lems9520138:
lgtm

PR #6382, opened 2026-07-17
Editorial - Candy Cane Feast

Sosuke23:


solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+[Official Editorial (C++, Java, Python)](https://usaco.org/current/data/sol_prob1_bronze_dec23.html)
+
+## Explanation
+
+To be straightforward, we can iterate over every candy cane height
+one by one and for each one we will simply calculate how much
+candy each of Farmer John's cow can eat from that particular candy
+cane and increment their height based on that and repeat this for
+the other candy canes.
```
johnzhou721: In the code you have an optimization checking for `curr_candy_height < candy_height[i]` so might wanna explain that intuitively (just say that we stop as soon as the candy cane is eaten up) and then justify the time complexity of that being able to pass both intuitively (the cows get larger and thus eat more candycane each time and thus we need to iterate over less cows) and mathematically derive it

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## Implementation
+
+**Time Complexity:** $\mathcal{O}(M + N \log (10^9))$
```
johnzhou721: Don't use constants in TC?  Define a variable for the maximum height of candycanes/cows (whichever one leads to the 1e9)

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+<LanguageSection>
+<CPPSection>
+
+```cpp
+#include <iostream>
+#include <vector>
+
```
johnzhou721, line 40: I think this would benefit from using namespace std.

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,58 @@
...
+			int can_add =
+			    std::max(0, (int)std::min(cow_height[j], (long long)candy_height[i]) -
+			                    curr_candy_height);
+			cow_height[j] += can_add;
+			curr_candy_height += can_add;
+			j++;
+		}
+	}
+
+	for (int i = 0; i < n; i++) { std::cout << cow_height[i] << '\n'; }
```
johnzhou721, line 64: Use range-based for ``for (int x: cow_height)`` may be better.

content/extraProblems.json
```diff
@@ -1653,12 +1653,11 @@
       "name": "Candy Cane Feast",
       "url": "http://www.usaco.org/index.php?page=viewproblem2&cpid=1347",
       "source": "Bronze",
-      "difficulty": "N/A",
+      "difficulty": "Easy",
```
johnzhou721, line 1655: usaco power surge lol (this is good)

johnzhou721 (changes requested):
I think there's an optimization you should lay out explicitly since this is Bronze, some other comments about impl as well.

solutions/bronze/usaco-1347.mdx
Sosuke23: I just don't use it anymore

solutions/bronze/usaco-1347.mdx
Sosuke23: done

solutions/bronze/usaco-1347.mdx
Sosuke23: done

solutions/bronze/usaco-1347.mdx
johnzhou721: Sure then.  it does get a bit repetitive, but it works.

solutions/bronze/usaco-1347.mdx
johnzhou721: This is still unresolved, but minor.

johnzhou721 (approved):
Tweaked a bit but LGTM

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,65 @@
...
+For each candy cane, we will simply calculate how much
+candy each of Farmer John's cow can eat from that particular candy
+cane and increment their height based on that and repeat this for
+the other candy canes.
+
+Here, we stop processing the remaining cows if there is no candy
+left in the current candy cane. Since after each new candy cane is
+hung, Farmer John's cows would grow taller and thus consume
+more of each candy cane, the remaining cows will have fewer chances
+to eat, so our solution runs faster than $\mathcal O(NM)$ in practice.
```
lems9520138: I think it would be nice to expand on this a little bit, including the part where like the first cow will double in height every operation, which is why we have log H and why the process is faster than O(NM) in practice.

Feels a bit incomplete without that imo.

lems9520138 (changes requested):

solutions/bronze/usaco-1347.mdx
johnzhou721: Yep.  Feel like a rigorous approach would be justified here...

solutions/bronze/usaco-1347.mdx
Sosuke23: i will leave this for preference

solutions/bronze/usaco-1347.mdx
Sosuke23: updated it.

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,113 @@
...
+For each candy cane, we will simply calculate how much
+candy each of Farmer John's cow can eat from that particular candy
+cane and increment their height based on that value, repeating
+this process for the other candy canes.
+
+Here, we stop processing the remaining cows if there is no candy
+left in the current candy cane. Since after each new candy cane is
+hung, Farmer John's cows grow taller and thus consume more of each
+candy cane, the remaining cows have fewer chances to eat. To be exact,
+every time a cow eats from a candy cane, its height at least doubles,
```
4nav: wdym its height doubles? im probs interpreting this wrongly but like if theres a candy cane at height 10, and 2 cows of height 9 and 10 respectively, then the first cow would eat 9 and the next one could only eat height 1?

4nav:
funnily enough o(n*m logn) passes here with a segtree

solutions/bronze/usaco-1347.mdx
Sosuke23: good catch... let me fix this

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,115 @@
...
+this process for the other candy canes.
+
+Here, we stop processing the remaining cows if there is no candy
+left in the current candy cane. Since after each new candy cane is
+hung, Farmer John's cows grow taller and thus consume more of each
+candy cane, the remaining cows have fewer chances to eat. To be exact,
+every time a cow eats from a candy cane, it eats at least as much
+as its current height, provided the candy cane is tall enough. In
+such cases, the cow's height at least doubles,meaning each cow can
+actively eat at most $\log H$ times across all candy canes, where
```
eysbutno: i don't think this is correct?

the argument is that the first cow will either eat the entirety of the cane we iterate on (in which we can quickly break out of the loop), or that the first cow's height will double -- and that means it can double at most 30 times, meaning we can just naively simulate it w/ an optimization

solutions/bronze/usaco-1347.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+	for (int &x : candy_height) { std::cin >> x; }
+
+	for (int i = 0; i < m; i++) {
+		int curr_candy_height = 0;
+		int j = 0;
+		while (j < n and curr_candy_height < candy_height[i]) {
+			int can_add =
+			    std::max(0, (int)std::min(cow_height[j], (long long)candy_height[i]) -
+			                    curr_candy_height);
```
eysbutno: formatting here is weird -- maybe make an additional variable for the bounding height?

solutions/bronze/usaco-1347.mdx
Sosuke23: i think the current exp is correct wrt to all cows

eysbutno: upon rereading i don't think it's incorrect but it doesn't exactly make sense, as only the first cow's height will at least double

Sosuke23: I have updated the explanation

eysbutno (approved):

lems9520138 (approved):

PR #6399, opened 2026-07-22
removed template from cses-1636

4nav:

- [x] I have tested my code.

### DIFF
--- a/solutions/gold/cses-1636.mdx
+++ b/solutions/gold/cses-1636.mdx
@@ -51,31 +51,30 @@ using namespace std;
 const int MOD = 1e9 + 7;
 
 int main() {
-    ios::sync_with_stdio(false);
-    cin.tie(nullptr);
+	ios::sync_with_stdio(false);
+	cin.tie(nullptr);
 
-    int n, x;
-    cin >> n >> x;
+	int n, x;
+	cin >> n >> x;
 
-    vector<int> coins(n);
-    for (int i = 0; i < n; i++) {
-        cin >> coins[i];
-    }
+	vector<int> coins(n);
+	for (int i = 0; i < n; i++) { cin >> coins[i]; }
 
-    vector<long long> dp(x + 1);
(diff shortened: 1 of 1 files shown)


solutions/gold/cses-1636.mdx
```diff
@@ -47,44 +47,34 @@ with the same set of coins ordered differently.
...
+	for (int coin : coins) {
+		for (int sum = coin; sum <= x; sum++) {
+			dp[sum] += dp[sum - coin];
+			dp[sum] %= MOD;
 		}
 	}
+
 	cout << dp[x] << '\n';
+
+	return 0;
```
eysbutno:
```suggestion
	// Process each coin once so each combination is counted only once 
	// (Order of the coins does not matter here)
	for (int coin : coins) {
		for (int sum = coin; sum <= x; sum++) {
			dp[sum] += dp[sum - coin];
			dp[sum] %= MOD;
}
}

cout << dp[x] << '\n';
```

also, idk if "order does not matter" is the best way to describe it, since the problem specifically says that we consider the ordered sets -- i think people may be confused by that. I think we can just remove that part of the comment

eysbutno (approved):

solutions/gold/cses-1636.mdx
4nav: We consider unordered sets tho unless im having a brain fart like if they were ordered (2,3,5) !=(5,3,2)

johnzhou721: took a look and the use of the word "ordered" in the problem seems quite arbitrary

so if we clarify order of the coins does not matter then i think it's not gonna be confusing

eysbutno's suggestion has some formatting issues but it's fine as pre-commit will fix it.

johnzhou721 (approved):
I'm gonna review this since I find it very interesting how people somehow finds writing with templates *easier*.

looks good though, the one thing flagged by eysbutno still applies

PR #6400, opened 2026-07-22
Tag a bunch of extraProblems

johnzhou721:

As title.

- [x] I have tested my code.

### DIFF
--- a/content/extraProblems.json
+++ b/content/extraProblems.json
@@ -338,7 +338,7 @@
       "source": "Silver",
       "difficulty": "Normal",
       "isStarred": false,
-      "tags": ["Connected Components", "Graphs", "DSU"],
+      "tags": ["Connected Components", "Graphs"],
       "solutionMetadata": {
         "kind": "internal"
       }
--- a/content/extraProblems.json
+++ b/content/extraProblems.json
@@ -97,7 +97,7 @@
       "source": "Bronze",
       "difficulty": "Normal",
       "isStarred": false,
-      "tags": ["Simulation", "Greedy"],
+      "tags": ["Simulation"],
       "solutionMetadata": {
         "kind": "internal"
       }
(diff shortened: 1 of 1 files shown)


johnzhou721:
(FWIW:  I requested reviews on some people I've usually interacted with and/or people who manages problems a lot like Benq.  There's no obligation for you to review anything or approve/request changes the whole PR, since this tags a lot of problems from different stuff I just want some perspective as to any alternative tags y'all may come up with.  If you don't wanna do that that is fine.)

content/extraProblems.json
```diff
@@ -338,7 +338,7 @@
       "source": "Silver",
       "difficulty": "Normal",
       "isStarred": false,
-      "tags": [],
+      "tags": ["DSU", "Graphs"],
```
bqi343: should this have "Connected Components" instead of "DSU"

content/extraProblems.json
```diff
@@ -97,7 +97,7 @@
       "source": "Bronze",
       "difficulty": "Normal",
       "isStarred": false,
-      "tags": [],
+      "tags": ["Simulation", "Greedy"],
```
bqi343: why greedy?

content/extraProblems.json
johnzhou721: The internal solution is DSU so that's a valid approach, but Imma add connected components and put DSU to back.

Good catch

```suggestion
      "tags": ["Connected Components", "Graphs", "DSU"],
```

content/extraProblems.json
johnzhou721: Because prioritizing the first number prioritizes the overall lexicographical order

but that's inherent to lexicographical order and not anything we're inventing so removing greedy

```suggestion
      "tags": ["Simulation"],
```

content/extraProblems.json
bqi343: we generally don't tag problems with DSU if DSU is not necessary to solve. Plus, it's a Gold topic and this problem is in Silver.

johnzhou721: sure thx will remove

bqi343 (approved):

johnzhou721:
Thanks!

PR #6424, opened 2026-08-01
Sync Problems Page filters with the URL

### LINKED ISSUE #5841: Contact Form Submission - Suggestion (https://usaco.guide/problems/)

**URL**: https://usaco.guide/problems/
**Module**: https://usaco.guide/problems/
**Topic**: Suggestion
**Message**: 
I lose my filter options after reload.
Can it like save my filter options or give me url like `usaco.guide/problems/?filter=...`?. So, I can bookmark the URL.

Also, **sort option is missing.**

bqi343:
Closes #5841 — filter options were lost on reload, and there was no way to bookmark/share a filtered view.

Search query, filters, sort mode, and page number on the Problems Page are now reflected in the URL and restored on load, so filtered views survive reloads and can be bookmarked/shared. Example:

```
/problems?q=knapsack&difficulty=Easy&difficulty=Normal&section=Gold&status=Solved&sort=Difficulty%20(Ascending)&page=2
```

- **`routing.ts` (new)**: InstantSearch routing with a custom state mapping producing short, readable params (`q`, `page`, `difficulty`, `modules`, `source`, `tags`, `starred`, `section`). Multi-value filters use repeated params rather than comma-joins since titles can contain commas. The Section filter is stored as section labels and expanded back into module IDs from `MODULE_ORDERING`, and the custom `createURL` preserves the page-managed `sort`/`status` params that InstantSearch doesn't know about.
- **`Selection.tsx`**: the dropdowns previously kept their own state and pushed it into the InstantSearch UI state via an effect that (a) clobbered URL-restored state on mount and (b) replaced the whole index UI state, wiping the query and page on every filter change. They now derive their value directly from `indexUiState` (making them controlled, so restored filters are visible) and write with the previous state spread in.
- **`SearchBox.tsx`**: same clobber fix for the query — the (initially empty) search term is no longer written to InstantSearch until the user types, and a query restored from the URL is reflected in the input.
- **`index.tsx`**: `sort` and `status` are managed outside InstantSearch, so they're read from / written to the URL directly. `?status=Solved` stores status *labels*; `StatusFilterSync` expands them into matching problem IDs once user progress loads (until then results briefly render unfiltered rather than flashing zero results).

Known limitation: browser back/forward restores the InstantSearch-managed params but not `sort`/`status`, which are only read on initial load.

Verified with `tsc --noEmit`, ESLint (no new warnings beyond pre-existing patterns), and a dev-server smoke test of `/problems` with params (HTTP 200, no server errors). Client-side behavior (restore-on-reload, URL updates while filtering) should be checked in a browser before merging.

### DIFF
--- a/src/components/ProblemsPage/SearchBox.tsx
+++ b/src/components/ProblemsPage/SearchBox.tsx
@@ -7,14 +7,16 @@ export default function SearchBox(props: UseSearchBoxProps): JSX.Element {
   // https://stackoverflow.com/questions/53314857/how-to-focus-something-on-next-render-with-react-hooks
   const { query, refine: setQuery } = useSearchBox(props);
   const inputRef = useRef<HTMLInputElement>(null);
-  const [searchTerm, setSearchTerm] = React.useState('');
+  const [searchTerm, setSearchTerm] = React.useState(query);
+  // Don't write the (initially empty) search term to InstantSearch until the
+  // user has typed something, so we don't clobber a query restored from the URL
+  const dirty = useRef(false);
   const debouncedSearchTerm = useDebounce(searchTerm, 200);
   useEffect(() => {
-    if (debouncedSearchTerm) {
-      setQuery(searchTerm);
-    } else {
-      setQuery('');
-    }
+    if (!dirty.current) setSearchTerm(query);
+  }, [query]);
+  useEffect(() => {
+    if (dirty.current) setQuery(debouncedSearchTerm || '');
   }, [debouncedSearchTerm]);
   useEffect(() => {
     if (inputRef.current) inputRef.current.focus();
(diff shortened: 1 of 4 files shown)


bqi343:
Fixed the bug where the difficulty filter disappeared on reload (and results could briefly render unfiltered).

Root cause: react-instantsearch deep-compares each hook's props and, when they change, tears down and recreates the widget — and a disposed refinement-list widget drops its active refinement from the search state. The difficulty filter's `sortBy` comparator (added in #6422) was an inline arrow function, so its identity changed on every page re-render — e.g. when user progress finishes loading right after page load — remounting the widget and silently discarding the difficulty refinement, after which the router rewrote the URL without it. (This dropping predates this PR; the URL sync just made it visible.) The other filters were unaffected because their `sortBy: ['name:asc']` arrays compare deep-equal.

Fix: hoist the comparator to module scope so its identity is stable.

Verified with Playwright against a local dev server: loading `/problems?difficulty=Very%20Easy&section=General` and reloading now keeps both params in the URL, shows both chips, and every rendered hit is Very Easy with all "Appears In" links under `/general/`.

PR #6433, opened 2026-08-02
navigation shortcut

Sosuke23:


bqi343 (approved):
updated the keyboard shortcut so that it works on mac (plus other minor changes)

PR #6448, opened 2026-08-04
Add usaco silvre 2024 december- Deforestation to coordinate compression  

4nav:

- [x] I have tested my code.

rn its under priority queues, but 1) we need coordinate compression and 2) we need custom comparators so it shd be here. we can also solve it with a bit and avoid the pq sol

### DIFF
--- a/content/3_Silver/Sorting_Custom.problems.json
+++ b/content/3_Silver/Sorting_Custom.problems.json
@@ -177,6 +177,18 @@
       }
     },
     {
+      "uniqueId": "usaco-1447",
+      "name": "Deforestation",
+      "url": "http://www.usaco.org/index.php?page=viewproblem2&cpid=1447",
+      "source": "Silver",
+      "difficulty": "Very Hard",
+      "isStarred": false,
+      "tags": ["Sorting", "Coordinate Compression", "Binary Indexed Trees", "Greedy"],
+      "solutionMetadata": {
+        "kind": "internal"
+      }
+    },
+    {
       "uniqueId": "usaco-967",
       "name": "Meetings",
       "url": "http://www.usaco.org/index.php?page=viewproblem2&cpid=967",
--- a/content/3_Silver/Sorting_Custom.problems.json
+++ b/content/3_Silver/Sorting_Custom.problems.json
@@ -183,7 +183,7 @@
       "source": "Silver",
       "difficulty": "Very Hard",
       "isStarred": false,
-      "tags": ["Sorting", "Coordinate Compression", "Binary Indexed Trees", "Greedy"],
+      "tags": ["Sorting", "Binary Search"],
       "solutionMetadata": {
         "kind": "internal"
       }
--- a/solutions/silver/usaco-1447.mdx
+++ b/solutions/silver/usaco-1447.mdx
@@ -5,7 +5,7 @@ title: Deforestation
 author: Arnav Gokhale
 ---
 
-[Official Analysis (C++)](https://usaco.org/current/data/sol_prob2_silver_dec24.html)
+[Official Analysis (Python)](https://usaco.org/current/data/sol_prob2_silver_dec24.html)
 
 ## Explanation
 


4nav:
i had written a whole edi till i realised it alr existed, just in the pq module

Sosuke23:
> i had written a whole edi till i realised it alr existed, just in the pq module

you can mention it under Alternate Solution

johnzhou721 (approved):
👍

4nav:
> > i had written a whole edi till i realised it alr existed, just in the pq module
> 
> you can mention it under Alternate Solution

ideally how short should an alternate solution be?

johnzhou721:
> > > i had written a whole edi till i realised it alr existed, just in the pq module
> > 
> > 
> > you can mention it under Alternate Solution
> 
> ideally how short should an alternate solution be?

From my experience, depends on the content.  Usually if it just duplicates most of the solution and only uses an alternative approach for a small part of the problem, then infobox it.  You can also just point people at observations made in the first solution as well if they're in common.

Sosuke23:
> > > i had written a whole edi till i realised it alr existed, just in the pq module
> > 
> > 
> > you can mention it under Alternate Solution
> 
> ideally how short should an alternate solution be?

doesn't matter. consider it as an editorial only.

4nav:
> > > > i had written a whole edi till i realised it alr existed, just in the pq module
> > > 
> > > 
> > > you can mention it under Alternate Solution
> > 
> > 
> > ideally how short should an alternate solution be?
> 
> doesn't matter. consider it as an editorial only.

I had written this:

---
id: usaco-1447
source: USACO Silver 2024 December
title: Deforestation
author: Abhinav Singh
---
[Official Analysis (Python)](https://usaco.org/current/data/sol_prob2_silver_dec24.html)

## Explanation
Maximizing the number of trees cut is the same thing as minimizing the number of trees kept, so it's easier to think of this in reverse as a "keeping" problem rather than a "cutting" one: each constraint forces at least $t_i$ trees to survive inside $[l_i, r_i]$, and we want to find the smallest possible set of surviving trees that satisfies every constraint at once.

The coordinates are as large as $10^9$, so we first sort the tree positions and convert each query into an interval of tree indices. For a query $[l, r, t]$, let

- $L$ be the index of the first tree with position $\ge l$,
- $R$ be the index of the last tree with position $\le r$.

These can be found using `lower_bound` and `upper_bound`. From this point onward, each query simply asks that at least $t$ trees be kept among indices $[L, R]$.

Now that every query is an interval over tree indices, the problem reduces to selecting the smallest possible set of trees to keep while satisfying every interval constraint. Process the queries in increasing order of their right endpoint. Suppose we are processing a query $[L, R]$ requiring at least $t$ kept trees. If there are already at least $t$ kept trees in the interval, we do nothing. Otherwise, we must keep additional trees. Which ones should we choose?

We claim the greedy choice is to always keep the **rightmost** available tree(s) in the interval.

<Spoiler title="Proof">

Suppose the current query $[L, R]$ requires us to keep one more tree, and we have two available choices $x < y$ within the interval. Choosing $y$ instead of $x$ cannot invalidate any previous or current query, and it can only help satisfy future queries, since future queries end no earlier than $R$ and may have larger left endpoints. Therefore, it is always optimal to keep the **rightmost** available tree.

</Spoiler>

To implement this greedy efficiently, we need to support two operations:

- Count how many trees have already been kept in an interval.
- Mark a tree as kept.

We maintain a Fenwick Tree where each available tree contributes `1`. Once a tree is kept, its value becomes `0`.

Thus, for an interval $[L, R]$,

$$
\text{kept} = (R - L + 1) - \text{rangesum}(L, R).
$$

If fewer than $t$ trees have been kept, we repeatedly keep the **rightmost** unkept tree until the constraint is satisfied. To find it, binary search on the index. For a candidate position `mid`, check whether the suffix $[mid, R]$ contains an unkept tree:

$$
\text{rangesum}(mid, R) > 0.
$$

Since this predicate is monotonic, binary searching for the last valid position returns the rightmost unkept tree. After processing all queries, if `kept` is the number of trees we were forced to keep, the answer is simply $N - \text{kept}.$

# Implementation
**Time Complexity:** Finding the rightmost unkept tree takes $O(\log^2 N)$ time. As each tree is kept at most once, this operation is performed at most $N$ times, resulting in an overall time complexity of $O(N\log^2 N)$.

<LanguageSection>
<CPPSection>

```cpp
#include <bits/stdc++.h>
using namespace std;

// BeginCodeSnip{Fenwick Tree}
struct BIT {
    int n;
    vector<int> bit;

    BIT(int n) : n(n), bit(n + 1) {}

    // Add val to index idx.
    void add(int idx, int val) {
        while (idx <= n) {
            bit[idx] += val;
            idx += idx & -idx;
        }
    }

    // Returns the prefix sum from 1...idx.
    int sum(int idx) {
        int res = 0;
        while (idx > 0) {
            res += bit[idx];
            idx -= idx & -idx;
        }
        return res;
    }

    // Returns the sum on the interval [l, r].
    int range_sum(int l, int r) {
        if (l > r) {
            return 0;
        }
        return sum(r) - sum(l - 1);
    }
};
// EndCodeSnip

struct Query {
    int l, r, t;
};

// Returns the rightmost unkept tree with index <= r.
int find_rightmost(int r, BIT &ft) {
    if (ft.sum(r) == 0) {
        return -1;
    }
    int lo = 1, hi = r;
    while (lo < hi) {
        int mid = (lo + hi) / 2;
        if (ft.range_sum(mid + 1, r) > 0) {
            lo = mid + 1;
        } else {
            hi = mid;
        }
    }
    return lo;
}

int main() {
    ios::sync_with_stdio(false);
    cin.tie(nullptr);

    int test_cases;
    cin >> test_cases;

    while (test_cases--) {
        int n, k;
        cin >> n >> k;

        vector<int> pos(n);
        for (int &x : pos) {
            cin >> x;
        }

        sort(pos.begin(), pos.end());

        vector<Query> queries(k);
        for (Query &q : queries) {
            int l, r;
            cin >> l >> r >> q.t;

            q.l = lower_bound(pos.begin(), pos.end(), l) - pos.begin() + 1;
            q.r = upper_bound(pos.begin(), pos.end(), r) - pos.begin();
        }

        // Process queries in increasing order of their right endpoint.
        sort(queries.begin(), queries.end(),
             [](const Query &a, const Query &b) {
                 return a.r < b.r;
             });

        BIT ft(n);

        // Initially, every tree is unkept.
        for (int i = 1; i <= n; i++) {
            ft.add(i, 1);
        }

        int kept = 0;

        for (const Query &q : queries) {
            // Compute how many trees have already been kept in this interval.
            int kept_in_interval = (q.r - q.l + 1) - ft.range_sum(q.l, q.r);

            // How many additional trees should we keep until the constraint is satisfied.
            int need = q.t - kept_in_interval;

            while (need > 0) {
                // Greedily keep the rightmost available tree.
                int idx = find_rightmost(q.r, ft);
                ft.add(idx, -1);
                kept++;
                need--;
            }
        }

        cout << n - kept << '\n';
    }

    return 0;
}
```

</CPPSection>
</LanguageSection

johnzhou721:
@4nav Up to this paragraph your solution pretty much duplicates the existing one.  The existing solution says you cut down trees that are to the left as possible, yours says preserve trees to the right, so perhaps briefly bridge the two approachs together, you say "treat this problem as preserving the minimum amount of trees" and just say "Preserving the rightmost trees (or cutting down the leftmost trees as shown in the previous solution) is the most optimal", then your spam your data structure details starting from below and add the impl

sounds good?

> To implement this greedy efficiently, we need to support two operations:
> 
> * Count how many trees have already been kept in an interval.
> * Mark a tree as kept.
> 
> We maintain a Fenwick Tree where each available tree contributes `1`. Once a tree is kept, its value becomes `0`.

bqi343:
> rn its under priority queues, but 1) we need coordinate compression and 2) we need custom comparators so it shd be here. we can also solve it with a bit and avoid the pq sol

- bit isn't a silver topic, I don't see how the existence of a solution with a bit is a reason to move the problem into an earlier silver module
- given that the custom comparators / coord compression module is a prereq for the priority queues module, I don't see the problem with leaving it in priority queues

4nav:
> > rn its under priority queues, but 1) we need coordinate compression and 2) we need custom comparators so it shd be here. we can also solve it with a bit and avoid the pq sol
> 
> - bit isn't a silver topic, I don't see how the existence of a solution with a bit is a reason to move the problem into an earlier silver module
> - given that the custom comparators / coord compression module is a prereq for the priority queues module, I don't see the problem with leaving it in priority queues

No I mean even the pq solution is utilizing coordinate compression and using a custom comparator

bqi343:
yes, and custom comparators module is a prereq for the pq module. if a problem involves concepts from two modules where the two modules are in sequence then the problem is placed in the later module

4nav:
> yes, and custom comparators module is a prereq for the pq module. if a problem involves concepts from two modules where the two modules are in sequence then the problem is placed in the later module

Gotcha thanks! Ima close the thread then

PR #6546, opened 2026-08-25
Added Smol Edi  (USACO)

4nav:
Place an "x" in the corresponding checkbox if it is done or does not apply to this pull request._

- [x] I have tested my code.


solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+If the remaining sum is odd, and at least $3$, then we can instead add $1$ and
+$2$. Their sum is $3$, while their popcount XOR contribution is $0$. 
+This leaves only the case where our remaining sum is $1$. Note that if $K$ was
+odd, our initial construction would have added a $1$ to the sequence, which we
+can replace with a $2$: the popcount remains the same, while the sum increases
+by $1$. Otherwise, no solution exists.
+
+### Implementation
+**Time Complexity:** We iterate over the bits in $K$, so each test case takes $\mathcal{O}(\log K)$ time.
```
eysbutno: we always directly state the time complexity -- if it's nontrivial then you justify it in explanation or add a note

also why is this a H3 header? use `##` instead

solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,123 @@
...
+In particular, adding two copies of $2^{b-1}$ increases the sum by $2^b$ without
+changing the XOR. Thus, we can construct any even remaining sum by considering
+its set bits.
+
+If the remaining sum is odd, and at least $3$, then we can instead add $1$ and
+$2$. Their sum is $3$, while their popcount XOR contribution is $0$. 
+This leaves only the case where our remaining sum is $1$. Note that if $K$ was
+odd, our initial construction would have added a $1$ to the sequence, which we
+can replace with a $2$: the popcount remains the same, while the sum increases
+by $1$. Otherwise, no solution exists.
```
eysbutno:
```suggestion
To have
$\text{popcount}(a_1)\oplus\text{popcount}(a_2) \oplus\cdots\oplus\text{popcount}(a_N)=K$,
then for every set bit $i$ of $K$, we need the XOR to contain $2^i$. The
smallest number we can use is one whose popcount is exactly $2^i$, since $2^i$
has only its $i$-th bit set. To minimize this number, we should put all the
$2^i$ set bits as far to the right as possible, giving us

$$
2^{2^i}-1,
$$

whose binary representation consists of $2^i$ consecutive $1s$. Therefore, for
every set bit $i$ in $K$, we can add $2^{2^i}-1$ to our sequence. This
construction produces a sequence with our desired XOR, while minimising the total sum, which may
be smaller than $M$. 

If so, we need a way to increase the sum while maintaining
our XOR. The easiest way to do this is by adding 2 copies of the same number,
since $\text{popcount}(x)\oplus\text{popcount}(x)=0.$ In particular, adding two copies of $2^{b-1}$ increases the sum by $2^b$ without
changing the XOR. Thus, if the remaining sum is even, we add $2^{i-1}$ to our sequence for each set bit $i$ in it.

If the remaining sum is odd and at least $3$, then we instead add $1$ and
$2$. Their sum is $3$, while their popcount XOR contribution is $0$. 
This leaves only the case where our remaining sum is $1$. Observe that if $K$ is
odd, our initial construction adds a $1$ to the sequence, which we
can replace with a $2$: the popcount remains the same, while the sum increases
by $1$. Otherwise, no solution exists.
```

eysbutno (approved):

solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,117 @@
...
+author: Abhinav Singh
+---
+
+[Official Analysis (C++)](https://usaco.org/current/data/sol_prob1_silver_open25.html)
+
+## Explanation
+
+To have
+$\text{popcount}(a_1)\oplus\text{popcount}(a_2) \oplus\cdots\oplus\text{popcount}(a_N)=K$,
+then for every set bit $i$ of $K$, we need the XOR to contain $2^i$. The
```
johnzhou721, line 14: `set bit` may be a bit jargony, `true bit` may be a bit better.

solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,117 @@
...
+smallest number we can use is one whose popcount is exactly $2^i$, since $2^i$
+has only its $i$-th bit set. To minimize this number, we should put all the
+$2^i$ set bits as far to the right as possible, giving us
```
johnzhou721, line 17: same here

solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,117 @@
...
+id: usaco-1518
+source: USACO Silver 2025 US Open
+title: Sequence Construction
+author: Abhinav Singh
+---
+
+[Official Analysis (C++)](https://usaco.org/current/data/sol_prob1_silver_open25.html)
+
+## Explanation
+
```
johnzhou721, line 11: Motivate the first paragraph by saying that we seek to construct the *minimum* sum for a before figuring out how to increase it.  Explain the reasons we want to do this, including how it's feasible to increase the sum properly at the end.  To accomplish the last part you might want to move the ending section to the front and reveal that it's possible to change sum by any amount under fixed XOR under normal conditions first, as that's what really motivates everything else imho

solutions/silver/usaco-1518.mdx
```diff
@@ -0,0 +1,117 @@
...
+$$
+
+whose binary representation consists of $2^i$ consecutive $1s$. Therefore, for
+every set bit $i$ in $K$, we can add $2^{2^i}-1$ to our sequence. This
+construction produces a sequence with our desired XOR, while minimising the total sum, which may
+be smaller than $M$.
+
+If so, we need a way to increase the sum while maintaining
+our XOR. The easiest way to do this is by adding 2 copies of the same number,
+since $\text{popcount}(x)\oplus\text{popcount}(x)=0.$ In particular, adding two copies of $2^{b-1}$ increases the sum by $2^b$ without
```
johnzhou721, line 30:
```suggestion
since $\text{popcount}(x)\oplus\text{popcount}(x)=0$. In particular, adding two copies of $2^{b-1}$ increases the sum by $2^b$ without
```

Good LaTeX hygiene

johnzhou721 (changes requested):
Extremely clean code; well done but pedagogically I think the structure can be slightly reworked

johnzhou721:
@4nav bump for resolving some of my minor things and then we can approve + eysbutno can merge

bqi343:
pre-commit.ci autofix

johnzhou721:
@4nav bump

PR #6573, opened 2026-09-01
Sort Gold Conclusion problems by difficulty

bqi343:
`cf-1582F1` "Korney Korneevich and XOR (easy version)" (Normal) was sitting at the very end of the Gold Conclusion list, after two Very Hard problems.

Stable-sorted the list by difficulty so it lands at the end of the Normal block (right after `cf-1606E` "Arena"). Every other problem keeps its exact relative position — the diff is a pure block move.

Also checked the Bronze, Silver, and Platinum conclusion lists: all three were already in non-decreasing difficulty order, so no changes there.

https://claude.ai/code/session_01V2rsBuxuYPtH6EgkVq8akH

### DIFF
--- a/content/4_Gold/Conclusion.problems.json
+++ b/content/4_Gold/Conclusion.problems.json
@@ -498,6 +498,18 @@
       }
     },
     {
+      "uniqueId": "cf-1582F1",
+      "name": "Korney Korneevich and XOR (easy version)",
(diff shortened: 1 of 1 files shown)


bqi343:
Duplicate of #6572, which makes the identical change. Closing in favor of that one.

PR #6571, opened 2026-08-31
Problem Suggestion: Add "Glass Half Spilled" to Gold - Additional Practice for USACO Gold

maggieliu05:
User `k5QXWVEUjsVWftJnk3U6wuYuPlq2` suggested adding the problem [Glass Half Spilled](https://codeforces.com/contest/1458/problem/B) to the `problems` table of the module [Gold - Additional Practice for USACO Gold](https://usaco.guide/gold/gold-conclusion#problemlist-problems).

**Automatically Generated JSON:**
```json
{
  "uniqueId": "cf-1458B",
  "name": "Glass Half Spilled",
  "url": "https://codeforces.com/contest/1458/problem/B",
  "source": "CF",
  "difficulty": "Normal",
  "isStarred": false,
  "tags": [
    "Knapsack"
  ],
  "solutionMetadata": {
    "kind": "autogen-label-from-site",
    "site": "CF"
  }
}
```
**Additional Notes**:
Should be an easy exercise on knapsack, probably fits better in the knapsack module

*This PR was automatically generated from a user-submitted problem suggestion on the USACO guide.*

PR #6576, opened 2026-09-03
Glass Half Spilled to Knapsack Module

Sosuke23:
#6571


eysbutno (approved):

eysbutno:
on second thought this is kind of like between an easy/normal imo, either is fine

PR #6579, opened 2026-09-05
Editorial - Restorer Distance

Sosuke23:
_Place an "x" in the corresponding checkbox if it is done or does not apply to this pull request._

- [ ] I have tested my code.
- [ ] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [ ] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions).
  - I understand that if it is clear that I have not attempted to follow these conventions, my PR will be closed.
  - If changes are requested, I will re-request a review after addressing them.
- [ ] I have linked this PR to any issues that it closes.

PR #6497, opened 2026-08-18
Create internal sol (python and C++) for Fenced In problem

raagavb:
_Place an "x" in the corresponding checkbox if it is done or does not apply to this pull request._

- [x] I have tested my code.
- [x] I have added my solution according to the steps [here](https://usaco.guide/general/adding-solution#steps).
- [x] I have followed the code conventions mentioned [here](https://usaco.guide/general/adding-solution/#code-conventions).
  - I understand that if it is clear that I have not attempted to follow these conventions, my PR will be closed.
  - If changes are requested, I will re-request a review after addressing them.
- [x] I have linked this PR to any issues that it closes.

PR #6622, opened 2026-09-08
Run the content checks on every pull request, so they can gate

bqi343:
Prerequisite for adding `check-mdx` and `check-problem-data` to master's required status checks.

A required check that never runs blocks a PR forever — GitHub waits on a status that will not arrive. A `paths` filter is precisely how a check comes not to run. #6596 shows the shape: it touched only `content/ordering.ts` and `src/`, and `check-solutions`, `check-problem-tags` and `code-snippets` did not report at all. That is why the three checks master requires today (`Vercel`, `build-tests`, `pre-commit.ci - pr`) are all unconditional.

So the filters come off `mdx.yml` and `problem-data.yml`. Nothing is lost by it: `check_problems_json.py` compares each problem against its copies in other modules, and `check_mdx.py` sweeps every `.mdx`, so both already do the same work whatever the diff contains. The filter was saving a few seconds off a job that takes a few seconds.

The other content workflows keep their filters and so remain unable to gate — worth revisiting separately.

Verified against all 11 open PRs merged with master: 2 would fail, both true positives (#6579's `std::cout.tie(0)`, and a new absolute in-guide link in #6497). No false positives.

https://claude.ai/code/session_01BBSt9Q5eeMr3X23ENJASSz

### DIFF
--- a/.github/workflows/mdx.yml
+++ b/.github/workflows/mdx.yml
@@ -5,15 +5,15 @@
 # each rule is noise without that.
 name: MDX
 
+# No `paths` filter. These are required status checks, and a required check that
+# never runs leaves a pull request waiting on a status that will not arrive -- so
+# it has to report on every pull request, not only the ones touching content.
+# Nothing is lost by that: py/check_mdx.py sweeps every .mdx,
+# so it does the same work whatever the diff contains, in a few seconds.
 on:
   pull_request:
     branches:
       - master
-    paths:
-      - 'content/**/*.mdx'
-      - 'solutions/**/*.mdx'
-      - 'py/check_mdx.py'
-      - '.github/workflows/mdx.yml'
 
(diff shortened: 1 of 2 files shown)


bqi343:
Not needed. `check-mdx` and `check-problem-data` are already gated: both run as pre-commit hooks, and `pre-commit.ci - pr` is a required check that runs unconditionally on every PR. #6579 shows it — `check-mdx` red and `pre-commit.ci - pr` red together.

The `paths` filters are therefore correct as they are; the workflows are a full-sweep supplement (they run `--all`, while the hooks see only changed files), and running them unconditionally would just duplicate work.

The real gap is `code-snippets` and `check-problem-tags`, neither of which pre-commit.ci covers — handled separately.

🤖 Written by [Claude Code](https://claude.com/claude-code), posted from @bqi343's account.

PR #6638, opened 2026-09-13
Linear Diophnatine Equation

Sosuke23:


bqi343:
the topic seems fine to introduce but the application is sort of boring imo? considering it only requires the linear diophantine equation with $c=2$ which is nearly the same as $c=1$

better to just add the application problem to the problems list

I'll edit it

bqi343 (approved):

scripts/watch-content.ts
bqi343, line 1: fixed some issues with local hot reload

Sosuke23: it's looks good now.

Issue #1594, opened 2021-05-25
Groups improvements 


thecodingwizard:
- leaderboard styling fixes, esp on mobile
- ability to modify point value or delete dumb submissions 
- ability to "group" posts into sections

Anything else?

cc @jeffreymeng

ghost:
Button for Going back to [Groups](https://usaco.guide/groups/) from [Group](https://usaco.guide/groups/d7eYGfddXq3m2trXG2xt/) would be a nice fix. Its feel tedious to remove part of url or to dig into navbar.

Similar to home button on this page.
<img width="953" height="517" alt="Image" src="https://github.com/user-attachments/assets/a723571a-1d2c-4223-ad05-fddeedf9af88" />

No Home button present for going back to [Groups](https://usaco.guide/groups/)
<img width="1238" height="806" alt="Image" src="https://github.com/user-attachments/assets/97c7f326-aba8-43d4-9459-2dbf76fcb780" />

MinimulThoughtfulCode:
> Button for Going back to [Groups](https://usaco.guide/groups/) from [Group](https://usaco.guide/groups/d7eYGfddXq3m2trXG2xt/) would be a nice fix. Its feel tedious to remove part of url or to dig into navbar.

Addressed #6454
Added a home button

aunncodes:
Added limit join links to certain emails

PR #6647, opened 2026-09-21
Mobile leaderboard fix 2

Discovered12345:
Fixes part of #1594

This shows LeaderboardTable on sm+ screens as before, and swaps to simpler LeaderboardList (name + total points) on mobile.


emhgit:
would it be possible to display their pfp as well?

Discovered12345:
removed restriction of being hidden below lg breakpoint

emhgit (approved):

